PR #4, opened 2023-04-27
Add support for mardown tables, sanitize the mardown before rendering

ashu8912:


app-catalog/src/components/charts/List.tsx
```diff
@@ -154,14 +154,16 @@ export function ChartsList() {
             return (
               <Box maxWidth="30%" width="400px" m={1}>
                 <Card>
+                  <div style={{height: '60px'}}>
```
joaquimrocha: Please use a Box from MUI and indent.

joaquimrocha (changes requested):

app-catalog/src/components/charts/List.tsx
ashu8912: updated

joaquimrocha (approved):

PR #7, opened 2023-06-08
app-catalog: Fix wrong version being used in Upgrade request

yolossn:
The chart name was being sent as chart version and this was causing the upgrade request to fail, this patch fixes it.


illume (approved):
👍 thanks

illume:
(had to run format on it)

PR #8, opened 2023-06-15
app-catalog: Add the backend token to the helm-related requests

joaquimrocha:
We have recently restricted some backend endpoints to the use of a
token generated by the app, so this plugin was failing since it needed
to use it.

### DIFF
--- a/app-catalog/src/api/releases.tsx
+++ b/app-catalog/src/api/releases.tsx
@@ -1,34 +1,42 @@
 import { ApiProxy } from '@kinvolk/headlamp-plugin/lib';
 
 const request = ApiProxy.request;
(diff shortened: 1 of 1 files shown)


app-catalog/src/api/releases.tsx
```diff
@@ -1,34 +1,42 @@
 import { ApiProxy } from '@kinvolk/headlamp-plugin/lib';
```
illume, line 1:
```ts
import { ApiProxy, getHeadlampAPIHeaders } from '@kinvolk/headlamp-plugin/lib';
```

illume (approved):
👍

PR #12, opened 2023-07-11
App catalog fixes

ashu8912:
Fixes:

- [x] Old version showing up in dialog box if new charts are opened
- [x] Notify user if adding a repo fails

Adds:
- [x] Notify the user about install, update and delete request being accepted 
- [x] Adds required description field in install chart Dialog to be entered by user


illume (approved):
👍 nice changes. Thanks.

PR #14, opened 2023-10-09
plugin: Init prometheus plugin

yolossn:
Tasks:
- [x] Add charts
- [x] Add a skeleton (placeholder chart + text informing the user about how to get the chart working)
- [x] Add a chart visibility toggle button
- [x] Remember user settings (what chart was enabled last time + whether it's visible or not)
- [ ] Review colors


prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+  prometheusPrefix: string;
+  autoRefresh: boolean;
+  XTickProps: {} | null;
+  YTickProps: {} | null;
+  CustomTooltip?: ({ active, payload, label }) => JSX.Element | null;
+}) {
+  //
+  const [metrics, setMetrics] = useState<any>({});
+  const [state, setState] = useState<any>(null);
+  const [error, setError] = useState<any>(null);
```
joaquimrocha: `any` should be avoided as it removes any type checking that we could otherwise benefit from.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+      setState('loading');
+    }
+    try {
+      for (const plot of plots) {
+        const response = await fetchMetrics(
+          props.prometheusPrefix,
+          plot.query,
+          `${tenMinutesBefore}`,
+          `${currentTime}`,
+          `${2}`
```
joaquimrocha: These literals should be avoided. These are numeric values. If we need them in string for this function, it should be the function making the conversion within, as it's an implementation details that should be encapsulated.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+    // Calculate the time that was 10 minutes before the current time
+    const tenMinutesInMilliseconds = 10 * 60; // 10 minutes in milliseconds
+    const tenMinutesBefore = currentTime - tenMinutesInMilliseconds;
+
+    const fetchedMetrics = {};
+
+    if (firstLoad) {
+      setState('loading');
+    }
+    try {
```
joaquimrocha: This big try block can be fired for any reason, unrelated to the issues we're likely trying to catch (the `fetchMetrics` erroring out).
It's a good practice to only catch exceptions for things we cannot really control.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const [error, setError] = useState<any>(null);
+
+  const fetchMetricsData = async (
+    plots: Array<{ query: string; name: string; dataProcessor: (data: any) => any }>,
+    firstLoad: boolean = false
+  ) => {
+    // Get the current time in milliseconds since the epoch
+    const currentTime = Date.now() / 1000;
+
+    // Calculate the time that was 10 minutes before the current time
```
joaquimrocha: This comment is not needed.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+    const tenMinutesInMilliseconds = 10 * 60; // 10 minutes in milliseconds
```
joaquimrocha: The comment here is the same as the var name.
Probably we also don't need this extra const if we're only using it to calculate the next one, and the next one is pretty clear in what it is.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+      // if all the plots are in error state, set the state to error
+      if (Object.values(fetchedMetrics).every(plot => plot.state === 'error')) {
+        setState('error');
+      }
+      // if all the plots are in no data state, set the state to no data
+      else if (Object.values(fetchedMetrics).every(plot => plot.state === 'no data')) {
+        setState('no data');
+      }
+      // if all the plots are in success state, set the state to success
+      else if (Object.values(fetchedMetrics).every(plot => plot.state === 'success')) {
```
joaquimrocha: Is there a case where we may have some metrics that are fetched while others aren't yet or have an error?
i.e. this logic considers you always get all metrics in the same state. It doesn't sound very robust but if this is expected (always the same state), then we don't need to check for all metrics to be in the same state before assuming it, in the above conditionals.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+        <Legend />
+        {props.plots.map(plot => (
+          <Area
+            stackId="1"
+            type="monotone"
+            dataKey={plot.name}
+            stroke={plot.strokeColor}
+            fill={plot.fillColor}
+            activeDot={{ r: 8 }}
+            animationDuration={props.autoRefresh ? 0 : 400} // Disable animation when refreshing
```
joaquimrocha, line 145: I like animations :) Is there a reason for turning them off when refreshing?

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,179 @@
...
+      <Box
+        width="100%"
+        height="100%"
+        p={2}
+        display="flex"
+        justifyContent="center"
+        alignItems="center"
+      >
+        <Box>
+          <Typography variant="h5">Error:{error}</Typography>
```
joaquimrocha: Maybe you need a space between `Error:` and `{error}`

prometheus/src/common.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+  return (
+    <SectionBox>
+      <SectionHeader
+        title="Metrics"
+        actions={[
+          <div style={{ paddingRight: '1rem' }}>
+            <IconButton
+              onClick={() => {
+                setRefresh(!refresh);
```
joaquimrocha: States can sometimes be tricky. The way to be sure that we're always affecting the current state is by using the function variant:
`setRefresh(isRefreshing => !isRefreshing)`

prometheus/src/common.tsx
```diff
@@ -0,0 +1,380 @@
...
+              }}
+            >
+              {refresh ? <Icon icon="tabler:refresh" /> : <Icon icon="tabler:refresh-off" />}
+            </IconButton>
+          </div>,
+          <ButtonGroup disableElevation size="small" aria-label="small outlined button group">
+            <Button
+              value="cpu"
+              onClick={handleChartVariantChange}
+              variant={chartVariant === 'cpu' ? 'contained' : 'outlined'}
```
joaquimrocha: I'll have to double check this style, so I may have more comments later.

prometheus/src/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+} from '@kinvolk/headlamp-plugin/lib';
+import Alert from '@material-ui/lab/Alert';
+import React, { useEffect, useState } from 'react';
+import { GenericMetricsChart } from './common';
+import { isPrometheusInstalled } from './request';
+
+function MetricsEnabled({ children }) {
+  const [prometheusInstalled, setPrometheusInstalled] = useState<boolean>(false);
+  const [promPodName, setPromPodName] = useState<string>(null);
+  const [promNamespace, setPromNamespace] = useState<string>(null);
```
joaquimrocha: I think a modern version of React does batch the state changes, so this should be okay. But this these are all related, maybe we could have it as: `useState<{podName: string; podNamespace: string} | null>`, null being "Not installed".

prometheus/src/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const [state, setState] = useState<any>(null);
```
joaquimrocha: Could be `isLoading` and be a null. Unless we will use it for other things.

prometheus/src/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <Alert severity="info">Fetching Prometheus Info</Alert>
+      ) : (
+        <Alert severity="info">Prometheus is not installed</Alert>
+      )}
+    </div>
+  );
+}
+
+function PrometheusMetrics(resource: DetailsViewSectionProps) {
+  console.log(resource, resource.kind);
```
joaquimrocha: If we should log, then let's add more context.

prometheus/src/request.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+  return [false, null, null];
+}
+
+export function fetchMetrics(
+  prefix: string,
+  query: string,
+  from: string,
+  to: string,
+  step: string
```
joaquimrocha: I'd add these as an object, as functions will many args are difficult to read.

joaquimrocha (changes requested):

prometheus/src/chart.tsx
yolossn: The chart repaints the plot with a transition and seeing that happen every 15 seconds when data refreshes is kind of disturbing.

prometheus/src/chart.tsx
yolossn: The reason why there is a need for all of them to be in success state is the logic to merge the data from multiple metrics into a single object will fail.

>  then we don't need to check for all metrics to be in the same state before assuming it, in the above conditionals.

Sorry I didn't get this part, can you expand?

prometheus/src/index.tsx
yolossn: I have moved this logic to the `GenericMetricsChart` component as the current logic doesn't even show the section header if Prometheus is not installed.

prometheus/src/chart.tsx
```diff
@@ -19,36 +19,37 @@ export function Chart(props: {
   CustomTooltip?: ({ active, payload, label }) => JSX.Element | null;
 }) {
   //
-  const [metrics, setMetrics] = useState<any>({});
```
joaquimrocha: There's an empty comment above this line.

prometheus/src/chart.tsx
```diff
@@ -19,36 +19,37 @@ export function Chart(props: {
...
     plots: Array<{ query: string; name: string; dataProcessor: (data: any) => any }>,
     firstLoad: boolean = false
   ) => {
-    // Get the current time in milliseconds since the epoch
     const currentTime = Date.now() / 1000;
 
-    // Calculate the time that was 10 minutes before the current time
-    const tenMinutesInMilliseconds = 10 * 60; // 10 minutes in milliseconds
-    const tenMinutesBefore = currentTime - tenMinutesInMilliseconds;
+    const tenMinutesBefore = currentTime - 10 * 60;
```
joaquimrocha, line 40: The previous comment said this was in milliseconds but I realize now that this is in seconds. Hope the end result is still fine.

prometheus/src/chart.tsx
```diff
@@ -61,35 +62,36 @@ export function Chart(props: {
...
+          const mergedElement = { timestamp: element.timestamp };
+          for (const plotName of Object.keys(fetchedMetrics)) {
+            mergedElement[plotName] = fetchedMetrics[plotName].data[index].y;
+          }
+          return mergedElement;
+        }
+      );
+      setMetrics(mergedData);
+      setState('success');
+    }
```
joaquimrocha, line 102: We need the default case as mentioned in our call.

prometheus/src/chart.tsx
```diff
@@ -19,36 +19,37 @@ export function Chart(props: {
   CustomTooltip?: ({ active, payload, label }) => JSX.Element | null;
 }) {
   //
-  const [metrics, setMetrics] = useState<any>({});
-  const [state, setState] = useState<any>(null);
-  const [error, setError] = useState<any>(null);
+  const [metrics, setMetrics] = useState<object>({});
+  const [state, setState] = useState<string | null>(null);
```
joaquimrocha: Let's use an enum for the state.

prometheus/src/chart.tsx
```diff
@@ -150,7 +152,7 @@ export function Chart(props: {
         alignItems="center"
       >
         <Box>
-          <Typography variant="h5">Error:{error}</Typography>
+          <Typography variant="h5">Error: {error}</Typography>
```
joaquimrocha: Please use the Empty common component (`<Empty color="error">`)

prometheus/src/common.tsx
```diff
@@ -16,6 +18,31 @@ export function GenericMetricsChart(props: {
...
 
+  const [prometheusInfo, setPrometheusInfo] = useState<{
+    podName: string;
+    podNamespace: string;
+  } | null>(null);
+  const [state, setState] = useState<string>('');
+
+  useEffect(() => {
+    (async () => {
+      setState('loading');
```
joaquimrocha: Let's use an enum here too.

prometheus/src/common.tsx
```diff
@@ -16,6 +18,31 @@ export function GenericMetricsChart(props: {
...
+          setPrometheusInfo({ podName, podNamespace: namespace });
+        } else {
+          setPrometheusInfo(null);
+        }
+      } catch (e) {
+        console.error(e);
+        setState('error');
+        return;
+      }
+      setState('installed');
```
joaquimrocha: This sets it as installed even when the prometheus info may be null from line 35.

prometheus/src/common.tsx
```diff
@@ -159,11 +212,11 @@ function CustomTooltipFormatBytes({ active, payload, label }) {
...
 }
 
-function formatBytes(bytes) {
+function formatBytes(bytes: number) {
   if (bytes === 0) return '0 B';
   const units = ['B', 'KB', 'MB', 'GB', 'TB'];
-  const i = parseInt(Math.floor(Math.log(bytes) / Math.log(1024)));
-  return Math.round(bytes / Math.pow(1024, i), 2) + ' ' + units[i];
+  const i = Math.floor(Math.log(bytes) / Math.log(1024));
+  return Math.round((bytes / Math.pow(1024, i)) * 100) / 100 + ' ' + units[i];
```
joaquimrocha: Is this * 100 / 100 correct? And do we need the space between number and unit? `100Mb vs 100 Mb`

prometheus/src/request.tsx
```diff
@@ -16,28 +16,31 @@ export async function isPrometheusInstalled() {
...
+  query: string;
+  from: number;
+  to: number;
+  step: number;
+}) {
   const params = new URLSearchParams();
-  if (from) {
-    params.append('start', from);
+  if (data.from) {
+    params.append('start', `${data.from}`);
```
joaquimrocha: Use `.toString()` instead of the literals here and in the other examples. Literals should be used for string formatting.

joaquimrocha (changes requested):
Thanks. Here is the review with a few things we had discussed in the call.

prometheus/src/chart.tsx
yolossn: Yes the result is correct, verified it.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,190 @@
...
+          continue;
+        }
+
+        if (response['data']['result'].length === 0) {
+          fetchedMetrics[plot.name] = { data: [], state: 'no data' };
+          continue;
+        }
+
+        const data = plot.dataProcessor(response);
+        fetchedMetrics[plot.name] = { data: data, state: 'success' };
```
joaquimrocha: These should be outside of the try block as I understand.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,190 @@
...
+        setError(e);
+        setState(chartState.ERROR);
+        break;
+      }
+    }
+
+    // if all the plots are in error state, set the state to error
+    if (Object.values(fetchedMetrics).every(plot => plot?.state === 'error')) {
+      setState(chartState.ERROR);
+    }
```
joaquimrocha: You don't need this block as the error will be set as the default case.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,190 @@
...
+          for (const plotName of Object.keys(fetchedMetrics)) {
+            mergedElement[plotName] = fetchedMetrics[plotName].data[index].y;
+          }
+          return mergedElement;
+        }
+      );
+      setMetrics(mergedData);
+      setState(chartState.SUCCESS);
+    } else {
+      // if the plots are in mixed state, set the state to error
```
joaquimrocha: When you remove the first if, this comment needs to be updated.

joaquimrocha (changes requested):

prometheus/src/chart.tsx
yolossn: the `dataProcessor` function is from an argument to the component and there can be a case where some error occurs in its implementation, that is why this is also in the try block.

prometheus/src/chart.tsx
yolossn: Though I agree, I do think keeping it this way makes it easy to understand the code. If the concern here is due to performance then we can change it.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,186 @@
...
+      try {
+        const response = await fetchMetrics({
+          prefix: props.prometheusPrefix,
+          query: plot.query,
+          from: tenMinutesBefore,
+          to: currentTime,
+          step: 2,
+        });
+        if (response.status !== 'success') {
+          fetchedMetrics[plot.name] = { data: [], state: 'error' };
```
joaquimrocha: The state should be `ChartState`, not a string.

prometheus/src/chart.tsx
```diff
@@ -0,0 +1,186 @@
...
+    strokeColor: string;
+    dataProcessor: (data: any) => any[];
+  }>;
+  prometheusPrefix: string;
+  autoRefresh: boolean;
+  XTickProps: {} | null;
+  YTickProps: {} | null;
+  CustomTooltip?: ({ active, payload, label }) => JSX.Element | null;
+}) {
+  enum chartState {
```
joaquimrocha: Let's call this `ChartState` (capitalized).

prometheus/src/chart.tsx
joaquimrocha: You can declare the `dataProcessor` as non-const outside of the try.
What other errors can occur in the response/fetchedMetrics?

My point it: It's normally not a good practice to include things in a try unless you're really expecting them to raise an exception. Which I don't think the `fetchedMetrics` set up is expecting.

joaquimrocha (changes requested):

prometheus/job.yaml
ashu8912, line 1: this shouldn't be here

prometheus/src/common.tsx
```diff
@@ -0,0 +1,432 @@
...
+    prevRenderedTimestamp = currentTimestamp;
+
+    return shouldRenderDate ? `${date.getHours()}:${date.getMinutes()}` : '';
+  };
+}
+
+function dataProcessor(response: any): { timestamp: number; y: number }[] {
+  const data: { timestamp: number; y: number }[] = [];
+  // convert the response to a JSON object
+  response['data']['result'][0]['values'].forEach(element => {
```
ashu8912: the properties should only be accessed if they exist

prometheus/src/chart.stories.tsx
```diff
@@ -0,0 +1,363 @@
...
+  //     fetchMock.get('*', () => {
+  //       return data;
+  //     });
+
+  //   React.useEffect(() => {
+  //     mockData(data);
+  //     return () => {
+  //       fetchMock.restore(); // Cleanup the mock when the component unmounts
+  //     };
+  //   }, []);
```
joaquimrocha: Should this be deleted? If it's here to help with debug/dev, we should add a comment indicating that.

prometheus/src/chart.stories.tsx
```diff
@@ -0,0 +1,363 @@
...
+      return (
+        <BrowserRouter>
+          <Story />
+        </BrowserRouter>
+      );
+    },
+  ],
+} as Meta;
+
+// fetchMock.config.fallbackToNetwork = true;
```
joaquimrocha: Same comment as for the commented out code below.

prometheus/src/chart.tsx
```diff
@@ -2,29 +2,31 @@ import { EmptyContent, Loader } from '@kinvolk/headlamp-plugin/lib/CommonCompone
...
-export function Chart(props: {
+export interface ChartProps {
   plots: Array<{
     query: string;
     name: string;
     fillColor: string;
     strokeColor: string;
     dataProcessor: (data: any) => any[];
   }>;
+  fetchMetrics: (query: object) => Promise<any>;
```
joaquimrocha, line 14: We should avoid having `any` types unless we really accept/expect anything.

ashu8912: it is the type of request function return which is itself any

prometheus/src/chart.stories.tsx
ashu8912: updated

prometheus/src/chart.tsx
ashu8912: updated

prometheus/src/chart.stories.tsx
ashu8912: updated

prometheus/src/chart.tsx
joaquimrocha: That's... not really what we should expect but let's take it just so we can move on.

PR #17, opened 2023-11-03
Fix colors and enable visibility button only on chart enabled resources

joaquimrocha:
These changes fix the colors used for the charts, according to design, and also make sure the visibility button is only shown where the chart is enabled (Workloads).


ashu8912 (approved):
LGTM, works very well

PR #22, opened 2023-11-16
AI plugin

ashu8912:


illume:
Our notes from meeting with @ashu8912 

- event PR needs review/merge
- instructions on how to run ai plugin and push changes to PR
  - run it and paste in the GPT API key
  - accepts both azure and openapi gpt
  - GPT 3.5 accepted, or GPT 4 only... because quality.
- explain, validate, and "Look for issues" buttons need an explaination / tooltip
- [Context: DaemonSet] looks the same as the other buttons 
- The apply button seems to be not a good idea maybe. If it only works 50% or even 80% of the time, then it's dangerous.
  - diff view? To see what's changed.
  - detect if it's an example of how to change something. Eg, a rediness snippet that is not the whole yaml should not be applyable.
  - open the yaml inside an editor? So it can be validated.
    - there's no way for plugins to do this yet.
- icon
- when changing screens the explaination is still there for the old screen.
  - is this confusing?
  - Apply should most probably not be allowed when changing views
- closing the drawer?
- the drawer covers up things. Maybe as a side panel when the screen is big enough.
  - maybe a drawer on medium view and a side bar on desktop?
- mobile view design?
- knowing that the text scrolls 
- needs a danger warning "this may be nonsense / incorrect be careful and check everything"
- can you explain or validate on a list? Maybe don't show that.
- we should have a test plan
  - test 10 times each function to see if it works

PR #28, opened 2024-01-03
prometheus: Add chart for PVC

knrt10:
This adds charts for pvc disk usage and capacity.

Screenshot 2024-01-14 at 10 48 17 AM

### DIFF
--- a/prometheus/src/common.tsx
+++ b/prometheus/src/common.tsx
@@ -194,6 +194,127 @@ export function GenericMetricsChart(props: {
   );
 }
 
+export function DiskMetricsChart(props: {
+  usageQuery?: string;
+  capacityQuery?: string;
+}) {
+  enum prometheusState {
+    UNKNOWN,
+    LOADING,
+    ERROR,
+    INSTALLED,
+  }
+
+  const classes = useStyles();
+  const pluginSettings = usePluginSettings();
+  const [refresh, setRefresh] = useState<boolean>(true);
+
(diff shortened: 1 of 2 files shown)


prometheus/src/common.tsx
```diff
@@ -197,6 +197,142 @@ export function GenericMetricsChart(props: {
...
+          ? [
+            <ToggleButtonGroup
+              onChange={handleChartVariantChange}
+              size="small"
+              aria-label="metric chooser"
+              value={chartVariant}
+              exclusive
+            >
+              <ToggleButton value="disk">Disk</ToggleButton>
+            </ToggleButtonGroup>,
```
yolossn: Since we have only one metric here I don't think we need a Toggle button group here. Just displaying the name should be enough.

prometheus/src/common.tsx
```diff
@@ -197,6 +197,142 @@ export function GenericMetricsChart(props: {
...
+  enum prometheusState {
+    UNKNOWN,
+    LOADING,
+    ERROR,
+    INSTALLED,
+  }
+
+  const classes = useStyles();
+  const pluginSettings = usePluginSettings();
+  const [chartVariant, setChartVariant] = useState<string>('disk');
```
yolossn: In this case there is only one chart variant, remove this and all logic specific to multiple chart variants.

yolossn (dismissed):

yolossn (approved):
LGTM

PR #29, opened 2024-01-03
prometheus: Fix charts CSS

knrt10:
Chart toggle buttons and play/pause buttons were not correctly aligned. Also the y-axis for all charts were going outside the div box.

### DIFF
--- a/prometheus/src/common.tsx
+++ b/prometheus/src/common.tsx
@@ -88,11 +88,10 @@ export function GenericMetricsChart(props: {
         display="flex"
         justifyContent="space-around"
         alignItems="center"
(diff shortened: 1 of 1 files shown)


yolossn:
@joaquimrocha was the chart variant selection Toggle group aligned to the left intentionally?

joaquimrocha:
> @joaquimrocha was the chart variant selection Toggle group aligned to the left intentionally?

I actually wanted the pause button to the right so it's not associated with the other buttons, but it does look better centered liked this so let's keep it.

yolossn (approved):
LGTM

PR #32, opened 2024-03-14
prometheus: Handle Non JSON errors in response

yolossn:
this patch handles the Non JSON errors that
are thrown by the k8s proxy when the pod isn't
in Ready state.

### DIFF
--- a/prometheus/src/request.tsx
+++ b/prometheus/src/request.tsx
@@ -16,13 +16,24 @@ export async function isPrometheusInstalled() {
   return [false, null, null];
 }
 
-export function fetchMetrics(data: {
+/**
+ * Fetches metrics data from Prometheus using the provided parameters.
+ * @param {object} data - The parameters for fetching metrics.
+ * @param {string} data.prefix - The namespace prefix.
+ * @param {string} data.query - The Prometheus query string.
+ * @param {number} data.from - The start time for the query (Unix timestamp).
+ * @param {number} data.to - The end time for the query (Unix timestamp).
+ * @param {number} data.step - The step size for the query (in seconds).
+ * @returns {Promise<object>} - A promise that resolves to the fetched metrics data.
+ * @throws {Error} - Throws an error if the request fails.
+ */
+export async function fetchMetrics(data: {
   prefix: string;
   query: string;
   from: number;
   to: number;
   step: number;
-}) {
+}): Promise<object> {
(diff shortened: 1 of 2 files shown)


prometheus/src/request.tsx
```diff
@@ -37,10 +48,17 @@ export function fetchMetrics(data: {
...
+  const response = await request(
     `/api/v1/namespaces/${data.prefix}/proxy/api/v1/query_range?${params.toString()}`,
     {
       method: 'GET',
+      isJSON: false,
     }
   );
+  if (response.status === 200) {
+    return response.json();
+  }else{
```
knrt10:
```suggestion
  } else{
```

prometheus/src/request.tsx
```diff
@@ -16,7 +16,18 @@ export async function isPrometheusInstalled() {
...
+ * @param {object} data - The parameters for fetching metrics.
+ * @param {string} data.prefix - The namespace prefix.
+ * @param {string} data.query - The Prometheus query string.
+ * @param {number} data.from - The start time for the query (Unix timestamp).
+ * @param {number} data.to - The end time for the query (Unix timestamp).
+ * @param {number} data.step - The step size for the query (in seconds).
+ * @returns {Promise<object>} - A promise that resolves to the fetched metrics data.
+ * @throws {Error} - Throws an error if the request fails.
+ */
+export async function fetchMetrics(data: {
```
knrt10, line 30: It should say `fetchMetrics(data): Promise<any>{}`

prometheus/src/request.tsx
```diff
@@ -37,10 +48,17 @@ export function fetchMetrics(data: {
...
     `/api/v1/namespaces/${data.prefix}/proxy/api/v1/query_range?${params.toString()}`,
     {
       method: 'GET',
+      isJSON: false,
     }
   );
+  if (response.status === 200) {
+    return response.json();
+  }else{
+    const error = new Error(response.statusText);
```
knrt10, line 61: We could improve error handling by using `try, catch` block.

knrt10 (commented):
Nice work Santhosh, some suggestions :-)

prometheus/src/request.tsx
yolossn: The request function which is a wrapper of `fetch` handles the errors internally.

knrt10 (approved):

PR #38, opened 2024-04-26
Add Kompose pluin

joaquimrocha:
It's still at a very early stage but it's usable already.


kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+                } else if (!!output.data) {
+                  setKomposeOutput(output.data);
+                  setError(false);
+                  setProgressStatus('');
+                  setStep(2);
+                }
+              }
+            },
+            {
+              tailLines: 500,
```
yolossn: this will fail if the output is more than 500 lines right?

kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+                }}
+                disabled={step === 0}
+              >
+                Restart
+              </Button>
+            </Box>
+            <Box sx={{ flex: '1 1 auto' }} />
+            <Box>
+              {step !== 2 ? (
+                <Button
```
yolossn, line 263: If the user doesn't have access to create a resource in the cluster the Button should be disabled and an error message be displayed that the user doesn't have the required rights in the cluster.

kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+                  onClick={() => {
+                    // @todo: We should validate the YAML before we enable this button
```
yolossn, line 265: Validation to be added.

kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+import Button from '@mui/material/Button';
+import Step from '@mui/material/Step';
+import StepLabel from '@mui/material/StepLabel';
+import Stepper from '@mui/material/Stepper';
+import Typography from '@mui/material/Typography';
+import {useTheme} from '@mui/material';
+import React from 'react';
+import { getKomposeOuput, komposeNamespace, komposePod } from './kompose';
+
+export function KomposeUI() {
```
yolossn, line 17: could we add a Storybook test for the `KomposeUI` component?

kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+import MonacoEditor from '@monaco-editor/react';
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import Step from '@mui/material/Step';
+import StepLabel from '@mui/material/StepLabel';
+import Stepper from '@mui/material/Stepper';
+import Typography from '@mui/material/Typography';
+import {useTheme} from '@mui/material';
+import React from 'react';
+import { getKomposeOuput, komposeNamespace, komposePod } from './kompose';
```
yolossn:
```suggestion
import { getKomposeOutput, komposeNamespace, komposePod } from './kompose';
```

kompose/src/kompose.tsx
```diff
@@ -0,0 +1,65 @@
...
+              ],
+            },
+          ],
+        },
+      },
+    },
+  };
+};
+
+export function getKomposeOuput(logs: string) {
```
yolossn:
```suggestion
export function getKomposeOutput(logs: string) {
```

yolossn: also add jsdoc

kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+                size="small"
+                variant="outlined"
+              >
+                Stop
+              </Button>
+            </>
+          ) : (
+            <Button
+              onClick={() => {
+                // @todo: The retry is not working apparently
```
yolossn: Retry to be handled

kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+            </Step>
+            <Step>
+              <StepLabel>Convert</StepLabel>
+            </Step>
+            <Step>
+              <StepLabel>Kubernetes YAML</StepLabel>
+            </Step>
+          </Stepper>
+          <Box sx={{ display: 'flex', flexDirection: 'row', pt: 2 }}>
+            <Box>
```
yolossn, line 230: Instead of Restart having a back button will be much clearer.

kompose/src/ui.tsx
```diff
@@ -0,0 +1,290 @@
...
+        />
+      </Box>
+    );
+  }, [theme, komposeOutput, step, progressStatus, error]);
+
+  return (
+    <SectionBox title="Kompose (beta)" paddingTop={2}>
+      <Box>
+        <Typography variant="body1">
+          Kompose is a tool to help users who are familiar with docker-compose move to Kubernetes.
```
yolossn, line 229: Please mention that docker-compose files which contain `build` wont be converted, only compose files with container  image is accepted.

yolossn:
I get this truncated output for the following docker-compose file
```yaml
version: '3.6'

services:
  keycloak_web:
    image: quay.io/keycloak/keycloak:23.0.2
    container_name: keycloak_web
    environment:
      KC_DB: postgres
      KC_DB_URL: jdbc:postgresql://keycloakdb:5432/keycloak
      KC_DB_USERNAME: keycloak
      KC_DB_PASSWORD: keycloak

      KC_HOSTNAME: localhost
      KC_HOSTNAME_PORT: 8080
      KC_HOSTNAME_STRICT: 'false'
      KC_HOSTNAME_STRICT_HTTPS: 'false'

      KC_LOG_LEVEL: info
      KC_METRICS_ENABLED: 'true'
      KC_HEALTH_ENABLED: 'true'
    command: start-dev
    depends_on:
      - keycloakdb
    ports:
      - '8080:8080'


  keycloakdb:
    image: postgres:16
    volumes:
      - postgres_data:/var/lib/postgresql/data
    environment:
      POSTGRES_DB: keycloak
      POSTGRES_USER: keycloak
      POSTGRES_PASSWORD: keycloak

volumes:
  postgres_data:
```
image

kompose/src/ui.tsx
joaquimrocha: I don't think so, it may fail if what we look for is beyond the 500 lines of logs. I will remove the limit.

kompose/src/ui.tsx
joaquimrocha: For now I will just add an extra string to explain the role needs to be allowed to create namespaces/jobs.

kompose/src/ui.tsx
joaquimrocha: Fixed it.

kompose/src/ui.tsx
joaquimrocha: I cannot add it ATM.

kompose/src/ui.tsx
joaquimrocha: I think we can leave this for later as it's a convenience feature.

kompose/src/kompose.tsx
joaquimrocha: Done.

joaquimrocha:
Thanks @yolossn . I have fixed the truncation problem and a few other things you noted.

yolossn (approved):

PR #40, opened 2024-05-01
plugins: Add opencost plugin

yolossn:
**How to install opencost in cluster**

1. Install prometheus using helm
>  helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
> helm install prom prometheus-community/kube-prometheus-stack

2. Create a values.yaml
```yaml
opencost:
  prometheus:
    internal:
      enabled: true
      namespaceName: default
      port: 9090
      serviceName: prom-kube-prometheus-stack-prometheus
```
3. Install opencost
> helm repo add opencost https://opencost.github.io/opencost-helm-chart
> helm install -f values.yaml  my-opencost opencost/opencost

https://github.com/headlamp-k8s/plugins/assets/22731013/6aba7f18-3927-4ebb-8bbb-dc6fb721efba


opencost/src/settings.tsx
```diff
@@ -0,0 +1,85 @@
...
+import MenuItem from '@mui/material/MenuItem';
+import Select from '@mui/material/Select';
+import TextField from '@mui/material/TextField';
+
+export function Settings(props) {
+  const { data, onDataChange } = props;
+
+  const settingsRows = [
+    {
+      name: 'Enable in list view',
```
joaquimrocha, line 13: What does this mean?
I noticed there was a column in the namespaces list view for open cost even if this shows as disabled by default.

opencost/src/settings.tsx
```diff
@@ -0,0 +1,85 @@
...
+          }
+        />
+      ),
+    },
+    {
+      name: 'Namespace',
+      value: (
+        <TextField
+          fullWidth
+          helperText="Enter the namespace of Opencost service"
```
joaquimrocha: What is this needed for? I think we need more explanation on what this is.

opencost/src/settings.tsx
```diff
@@ -0,0 +1,85 @@
...
+          defaultValue={data?.namespace ? data.namespace : ''}
+          onChange={e => onDataChange({ ...data, namespace: e.target.value })}
+          variant="standard"
+        />
+      ),
+    },
+    {
+      name: 'Service URL',
```
joaquimrocha, line 35: Is this because the plugin places a link to that URL? If so, we need a better name and description.

opencost/src/settings.tsx
```diff
@@ -0,0 +1,85 @@
...
+          fullWidth
+          helperText="Enter the service URL of Opencost web UI service"
+          defaultValue={data?.serviceURL ? data.serviceURL : ''}
+          onChange={e => onDataChange({ ...data, serviceURL: e.target.value })}
+          variant="standard"
+        />
+      ),
+    },
+    {
+      name: 'Default timespan',
```
joaquimrocha, line 47: This should have a default value selected.

opencost/src/settings.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <MenuItem value={'week'}>Week</MenuItem>
+            <MenuItem value={'lastweek'}>Last week</MenuItem>
+            <MenuItem value={'7d'}>7 days</MenuItem>
+            <MenuItem value={'14d'}>14 days</MenuItem>
+          </Select>
+        </>
+      ),
+    },
+    {
+      name: 'Display currency',
```
joaquimrocha, line 67: It'd be great if open cost also provided the currency name/symbol so we could use it by default.

joaquimrocha (changes requested):
Left a couple of comments. I also think it should be nice to link to the instructions of how to set up OpenCost.
This could be done in the settings view, and we should have some way to warn the user about the need to set up OpenCost. E.g. a link "Please set up" instead of the value or ..., but this link should only be visible on hover (I think the new table supports this more easily). When OpenCost is not available and it's not configured, I think we should have `!` as the value. So users do not think the values are still loading.

opencost/src/settings.tsx
yolossn: Nope, opencost doesn't provide the symbol it just prepends the symbol the user selects in the UI.

opencost/src/settings.tsx
yolossn: Nope this is the service URL from which Opencost API can be accessed.

opencost/src/settings.tsx
yolossn: Yes it is to show/hide the opencost in the list view. I have fixed the default value for this setting,

opencost/src/settings.tsx
yolossn: Improved the text to `Please enter the namespace where Opencost is Installed`

opencost/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "opencost",
+  "version": "0.0.1",
+  "description": "Your Headlamp plugin",
```
joaquimrocha: Please give it a good description.

opencost/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "opencost",
+  "version": "0.0.1",
```
joaquimrocha: Make it 0.1.0.

opencost/src/settings.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const settingsRows = [
+    {
+      name: 'Enable in list view',
+      value: (
+        <Checkbox
+          checked={data?.isEnabledInListView || true}
+          onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+            onDataChange({ ...data, isEnabledInListView: event.target.checked })
+          }
+        />
```
joaquimrocha: The checkbox is always checked and I cannot uncheck it in my tests. Also, I think it should be a switch, not a checkbox.

opencost/src/settings.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <MenuItem value={'14d'}>14 days</MenuItem>
+          </Select>
+        </>
+      ),
+    },
+    {
+      name: 'Display currency',
+      value: (
+        <TextField
+          fullWidth
```
joaquimrocha: Remove this prop since usually the symbol will be very short in length.

joaquimrocha (changes requested):

yolossn:
Adding few screenshots for reference.
image
image
image
image
image
image
image

opencost/dist/main.js
joaquimrocha, line 1: This is not supposed to be checked in.

joaquimrocha (changes requested):
Found that the main.js is checked in, when it should not.
Also please rebase since there are now fixes in main for CI issues.

yolossn:
image
image
image

PR #41, opened 2024-05-10
Add Tooltip to Pause Button in Prometheus Plugin

vyncent-t:
# Add Tooltip to Pause Button in Prometheus Plugin

## Description

This PR addresses the issue where the pause button in the Prometheus plugin lacked a descriptive label, making it unclear to users what the button's function was upon hovering. A tooltip has been added to provide clear guidance and improve user experience.

## Changes

- [x] Implemented a tooltip for the pause button in the Prometheus plugin that appears on hover.

## Verification

- [x] Tested the tooltip to ensure it displays correctly when hovering over the pause button.
- [x] Verified that the tooltip accurately describes the button's functionality.
- [x] Ensured the tooltip's style and timing align with the application's user interface guidelines.

## Screenshots

image

image

### DIFF
--- a/prometheus/src/common.tsx
+++ b/prometheus/src/common.tsx
@@ -1,6 +1,7 @@
 import { Icon } from '@iconify/react';
 import { Loader, SectionBox } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
 import { Box, Button, Grid, IconButton, Link, Paper, Typography, useTheme } from '@mui/material';
+import { Tooltip } from '@mui/material';
 import Alert from '@mui/material/Alert';
 import ToggleButton from '@mui/material/ToggleButton';
 import ToggleButtonGroup from '@mui/material/ToggleButtonGroup';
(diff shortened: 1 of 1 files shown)


prometheus/src/common.tsx
```diff
@@ -148,7 +149,8 @@ export function GenericMetricsChart(props: {
                 }}
                 size="large"
               >
-                {refresh ? <Icon icon="mdi:pause" /> : <Icon icon="mdi:play" />}
+                {refresh ? <Tooltip title="Pause metrics"> <Icon icon="mdi:pause" /> </Tooltip>
+                  : <Tooltip title="Resume metrics"> <Icon icon="mdi:play" /> </Tooltip>}
```
joaquimrocha: Please format this according to the style we use in Headlamp. I am not sure why the linter is not catching this, as the error in CI seems unrelated.

joaquimrocha (changes requested):

prometheus/src/common.tsx
vyncent-t: fixed!

illume (approved):
👍

Issue #21, opened 2023-11-10
app-catalog: Filter apps based on being verified, official, CNCF


joaquimrocha:
Once we implement #20 , we should have a filter to choose whether we show apps with/without those characteristics.
By default we should show only apps that are verified+official, for security reasons (this makes it more clear that one is installing the expected apps). So users would have to use this new filter in order to show apps that aren't verified or official.

PR #43, opened 2024-05-28
frontend: Adds filter feature for app catalog

vyncent-t:
# Add Filter for Verified and Official Apps in App Catalog

## Description

Fixes Issue #21 

This PR introduces a new filter in the app catalog that allows users to choose whether to display apps based on their verification and official status. By default, the app catalog will only show apps that are verified and official, enhancing security and ensuring that users are installing trusted applications. Users can use the new filter to include apps that are not verified or official if needed.

## Changes

- [x] Added a filter option in the app catalog UI to toggle the display of verified and official apps.
- [x] Set the default filter state to only show verified and official apps for enhanced security.
- [x] Implemented functionality to allow users to change the filter settings to include apps that are not verified or official.
- [x] Updated the app catalog logic to respect the filter settings and dynamically display the appropriate apps.

## Verification

- [x] Verified that the default state of the filter shows only verified and official apps.
- [x] Tested the filter functionality to ensure users can toggle the display of unverified and unofficial apps.


vyncent-t:
-- the last push
- have options to filter by verified, official, and cncf (commented out verified and official)
- sets default search params to be true for verified and official
- note: any non verified or official charts will not display, to access them turn on commented buttons and toggle

vyncent-t:
currently the issue is that we do not render anything that does not match the search parameters (that are currently set on a default) so charts that are not verified or official do not show up

this is tricky as we use the facets object from the api response to use for our displayed charts and pages navigation

vyncent-t:
### Previous push
## Fixed
- [x] fixed issue where toggle would reset on leaving page
- [x] fixed issue where local storage would not save
- [x] Added table for additional plugin options
- [x] Added working switch for "show verified" items

## Next
- [ ] need to fix issue where last page would not render correct items (renders empty error message)
- [ ] need to fix or add a redirect link

vyncent-t:
### Previous push

## Fixed

- [x] Fixed last page not rendering the items for default
- [x] Fixed total number of items to accurate helm items

## To do
- [ ] Need to fix last page of filter search (currently using filter options like Networking and CNCF does not display the right number of items for that criteria)

vyncent-t:
### Previous push

## Fixed

- [x] **! Refactoring the way pages are set and how we find totals for items** 
- Before we were taking the total number of all the different kinds in facets and adding them together to get the total of items and not using the number of helm items that are available like on the main site
- [x] **! NICE Fixed total number of items to accurate helm items (USING THE HEADER TO FIND THE TOTAL NUMBER OF ITEMS)**

- [ ] MOSTLY FIXED Need to fix last page of filter search (DOUBLE LOADS ON OPTIONS CHANGE AND DOES NOT KEEP UP WITH OPTIONS CHANGE)

## To do

- [ ] Need to fix new offset setting
- [ ] NEED Fix last page not rendering the items for default

vyncent-t:
### Prev push feat progress

## Done
- [x] Add section for settings using name value table in the extra settings
- [x] Add hover tool tip for show verified
- [x] Add a link to the settings page from the catalog page
- [x] Remove buttons from catalog view
- [x] Remove CNCF filter options from logic

vyncent-t:
### Current screenshots

Main Catalog Page

image
image

When clicking settings link

image
image

---

App-Catalog Settings Page

image

app-catalog/src/api/charts.tsx
```diff
@@ -1,23 +1,39 @@
...
+  // note: we are currently defaulting to search by verified and official as default
+
   if (!category || category.value === 0) {
-    return fetch(
+    const response = await fetch(
       `https://artifacthub.io/api/v1/packages/search?kind=0&ts_query_web=${search}&sort=relevance&facets=true&limit=${limit}&offset=${
         (page - 1) * limit
-      }`
-    ).then(response => response.json());
+      }&verified_publisher=${verified}`
```
sniok: There's a way to work links that's a bit easier to manage. You can use URL object like this 

js
```
const url = new URL('https://artifacthub.io/api/v1/packages/search');
url.searchParams.set('kind', '0');
url.searchParams.set('ts_query_web', search);
url.searchParams.set('sort', 'relevance');
url.searchParams.set('facets', 'true');
url.searchParams.set('limit', limit);
url.searchParams.set('offset', (page - 1) * limit);
url.searchParams.set('verified_publisher', verified);
```
and then you can use url as a string `fetch(url.toString())`

app-catalog/src/api/charts.tsx
```diff
@@ -1,23 +1,39 @@
...
+    );
+
+    const total = response.headers.get('pagination-total-count');
+    const dataResponse = await response.json();
+
+    return { dataResponse, total };
+  } else {
```
sniok: these two branches `if` `else` look almost identical, can you make it without `if`?

app-catalog/src/components/charts/List.tsx
```diff
@@ -47,45 +47,24 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
-        );
-        setChartsTotalCount(totalCount);
-        return;
-      }
-      const totalCount = categoryOptions.find(
-        (option: { name: string; total: number }) => option.name === chartCategory?.title
-      ).total;
-      setChartsTotalCount(totalCount);
+    const showVerified = JSON.parse(localStorage.getItem('showVerified') ?? 'true');
+    fetchCharts(search, showVerified, chartCategory, page).then(response => {
```
sniok: fetch inside fetchCharts might throw an error, we should add error handling

app-catalog/src/components/charts/List.tsx
```diff
@@ -140,14 +119,41 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+            color: theme.palette.mode === 'dark' ? 'white' : 'black',
+          }}
+        >
+          Settings
+        </Typography>
+      </RouterLink>
+    );
+  }
+
+  function TitleOptions() {
```
sniok: is there a reason this is a separate component?

sniok (changes requested):

app-catalog/src/api/charts.tsx
vyncent-t: I like this idea! thanks!

app-catalog/src/api/charts.tsx
vyncent-t: cleaned it!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: Yes! The title is meant to have the settings link next to it, if we use it within the extra options on the page it renders on the right along with the search and category options.

vyncent-t:
## Push note
- refactoring improvements

app-catalog/src/settings.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { useState } from 'react';
+
+export function AppCatalogSettings() {
+  const [showVerified, setShowVerified] = useState<boolean>(() =>
+    JSON.parse(localStorage.getItem('showVerified') ?? 'true')
+  );
+
+  function toggleShowVerified() {
+    setShowVerified(!showVerified);
+    localStorage.setItem('showVerified', `${!showVerified}`);
```
joaquimrocha: Please check the change-logo example plugin to see how it's using a provided function for setting/getting settings:
```typescript
export const store = new ConfigStore<pluginConfig>('change-logo');
```

app-catalog/src/api/charts.tsx
```diff
@@ -1,23 +1,38 @@
...
+  url.searchParams.set('deprecated', 'false');
+  url.searchParams.set('verified_publisher', verified.toString());
+
+  const response = await fetch(url.toString());
+
+  console.log('url search', url.toString())
+  console.log('api response', response)
+  const total = await response.headers.get('pagination-total-count');
+  console.log('api response', response.headers)
+  console.log('api total', total)
```
joaquimrocha: Please remove logs.

app-catalog/src/components/charts/List.tsx
```diff
@@ -47,45 +47,29 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+    setCharts(null);
+    const showVerified = JSON.parse(localStorage.getItem('showVerified') ?? 'true');
+    fetchCharts(search, showVerified, chartCategory, page)
+      .then(response => {
+        const packages = response.dataResponse.packages;
+        const total = parseInt(response.total);
+        setCharts(packages);
+        // note: the total number of charts is located by the helm chart category with the id of 0 within "kind" facet
+        setChartsTotalCount(total);
+        console.log('all charts', total)
```
joaquimrocha: Remove this line.

app-catalog/src/components/charts/List.tsx
```diff
@@ -140,14 +124,41 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+  function SettingsLink() {
+    const theme = useTheme();
+
+    return (
+      <RouterLink routeName="/settings/plugins/app-catalog" tooltip="App-Catalog Settings">
+        <Typography
+          sx={{
+            size: '1rem',
+            marginLeft: '3rem',
+            color: theme.palette.mode === 'dark' ? 'white' : 'black',
```
joaquimrocha: This should be themed by default.

app-catalog/src/settings.tsx
```diff
@@ -0,0 +1,96 @@
...
+        marginTop: '3rem',
+      }}
+    >
+      <Typography variant="h5">App Catalog Settings</Typography>
+      <NameValueTable
+        rows={[
+          {
+            name: (
+              <HoverInfoLabel
+                label="Show Verified"
```
joaquimrocha: Please rename this as `Only verified`. Otherwise it seems like we will stop showing verified charts.

joaquimrocha (changes requested):

vyncent-t:
moving to WIP until solution for the response header proxy is in

vyncent-t:
Notes for the pagination issue:

When running 'dev only app' our localhost is set to 3000 via ELECTRION_ and we hit a cors wall which does not provide the header for pagination total count

If the app mode is ran with 'npm run dev' the origin is not set to localhost 3000 and cors does not apply
(This can be done by referencing the Network tab and looking for the 'origin' field under Request Headers)
* note that the X-Cache reads 'Miss from cloudfront' on this example where the pagination response header is not reached
image



This means it will not work with 'npm run start' and 'npm run dev only app' but for production and 'npm run dev' it will

app-catalog/src/settings.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: Done!

app-catalog/src/components/charts/List.tsx
vyncent-t: Done!

app-catalog/src/api/charts.tsx
vyncent-t: Done!

app-catalog/src/settings.tsx
vyncent-t: Done!

vyncent-t:
When reviewing the app catalog pages should be somewhere in the 500s meaning the default off a fresh start is set to true, if it is in the 1000s this means the default is set to false for new starts (which needs to be set to true and there is an error somewhere)

skoeva (commented):
The toggle is working great as expected, though is the default boolean value deterministic? Curious if this is not the case

app-catalog/dist/main.js
sniok, line 1: gitignore this

app-catalog/src/api/charts.tsx
```diff
@@ -1,23 +1,36 @@
...
+  url.searchParams.set('sort', 'relevance');
+  url.searchParams.set('deprecated', 'false');
+  url.searchParams.set('verified_publisher', verified.toString());
+
+  const response = await fetch(url.toString());
+  const total = response.headers.get('pagination-total-count');
+
+  const dataResponse = await response.json();
+
+  console.log('api response', response);
```
sniok: 👀

app-catalog/src/components/charts/List.tsx
```diff
@@ -16,15 +17,21 @@ import {
...
 } from '@mui/material';
 import { Autocomplete, Pagination } from '@mui/material';
 import { useEffect, useState } from 'react';
 //import { jsonToYAML, yamlToJSON } from '../../helpers';
 import { fetchChartsFromArtifact } from '../../api/charts';
-import CNCFLight from './cncf-icon-color.svg';
 //import { createRelease } from '../../api/releases';
 import { EditorDialog } from './EditorDialog';
 
+interface AppCatalogShow {
```
sniok: AppCatalogConfig would be a more fitting name for this

app-catalog/src/components/charts/List.tsx
```diff
@@ -39,53 +46,57 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
   const [page, setPage] = useState(1);
   const [chartsTotalCount, setChartsTotalCount] = useState(0);
   const [chartCategory, setChartCategory] = useState(helmChartCategoryList[0]);
   const [search, setSearch] = useState('');
   const [selectedChartForInstall, setSelectedChartForInstall] = useState<any | null>(null);
 
+  // note: since we default to true for showVerified and the settings page is not accessible from anywhere else but the list comp
+  // we must have the default value here and have it imported for use in the settings tab
+  const config = store.get();
+  const [currentConfig, setCurrentConfig] = useState(config);
```
sniok: this is not needed, we can use the hook provided by the ConfigStore
`const config = store.useConfig()`

if you want a default value for showVerified you co do it like so 

`const showVerified = config?.showVerified ?? true;`

then you can also remove the useEffect on line 62

app-catalog/src/settings.tsx
```diff
@@ -0,0 +1,106 @@
+import { HoverInfoLabel, NameValueTable } from '@kinvolk/headlamp-plugin/lib/components/common';
+import { Box, Switch, SwitchProps, Typography, useTheme } from '@mui/material';
+import { useState } from 'react';
+import { store } from './components/charts/List';
+
+export function AppCatalogSettings() {
+  const config = store.get();
```
sniok: same here, use `store.useConfig()`

app-catalog/src/components/charts/List.tsx
```diff
@@ -39,53 +46,57 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+    setCharts(null);
+    const showVerified = currentConfig?.showVerified;
+
+    async function fetchData() {
+      try {
+        const response: any = await fetchCharts(search, showVerified, chartCategory, page);
+        const packages = response.dataResponse.packages;
+        const total = parseInt(response.total);
+        setCharts(packages);
+        // note: the total number of charts is located by the helm chart category with the id of 0 within "kind" facet
```
sniok: I don't understand this comment at all, what does it mean?

sniok (changes requested):

app-catalog/dist/main.js
vyncent-t: done!

app-catalog/src/api/charts.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: using the store.useConfig() here seems to not work here

app-catalog/src/settings.tsx
vyncent-t: using the store.useConfig() here seems to not work here also

app-catalog/src/components/charts/List.tsx
vyncent-t: this is an old note removing it now

app-catalog/src/components/charts/List.tsx
vyncent-t: had to rework it a little

app-catalog/src/settings.tsx
vyncent-t: I think this should work with the changes as the list exports store into the settings which can be changed here, not sure if the creation of another `const store = new ConfigStore<AppCatalogConfig>('app-catalog');` would have issues as it would potentially creating another store item by app-catalog for each list and settings interactin?

app-catalog/src/components/charts/List.tsx
sniok: I just checked the code for the useConfig and it returns a function, a bit counterintuitive, so the usage would be:

```js
// outside the component
const useStoreConfig = store.useConfig();

// inside component
function ChartsList() {
  ...
  const store = useConfigStore()
  ...
```

app-catalog/src/components/charts/List.tsx
```diff
@@ -16,15 +17,22 @@ import {
...
 import { Autocomplete, Pagination } from '@mui/material';
 import { useEffect, useState } from 'react';
 //import { jsonToYAML, yamlToJSON } from '../../helpers';
 import { fetchChartsFromArtifact } from '../../api/charts';
-import CNCFLight from './cncf-icon-color.svg';
 //import { createRelease } from '../../api/releases';
 import { EditorDialog } from './EditorDialog';
 
+interface AppCatalogConfig {
+  showVerified: boolean;
```
illume: please add docs for this field

app-catalog/src/components/charts/List.tsx
```diff
@@ -39,53 +47,43 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+  useEffect(() => {
+    setCharts(null);
+    const showVerified = config?.showVerified ?? true;
+
+    async function fetchData() {
+      try {
+        const response: any = await fetchCharts(search, showVerified, chartCategory, page);
+        const packages = response.dataResponse.packages;
+        const total = parseInt(response.total);
+        setCharts(packages);
```
illume: Please set this as `setCharts(response.dataResponse.packages)`. As the packages variable name doesn't add much in this case because we have setCharts for the name.

app-catalog/src/components/charts/List.tsx
```diff
@@ -39,53 +47,43 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+        setChartsTotalCount(total);
```
illume: as above... `setChartsTotalCount(parseInt(response.total));`

app-catalog/src/components/charts/List.tsx
```diff
@@ -39,53 +47,43 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+      } catch (err) {
+        console.error('Error fetching charts', err);
+        setCharts([]);
+      }
+    }
+
+    fetchData();
+  }, [page, chartCategory, search]);
```
illume: This function also uses `    const showVerified = config?.showVerified ?? true;` so it should depend on showVerified.

So if showVerified changes this function will run again and show the correct results.

app-catalog/src/components/charts/List.tsx
```diff
@@ -39,53 +47,43 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+  // note: since we default to true for showVerified and the settings page is not accessible from anywhere else but the list comp
+  // we must have the default value here and have it imported for use in the settings tab
+  const config = appCatConf() as { showVerified: boolean };
 
+  // note: This useEffect goes first to prevent error render when changing catagetory from the last page of any view
   useEffect(() => {
     setPage(1);
   }, [chartCategory, search]);
 
+  useEffect(() => {
```
illume: Please add a function name here for documentation. So I know what this function is for.

This is what it does: gets data based on the page, chartCategory, search, showVerified

app-catalog/src/components/charts/List.tsx
```diff
@@ -39,53 +47,43 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
-        (option: { name: string; total: number }) => option.name === chartCategory?.title
-      ).total;
-      setChartsTotalCount(totalCount);
-    });
-  }, [page, chartCategory, search]);
+  // note: since we default to true for showVerified and the settings page is not accessible from anywhere else but the list comp
+  // we must have the default value here and have it imported for use in the settings tab
+  const config = appCatConf() as { showVerified: boolean };
 
+  // note: This useEffect goes first to prevent error render when changing catagetory from the last page of any view
```
illume: Maybe add this comment?
`// When the users changes the chartCategory or search, then we always go back to the first page.`

app-catalog/src/api/charts.tsx
```diff
@@ -1,23 +1,32 @@
...
-    `https://artifacthub.io/api/v1/packages/search?kind=0&ts_query_web=${search}&category=${
-      category.value
-    }&sort=relevance&facets=true&limit=${limit}&offset=${(page - 1) * limit}`
-  ).then(response => response.json());
+  url.searchParams.set('sort', 'relevance');
+  url.searchParams.set('deprecated', 'false');
+  url.searchParams.set('verified_publisher', verified.toString());
+
+  const response = await fetch(url.toString());
+  const total = response.headers.get('pagination-total-count');
```
illume, line 25: This does not work with `npm start` (because CORS?), but it does with with `npm run dev`.

app-catalog/src/api/charts.tsx
```diff
@@ -1,23 +1,32 @@
...
+
+  const dataResponse = await response.json();
```
illume, line 27: Please test this plugin with the built Headlamp app. Using make app-linux. Don't forget to modify app-build-manifest.json.

We have an issue here for documenting how to test plugins with the built app. https://github.com/headlamp-k8s/headlamp/issues/2218

app-catalog/src/components/charts/List.tsx
```diff
@@ -140,14 +138,41 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+          Settings
+        </Typography>
+      </RouterLink>
+    );
+  }
+
+  function TitleOptions() {
+    return (
+      <Box display="flex" justifyContent="space-between" alignItems="center">
+        <Typography variant="h4">Applications</Typography>
```
illume: Please check if this plugin has any extra react-axe a11y errors in the developer console.

I expect the heading are not correct here. Style it like a h4, but use the correct semantic heading. So the levels for the page/component are correct.

illume (changes requested):
Notes from our session.

app-catalog/src/components/charts/List.tsx
```diff
@@ -16,15 +17,22 @@ import {
...
-import CNCFLight from './cncf-icon-color.svg';
 //import { createRelease } from '../../api/releases';
 import { EditorDialog } from './EditorDialog';
 
+interface AppCatalogConfig {
+  showVerified: boolean;
+}
+
+export const store = new ConfigStore<AppCatalogConfig>('app-catalog');
+const appCatConf = store.useConfig();
```
sniok: appCatConf is a hook, hooks names have to start with `use`. something like `useAppCatalogConfig`

app-catalog/src/api/charts.tsx
vyncent-t: works with the installed linux app running ./headlamp

app-catalog/src/components/charts/List.tsx
vyncent-t: Changed to style and no errors in the dev inspect

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

sniok: Documentation for the field goes in the comment above it

```ts
interface AppCatalogConfig {
  /**
   * Show only verified packages. If set to false shows all the packages
   */
  showVerified: boolean;
}
```

Also comments shouldn't repeat information that is already there (like the type) or implementation details (usage in API calls)

vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: forgot to repost my comment I made in the sibling comment but I am using what we have elsewhere in the project for this implementation 
image
image

sniok (approved):
looks good

vyncent-t:
@illume @sniok 

After running the plugin in a new env I noticed that the feature does not default to true when loaded up for the first time. This is a main point of the feature for security reasons that it would need to default to the showVerified being true for new users (they can choose to disable it in the settings later). I am rolling back a few changes that clash with the intended functionality.

the changes for `store.useConfig()` does not provide the fields needed to extract the showVerified boolean, where as `store.get()` does provide an empty object on first load that when paired with the useEffect on page load would set the store to showVerified default to true. I am also removing the previous showVerified usage that is tied to a useEffect watch parameter as it would kick off an unneeded page reload for first time users when the showVerified is set to true.

The list view is a sperate page from the settings view, which does nothave a way for users to manually change the showVerified option from the list page (they would need to leave the page, click settings to nav them to the settings page, and then nav back to the list page, thus making the reloading of the useEffect on the List view whenever the showVerified changes also unneeded)

vyncent-t:
Trying to use the const `config = store.useConfig` and then trying to use config.showVerified does not work and will constantly return empty, this will always set the verified boolean to true in this example below

image


image
image

on the first box, they both default to true
the middle box is when users switch to true to false
the last box is when returning to the page, the config is not saved and config.showVerified will never exist

This is what I have previous minus the use of the empty useEffect hook. It still functions as intended and changes the boolean as needed from the settings.

vyncent-t:
closed the outdated ones

app-catalog/src/components/charts/List.tsx
illume: This wasn't done.

app-catalog/src/components/charts/List.tsx
illume: This now has a name of `storeConfig` but it is a hook, so should have the "use" in front. Probably `useStoreConfig` is a good name.

app-catalog/src/components/charts/List.tsx
```diff
@@ -16,15 +17,25 @@ import {
...
 import { fetchChartsFromArtifact } from '../../api/charts';
-import CNCFLight from './cncf-icon-color.svg';
 //import { createRelease } from '../../api/releases';
 import { EditorDialog } from './EditorDialog';
 
+interface AppCatalogConfig {
+  /**
+   * Show only verified packages. If set to false shows all the packages
+   */
+  showVerified: boolean;
```
illume: `showVerified` isn't accurate as a name. Really it is `showOnlyVerified`. Can you please change it?

app-catalog/src/components/charts/List.tsx
```diff
@@ -166,150 +204,177 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
-                            icon="mdi:check-decagram"
-                            style={{
-                              marginLeft: '0.5em',
-                              fontSize: '22px',
-                            }}
-                          />
-                        </Tooltip>
-                      )}
+                      <CardMedia
+                        image={`https://artifacthub.io/image/${chart.logo_image_id}`}
```
illume, line 216: This CardMedia is missing an alt attribute description for a11y.

app-catalog/src/components/charts/List.tsx
```diff
@@ -166,150 +204,177 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
                         style={{
                           overflow: 'hidden',
                           textOverflow: 'ellipsis',
                           whiteSpace: 'nowrap',
                         }}
                       >
-                        <Tooltip title={chart?.repository?.name || ''}>
-                          <span>{chart?.repository?.name || ''}</span>
+                        <Tooltip title={chart.name}>
+                          <Typography component="h5" variant="h5">
```
illume: Is h5 appropriate here for accessibility?

app-catalog/src/components/charts/List.tsx
```diff
@@ -166,150 +204,177 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+                        <Box
+                          marginLeft={1}
+                          style={{
+                            overflow: 'hidden',
+                            textOverflow: 'ellipsis',
+                            whiteSpace: 'nowrap',
+                          }}
+                        >
+                          <Tooltip title={chart?.repository?.name || ''}>
+                            <span>{chart?.repository?.name || ''}</span>
```
illume: Please use a mui component, not span?

illume (approved):
Left a few more notes.

illume (changes requested):
Requested changes... see notes. 

(Clicked approve before by accident, but can't edit it.)

app-catalog/src/components/charts/List.tsx
```diff
@@ -140,14 +145,47 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
...
+            color: theme.palette.text.primary,
+          }}
+        >
+          Settings
+        </Typography>
+      </RouterLink>
+    );
+  }
+
+  function AppCatalogTitle() {
```
illume: Can you please add stories for at least some of the main changes in this PR? AppCatalogTitle.tsx and AppCatalogTitle.stories.tsx, in a separate commit.

app-catalog/.gitignore
```diff
@@ -5,6 +5,8 @@
 /.pnp
 .pnp.js
 
+/dist
```
illume, line 8: This could be in a separate commit.

app-catalog/src/components/charts/List.tsx
```diff
@@ -318,7 +383,7 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
             size="large"
             shape="rounded"
             page={page}
-            count={Math.floor(chartsTotalCount / PAGE_OFFSET_COUNT_FOR_CHARTS)}
+            count={Math.ceil(chartsTotalCount / PAGE_OFFSET_COUNT_FOR_CHARTS)}
```
illume, line 358: This bug fix could be in a separate commit.

illume (commented):
I noticed a couple of things that could be in separate commits and left notes for them.

Maybe you know of some other things that can be atomic commits as well?

app-catalog/.gitignore
vyncent-t: new pr needed

app-catalog/src/components/charts/List.tsx
vyncent-t: new pr needed

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
```diff
@@ -140,14 +145,47 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
     );
   }
 
+  function SettingsLink() {
```
illume: SettingsLink could be added in a separate commit in a SettingsLink.tsx file. With a SettingsLink.stories.tsx

illume (commented):
Left some notes about adding these components in separate files and in separate commits.

app-catalog/src/settings.tsx
```diff
@@ -0,0 +1,106 @@
...
+              <EnableSwitch checked={!!currentConfig?.showOnlyVerified} onChange={toggleShowOnlyVerified} />
+            ),
+          },
+        ]}
+      />
+    </Box>
+  );
+}
+
+const EnableSwitch = (props: SwitchProps) => {
```
illume: EnableSwitch.tsx and EnableSwitch.stories.tsx in a separate commit.

app-catalog/src/settings.tsx
```diff
@@ -0,0 +1,106 @@
+import { HoverInfoLabel, NameValueTable } from '@kinvolk/headlamp-plugin/lib/components/common';
+import { Box, Switch, SwitchProps, Typography, useTheme } from '@mui/material';
+import { useState } from 'react';
+import { store } from './components/charts/List';
+
+export function AppCatalogSettings() {
```
illume: AppCatalogSettings.tsx and AppCatalogSettings.stories.tsx in a separate commit.

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/components/charts/List.tsx
vyncent-t: done!

app-catalog/src/settings.tsx
vyncent-t: done!

app-catalog/src/settings.tsx
vyncent-t: done!

illume:
hey hey.

I just noticed the commit messages messages need mending:
- `AppCatalog: Fixes stories for AppCatalogTitle and SettingsLink`
- `frontend: `

Use `app-catalog:` and the component(s) as the context.
Use a singular verb to start it. Fix, not Fixes.

`app-catalog: ChartsList: Fix verified publisher bug`
`app-catalog: ChartsList.stories: Add verified publisher states`

vyncent-t:
> hey hey.
> 
> I just noticed the commit messages messages need mending:
> 
> * `AppCatalog: Fixes stories for AppCatalogTitle and SettingsLink`
> * `frontend: `
> 
> Use `app-catalog:` and the component(s) as the context. Use a singular verb to start it. Fix, not Fixes.
> 
> `app-catalog: ChartsList: Fix verified publisher bug` `app-catalog: ChartsList.stories: Add verified publisher states`

Should be the right way now!

illume:
> Should be the right way now!

Awesome & thanks.

app-catalog/src/components/charts/AppCatalogTitle.tsx
```diff
@@ -0,0 +1,22 @@
+import {
+    Box,
+    Typography,
```
sniok: Can you check your prettier plugin is working? This formatting looks wrong

app-catalog/src/components/charts/ChartsList.stories.tsx
```diff
@@ -32,27 +32,67 @@ const mockCharts = [
...
+      name: 'MockRepo2y',
+      url: 'https://example2y.com',
+      verified_publisher: true,
+    },
+  },
 ];
 
-const initialState = {
+// IF SHOW ONLY VERIFIED IS FALSE RETURN ALL THE CHARTS IF IT IS TRUE RETURN ONLY THE CHARTS THAT HAVE VERIFIED PUBLISHER
+// TO DO: ADD VERIFIED PUBLISHER IN THE RIGHT PLACES
```
sniok: are these todo's still valid?

app-catalog/src/components/charts/List.tsx
```diff
@@ -21,13 +22,23 @@ import { Autocomplete, Pagination } from '@mui/material';
...
+  showOnlyVerified: boolean;
+}
+
+export const store = new ConfigStore<AppCatalogConfig>('app-catalog');
+const useStoreConfig = store.useConfig();
+
 export const PAGE_OFFSET_COUNT_FOR_CHARTS = 9;
 
-export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
+export function ChartsList({ fetchCharts = fetchChartsFromArtifact }, props) {
```
sniok: React components only have one parameter. That second `props` parameter looks wrong

sniok (changes requested):

app-catalog/src/components/charts/ChartsList.stories.tsx
vyncent-t: Done!

app-catalog/src/components/charts/List.tsx
vyncent-t: ah this was left from experimenting with storybook

app-catalog/src/components/charts/AppCatalogTitle.tsx
vyncent-t: forgot to comment but done!

app-catalog/.gitignore
illume: This is still to do.

app-catalog/src/components/charts/List.tsx
illume: Also this still to do.

illume (approved):
👍 thanks

PR #48, opened 2024-07-01
workflow: Fix github workflow

sniok:
Now runs checks for each plugin folder

### DIFF
--- a/.github/workflows/headlamp-plugin-github-workflow.yaml
+++ b/.github/workflows/headlamp-plugin-github-workflow.yaml
@@ -2,17 +2,27 @@ name: Headlamp plugin linting, type checking, and testing
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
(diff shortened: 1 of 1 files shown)


illume:
It would be good if new plugins added in PR would also be tested without having to remember to add each new plugin here.

.github/workflows/headlamp-plugin-github-workflow.yaml
```diff
@@ -22,15 +20,16 @@ jobs:
     strategy:
       matrix:
         node-version: [18.x]
+        plugin-name: [app-catalog, prometheus]
```
joaquimrocha: Yeah, maybe we can do what @illume was suggesting by running the steps below after navigating through each result of `grep -m1 -lnR @kinvolk/headlamp-plugin ./*/package.json` and running on the resulting paths?

sniok:
the workflow now checks all the plugins without hardcoded list

PR #49, opened 2024-07-15
Add ArtifactHub repo file

joaquimrocha:
First step to start making plugins available on ArtifactHub.


yolossn (changes requested):
We also have to `artifacthub-pkg.yaml` for each plugin. For example [refer](https://github.com/yolossn/headlamp-plugins/blob/main/prometheus/artifacthub-pkg.yml).

yolossn (approved):

PR #51, opened 2024-07-18
plugin(opencost): Remove main.js

yolossn:
this patch removes the main.js file and updates
gitignore.


joaquimrocha:
We needed main.js to be in the final tarball but not in git.
So I added those changes.

PR #68, opened 2024-07-25
app-catalog: Use the getHeadlampAPIHeaders from headlamp-plugin/lib

joaquimrocha:
Instead of relying on the URL search params.


illume:
This fixes 
- https://github.com/headlamp-k8s/headlamp/issues/2203

PR #71, opened 2024-08-06
opencost: Fix the Details View Header Action

yolossn:
the header action was breaking the app and it
is fixed now.

before:

https://github.com/user-attachments/assets/2f599f23-116e-482a-9838-4c15c0aa500d

after:

https://github.com/user-attachments/assets/fe2c62fc-0c50-4128-964b-50b90779c8e8

### DIFF
--- a/opencost/src/index.tsx
+++ b/opencost/src/index.tsx
@@ -16,9 +16,17 @@ import { Settings } from './settings';
 import { getConfigStore, getDisplayCurrency, getDisplayTimespan, getServiceDetails } from './utils';
 
 function IconAction() {
+  const handleClick = event => {
+    event.preventDefault();
+    const targetElement = document.getElementById('opencost-plugin-cost-section');
+    if (targetElement) {
+      targetElement.scrollIntoView({ behavior: 'auto' });
+    }
+  };
+
   return (
     <Tooltip title={'Cost'}>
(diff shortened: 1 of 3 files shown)


opencost/src/index.tsx
```diff
@@ -16,9 +16,17 @@ import { Settings } from './settings';
...
+    const targetElement = document.getElementById('cost');
+    if (targetElement) {
+      targetElement.scrollIntoView({ behavior: 'auto' });
+    }
+  };
+
   return (
     <Tooltip title={'Cost'}>
-      <IconButton href="#cost" aria-label={'Cost'} aria-haspopup="true">
+      <IconButton onClick={handleClick} href="#cost" aria-label={'Cost'} aria-haspopup="true">
```
joaquimrocha: If the href still needed?

joaquimrocha: While we're still at it, can you rename the anchor ID to something more specific, to avoid clashes with other eventual plugin generated UI? something like `opencost-plugin-cost-section`?

joaquimrocha (changes requested):

opencost/src/index.tsx
yolossn: Fixed it

Issue #24, opened 2023-12-15
prometheus: make backend configurable for now admin users


farodin91:
It would great can proxy, they know the the endpoint.

In my case, i would like to provide the endpoint by an different plugin which serves dynamic clusters.

yolossn:
Hey @farodin91 Can you explain a bit more on this? Do you want a way to specify the proxy URL for accessing prometheus?

farodin91:
@yolossn I updated the issue.

remram44:
I am running into this too. My cluster already has a Prometheus server installed, I don't want to deploy another one.

In addition, the message is not very clear: "Install Prometheus for accessing metrics charts" Does this mean installing a Prometheus server or the Prometheus plugin for Headlamp?

remram44:
In addition, note that Prometheus might be using HTTPS.

yolossn:
Hey @remram44 We are working on plugin settings feature which will allow plugins to make certain values configurable by the end user. The PR is under review now, once that is done, we will update the prometheus plugin to allow the user configure the endpoint to access prometheus.

remram44:
I am running headlamp as a web app for my cluster's users, in my case they don't need to configure this value, I would do it when installing. For example an environment variable would be perfect.

air3ijai:
We are running [VictoriaMetrics](https://victoriametrics.com/) as an drop-in replacement for Prometheus.

It would be very helpful if we would be able to set a custom endpoint in [Prometheus plugin](https://github.com/headlamp-k8s/plugins/tree/main/prometheus).

yolossn:
There can be two ways to solve this, the user can either provide the label to identify the prometheus pod and the plugin figures out how to reach the pod etc or the user can provide the namespace and pod/service url to reach the prometheus pod. 

If we choose to solve it by getting the namespace and pod/service url from the user, the following steps can be used to fix this.

1. Create a Settings component and register it using the [registerPluginSettings](https://headlamp.dev/docs/latest/development/api/modules/plugin_registry#registerpluginsettings) function.
The Settings component should take care of getting the namespace and service to access prometheus as inputs and store it in plugin conf.

2. In the components that use `fetchMetrics` get the config using the `ConfigStore` and generate the prefix and pass it to the `fetchMetrics` function.


References: 
1. [Introduction to Plugin Settings](https://headlamp.dev/blog/2024/05/27/plugin-settings)
[...]

Issue #37, opened 2024-04-20
Prometheus Plug-In: Also detect victoriametrics


micw:
Hello,

I run victoriametrics in my stacks (as replacement of prometheus). It should provide the same information (at least with lens it does) but it's not detected by the plug-in.

Could detection of victoriametrics be added? What is the exact mechanism to check if prometheus is available in the cluster or not?

joaquimrocha:
@yolossn Can you check this one?

yolossn:
The plugin searches the cluster for pods with the `app.kubernetes.io/name=prometheus` label to identify the prometheus pod to query the metrics. can you check if your prometheus pod has this label?

> kubectl get pods -l app.kubernetes.io/name=prometheus --all-namespace

We are working on making the prometheus url configurable so that the user can configure the prometheus endpoint to fetch the metrics from.

avanier:
Just chiming in to say I've had a look into this, and it's not possible to overwrite that annotation if you're deploying VictoriaMetrics through the the VM operator _(which is currently the recommended way to deploy this)_.

I'm assuming it's mostly a question of extending [this code](https://github.com/headlamp-k8s/plugins/blob/main/prometheus/src/request.tsx#L7) to support the relevant VictoriaMetrics targets such as `vmsingle` and `vmselect`.

vaskozl:
The new settings panel does allow configuring the endpoint to e.g. `victoria-metrics/vmsingle-stack:8429` but it needs to be configured per client as the settings are local when running in-cluster. There is also a chance they get cleared.

It would be better for users if this could be set with an env var or similar. As I result I have opted to use a auxiliary svc that creates the expected labels.

```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/component: server
    app.kubernetes.io/name: prometheus
  name: prom-adapter
  namespace: victoria-metrics
spec:
  ports:
  - name: http
    port: 8429
    protocol: TCP
    targetPort: 8429
  selector:
    app.kubernetes.io/component: monitoring
    app.kubernetes.io/name: vmsingle
  type: ClusterIP
```

iqraurooj111:
Hi all 👋

I looked into this issue and wanted to share what I found, along with a draft PR.

## Problem

[...]

PR #87, opened 2024-10-03
prometheus: Add settings to prometheus plugin

yolossn:
This patch adds settings to the prometheus plugin
allowing the users to configure the service
address, timespan, enable metrics for each
cluster.

Fixes:
https://github.com/headlamp-k8s/plugins/issues/24

Includes the fix suggested here:
https://github.com/headlamp-k8s/plugins/issues/37#issuecomment-2230345145

Steps to Test:

1. Run prometheus plugin from the `prom_settings` branch.
2. Go to settings and configure the prometheus plugin, ie choose auto detect or provide service details to reach the prometheus service in the cluster.
3. Check if the configuration is working for the cluster by going to pod/namespace detail view.


illume:
Is there an issue to link this to?

prometheus/src/chart.tsx
```diff
@@ -12,6 +55,7 @@ export interface ChartProps {
     dataProcessor: (data: any) => any[];
   }>;
   fetchMetrics: (query: object) => Promise<any>;
+  interval: string;
```
illume, line 33: Can you please document these fields?

prometheus/src/settings.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * Settings component for configuring Prometheus metrics.
+ * @param {Object} props - The properties passed to the component.
+ * @param {Object} props.data - The current configuration data.
+ * @param {Function} props.onDataChange - The function to call when the data changes.
+ * @returns {JSX.Element} The Settings component.
+ */
+export function Settings(props) {
```
illume: Can you please add a SettingsProps interface with the docs in there?

prometheus/src/settings.tsx
```diff
@@ -0,0 +1,170 @@
+import { NameValueTable } from '@kinvolk/headlamp-plugin/lib/components/common';
```
illume: Component filenames should usually be capital named the same as the component Settings.tsx. Putting them in a folder (Settings/Settings.stories.tsx, Settings/Settings.tsx etc) makes it a bit easier as the number of components grows or there begins to be a lot of files for that component.

prometheus/src/chart.tsx
```diff
@@ -3,6 +3,49 @@ import { Box, useTheme } from '@mui/material';
 import { useEffect, useState } from 'react';
 import { Area, AreaChart, Legend, ResponsiveContainer, Tooltip, XAxis, YAxis } from 'recharts';
 
+/**
+ * Calculates the time range based on the given interval.
+ * @param {string} interval - The time interval (e.g., '10m', '1h', '24h', 'week').
+ * @returns {Object} An object containing the 'from' timestamp, 'to' timestamp, and 'step' in seconds.
+ */
+export function getTimeRange(interval: string): { from: number; to: number; step: number } {
```
illume: Is there somewhere this is specified? A URL to some docs or code perhaps?

prometheus/src/chart.tsx
```diff
@@ -3,6 +3,49 @@ import { Box, useTheme } from '@mui/material';
...
+  const now = Math.floor(Date.now() / 1000);
```
illume: This function could use some tests.

prometheus/src/common.tsx
```diff
@@ -393,8 +445,13 @@ function formatBytes(bytes: number) {
   return (bytes / Math.pow(1024, i)).toFixed(2) + units[i];
 }
 
-export function CPUChart(props: { query: string; prometheusPrefix: string; autoRefresh: boolean }) {
-  const xTickFormatter = createTickTimestampFormatter();
+export function CPUChart(props: {
+  query: string;
```
illume: Can you please document these fields? CPUChart could be in a CPUChart.tsx, because in general one file one component is better.

Can you please create a CPUChartProps interface for these fields?

prometheus/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "prometheus",
+  "name": "@headlamp-k8s/prometheus",
```
illume, line 2: This will affect how this is installed.

How to handle this? At minimum a note in the release notes (or in the README?) should warn people to remove their old prometheus plugin, what the old one is called and what the new folder name is called.

This could also affect settings/permissions for running plugins.

illume:
With these changes will it work with Azure Monitoring?

illume:
This comment was made in this issue: https://github.com/headlamp-k8s/plugins/issues/24#issuecomment-1952884113
> In addition, the message is not very clear: "Install Prometheus for accessing metrics charts" Does this mean installing a Prometheus server or the Prometheus plugin for Headlamp?

Before closing #24 that should addressed IMHO.

prometheus/package.json
illume: I think the app/ bundling will need to be changed to get the new plugin name too?

prometheus/src/chart.tsx
yolossn: Not direct code but similar to the interval to time range conversions in grafana. 
https://github.com/grafana/grafana/blob/0bd0dc374a86ae5d54905fb665348831ebf8425d/packages/grafana-data/src/datetime/rangeutil.ts

prometheus/package.json
yolossn: Nope, we just provide the build of the plugin [here](https://github.com/headlamp-k8s/headlamp/blob/main/app/app-build-manifest.json) so this shouldn't affect.

prometheus/src/chart.tsx
yolossn: Done

prometheus/src/settings.tsx
yolossn: Created the SettingsProps and documented it

prometheus/src/settings.tsx
yolossn: Refactored this and other components into individual folders.

prometheus/src/chart.tsx
yolossn: added the tests.

yolossn:
> With these changes will it work with Azure Monitor? I started an issue for this here:
> 
> * [prometheus: Azure Monitoring integration #89](https://github.com/headlamp-k8s/plugins/issues/89)

Nope this is not addressed in this PR. This PR provides a way for the user to configure the plugin to fetch the prometheus metrics from a service with in the cluster.

> I think these are the related issues for this PR?
> 
> * [Prometheus Plug-In: Also detect victoriametrics #37](https://github.com/headlamp-k8s/plugins/issues/37)
> * [prometheus: make backend configurable for now admin users #24](https://github.com/headlamp-k8s/plugins/issues/24)

Yes these are related, now that we provide a way for the user to configure which service to fetch the prometheus metrics from, they should be able to point it to Victoria Metrics or other prometheus compatible service.

This PR doesn't address the requirement related to letting another plugin set the service for a cluster dynamically. (mentioned in #24).

yolossn:
> This comment was made in this issue: [#24 (comment)](https://github.com/headlamp-k8s/plugins/issues/24#issuecomment-1952884113)
> 
> > In addition, the message is not very clear: "Install Prometheus for accessing metrics charts" Does this mean installing a Prometheus server or the Prometheus plugin for Headlamp?
> 
> Before closing #24 that should addressed IMHO.

I have updated the message and added a link to the plugin config in the banner itself. If you think the update message is still not clear, please suggest one.

illume (approved):
👍 thanks folks

PR #90, opened 2024-10-22
plugin-catalog: Allow to reload plugin on install and uninstall

### LINKED ISSUE #86: Newly installed plugins not showing up in the plugins list
When installing plugins from the plugin catalog plugin, they are not showing up in the plugins list without refreshing the application.
This means users may be confused by it and lead to think nothing was installed.

**Acceptance Criteria:**
- [ ] Open the Plugin Catalog in Headlamp and install a plugin from there
- [ ] Now go to the Settings > Plugins section and verify that the plugin is showing there

ashu8912:
fixes #86

### DIFF
--- a/plugin-catalog/src/components/plugins/Detail.tsx
+++ b/plugin-catalog/src/components/plugins/Detail.tsx
@@ -118,6 +118,21 @@ export interface PurePluginDetailProps {
   onAlertClose: () => void;
 }
 
+const pluginSnackbarAction = (closeCallback: () => void) => {
+  return (
+    <>
+      <Button
+        onClick={() => {
+          window.location.reload();
+        }}
+      >
+        Activate Now
+      </Button>
+      <Button onClick={closeCallback}>Close</Button>
+    </>
+  );
+};
+
 export function PurePluginDetail({
   pluginDetail,
   currentAction,
@@ -153,7 +168,6 @@ export function PurePluginDetail({
           },
         }}
         open={!!alertMessage}
-        autoHideDuration={6000}
         onClose={onAlertClose}
         message={
           <Tooltip title={alertMessage || ''} arrow>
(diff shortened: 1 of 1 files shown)


plugin-catalog/src/components/plugins/Detail.tsx
```diff
@@ -118,6 +118,21 @@ export interface PurePluginDetailProps {
   onAlertClose: () => void;
 }
 
+const pluginActionSnackbar = (closeCallback: () => void) => {
```
joaquimrocha: Maybe this should be called `pluginSnackbarAction`?

plugin-catalog/src/components/plugins/Detail.tsx
```diff
@@ -118,6 +118,21 @@ export interface PurePluginDetailProps {
...
+  return (
+    <>
+      <Button
+        onClick={() => {
+          window.location.reload();
+        }}
+      >
+        Reload Now
+      </Button>
```
joaquimrocha: I think this may clash with what the alertMessage may be, so it feels like something we need to pass to the PurePluginDetail...

joaquimrocha (changes requested):

illume:
Please add `plugin-catalog:` in front of the commit message?

plugin-catalog/src/components/plugins/Detail.tsx
ashu8912: Updated to be rendered only for install and uninstall alert messages

plugin-catalog/src/components/plugins/Detail.tsx
ashu8912: updated

illume (approved):
👍 thanks

(I didn't test it, but it's small enough and simple enough... I think it will work)

plugin-catalog/src/components/plugins/Detail.tsx
joaquimrocha: In the end, I changed it to almost what you had originally. I think the code base is confusing, but we can address that later.

PR #91, opened 2024-10-22
backstage: Init backstage plugin

yolossn:
Adds backstage plugin for headlamp, In the details view of a resource if the resource has labels related to backstage an Icon Button is added to open the corresponding backstage service. The plugin also provides a way to configure the backstage url for each cluster. If headlamp with this plugin is used as an iframe it sends a message to the parent window with the path to redirect to for the parent to handle and redirect. 

Steps to test (Without running backstage):
1. Run the following command to create resources with backstage related labels
> kubectl apply -f https://raw.githubusercontent.com/backstage/backstage/refs/heads/master/plugins/kubernetes-backend/examples/dice-roller/dice-roller-manifests.yaml

2. Run the backend,frontend and the backstage plugin.
3.  Navigate to the plugin settings and configure URL and configure the value for the cluster to say `http://google.com`
image
4. Navigate to the detail view of any of the dice-roller resource, Click on the backstage button.
image
5. Click on the backstage button a new window should be opened with the following url
`http://google.com/catalog/default/component/dice-roller`


joaquimrocha:
@yolossn , PTAL at my changes. I added the Backstage official logo too, but I am not entire sure. If we keep it, then please add it locally, for air-gapped envs: https://icon-sets.iconify.design/simple-icons/backstage/
Ashu can help if you need.

yolossn:
> @yolossn , PTAL at my changes. 

Looks good to me, there were a couple of lint errors, I have fixed them. 

> I added the Backstage official logo too, but I am not entire sure. If we keep it, then please add it locally, for air-gapped envs: https://icon-sets.iconify.design/simple-icons/backstage/ Ashu can help if you need.

Sorry I didn't get why this will be an issue for simple-icons? We use mdi icons in a similar fashion right.

backstage/README.md
```diff
@@ -0,0 +1,19 @@
+# Backstage
+
+The Backstage plugin adds a Open with Backstage button to the resource detail view. If
+the plugin detects that the page is inside a Backstage iframe, it will send a
+message to the parent page to open the resource in the Backstage UI.
+
+This is the default template README for [Headlamp Plugins](https://github.com/headlamp-k8s/headlamp).
```
illume: This README need mending below here (this text is from the default template).

Probably a quickstart on how to get started with it.

yolossn: Added steps to run in dev mode and steps to build.

joaquimrocha (approved):

Issue #92, opened 2024-10-29
app-catalog broken


skoeva:
Opening app-catalog causes the app to crash:

![image](https://github.com/user-attachments/assets/19bb0501-960b-4b4d-959c-1c0dd70b4368)

Here are the console logs, looks like a reference to an undefined parameter is causing the issue:

![image](https://github.com/user-attachments/assets/5d6203d0-3918-4d6d-8bae-6a02192bebd1)

PR #93, opened 2024-10-29
app-catalog: Remove unused props

skoeva:
This change removes a reference to undefined props which is causing the app-catalog to crash.

Fixes: #92 

### Testing
- [X] Open the app catalog within Headlamp on the main branch and note the crash:

image

- [X] Open the app catalog within Headlamp on the fix-broken-charts branch and ensure it doesn't crash:

image


illume (approved):
👍 thanks for fixing the crash

Issue #36, opened 2024-04-16
Show YAML based on the Helm version


joaquimrocha:
A user has reported that when we are installing an app (Helm), changing the version doesn't seem to update the YAML being displayed, which means that it may end up trying to install a YAML that is not compatible with that version of the chart.

Acceptance Criteria:
- [ ] When changing the version of the chart in the Install dialog, the YAML is updated to the one corresponding to that version

PR #94, opened 2024-10-30
app-catalog: Update YAML based on Helm version

skoeva:
This change updates the YAML in the install dialog based on the Helm version of the chart, with the latest version set to display by default. For clarity, the version number is now displayed in the dialog rather than being flushed by the onMount function.

Fixes: #36 

### Testing:
- [X] Open the app catalog within Headlamp and navigate to the install dialog
- [X] Ensure the latest version appears at the top, and that changing the version updates the YAML

image


app-catalog/src/components/charts/EditorDialog.tsx
```diff
@@ -67,14 +67,23 @@ export function EditorDialog(props: {
...
-        setVersions(
-          response.available_versions.map(({ version }) => ({ title: version, value: version }))
-        );
+        const versionOptions = response.available_versions.map(({ version }) => ({ 
+          title: version, 
+          value: version 
+        }));
+        setVersions(versionOptions);
+        setSelectedVersion(versionOptions[0]);
+        handleChartValueFetch(chart);
```
joaquimrocha: This is repeated in te effect below that reacts to the selectedVersion. Maybe it's not needed here, but do check it after the other commit with the default values changes please.

joaquimrocha (changes requested):

illume:
There's a conflict now the other commit was merged, and I guess that's what the change request from J is related to.

illume (approved):
👍 thanks

PR #95, opened 2024-10-30
plugin-catalog: Replace "Activate Now" by "Reload Now"

joaquimrocha:
This is also shown when the plugin is removed.


sniok (approved):
nice

Issue #44, opened 2024-06-11
Add default values to the top options when installing apps/helm charts


joaquimrocha:
The top entries of the dialog when installing an app in the app catalog should get default values when we can. E.g. the version should be the latest one by default (and there should always be one selected).
The description should have something like `[APP NAME HERE] deployment` as a default value.

PR #96, opened 2024-10-30
app-catalog: Add default values to options

skoeva:
This change adds default values to the top options in the install UI. The namespace is set initially set to default, the version to the latest version for the respective app, and the description to [app name] deployment.

Fixes: #44 

### Testing
- [X] Open the app catalog within Headlamp and navigate to the install dialog
- [X] Ensure the top fields are filled with default values

image

### DIFF
--- a/app-catalog/src/components/charts/EditorDialog.tsx
+++ b/app-catalog/src/components/charts/EditorDialog.tsx
@@ -11,6 +11,11 @@ import { createRelease, getActionStatus } from '../../api/releases';
 import { addRepository } from '../../api/repository';
 import { jsonToYAML, yamlToJSON } from '../../helpers';
 
+type FieldType = {
+  value: string;
+  title: string;
+};
+
 export function EditorDialog(props: {
   openEditor: boolean;
   chart: any;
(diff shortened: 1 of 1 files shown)


app-catalog/src/components/charts/EditorDialog.tsx
joaquimrocha, line 1: It says the namespace is set to default but it's setting it to the 1st namespace in the list. This is fine for now since I think we don't expose the loading of cluster settings to the plugins.

app-catalog/src/components/charts/EditorDialog.tsx
```diff
@@ -45,6 +45,12 @@ export function EditorDialog(props: {
     setIsFormSubmitting(false);
   }, [openEditor]);
 
+  useEffect(() => {
+    if (!selectedNamespace && !!namespaceNames) {
+      setSelectedNamespace(namespaceNames[0])
+    }
+  }, [namespaceNames]);
```
joaquimrocha: The `selectedNamespace` needs to be in the dependencies of this effect, else it will never use an updated value. See my commit with the version defaults.

app-catalog/src/components/charts/EditorDialog.tsx
```diff
@@ -296,8 +298,7 @@ export function EditorDialog(props: {
                 editor.focus();
                 setReleaseName('');
                 setSelectedVersion(null);
-                setSelectedNamespace(null);
-                setChartInstallDescription('');
+                setChartInstallDescription(`${chart.name} deployment`);
```
joaquimrocha: Can't this be removed too and the new default value added when we declare the effect?

joaquimrocha (changes requested):
@skoeva Thanks for the PR. I think this is not entirely doing what was intended: it does set the namespace by default but normally the UI doesn't show it as selected, which is misleading. I spent some times investigating and added a new commit for the version default which does set up the autocomplete (it's the defaultValue, and cannot just rely on the value of the version/namespace, but rather has to use the versions[0]/namespaces[0]).

Please adapt the 1st commit after this review.

app-catalog/src/components/charts/EditorDialog.tsx
skoeva: I believe it is always at the top of the list, otherwise we could just search for it in the list but this seems like an unnecessary operation

app-catalog/src/components/charts/EditorDialog.tsx
skoeva: I think your commit might actually be breaking the default version. The initial change involved setting the selected version each time we opened a different app's install dialog, and now, since we only set it once in the `useEffect` at the top, this doesn't happen, so opening a different install dialog does not change the version number.

I also looked at the docs for Autocomplete, it seems that using `defaultValue` while having a React state is redundant, so I can make the change and lmk what you think

illume (approved):
👍 thanks

PR #103, opened 2024-11-06
Bump opencost version

yolossn:


opencost/artifacthub-pkg.yml
joaquimrocha, line 1: Why no double quotes? Seems unrelated. We should be consistent with the rest of the manifests. If it's single, let's keep this, but if it's double, let's change it back. In any case it's not a part of this change.

yolossn: I didn't specifically change it, but the vscode settings that is auto generated by `headlamp-plugin` init has set `editor.formatOnSave` to `true` and it got auto formatted by the default formatter, in my case `prettier`. If we want this to not repeat we should add a preference to use double quotes in the vscode settings. I went on to find why auto formatter is using single quotes over double quotes and found that
`In yaml, It is better to use single quotes over double because double quotes parses escape codes like "\n". `

yolossn: Looks like it is not even the default formatter that is doing this but the prettier-config that we have here
https://github.com/kinvolk/eslint-config/blob/main/prettier-config/index.js

joaquimrocha (approved):

PR #109, opened 2024-11-11
flux: Show suspended in status column

### LINKED ISSUE #108: Flux: status column should show suspended status
The list pages for HelmRelease and Kustomization show a status column which supports users to quickly review the installation and check for errors. 

The current status column only shows the ready state. The suspended status of a resource is also important for users. A kustomization may be suspended by an engineer in order to make a manual fix. Often it is forgotten to resume the Kustomization after the manual operation had been done. Suspended resources deviate from the GitOps model and should be shown as a warning. 

We can fix this by showing the suspended status in the status column. Note this is the same behavior as in the good old Weave GitOps frontend.

mgalesloot:
Fixes #108

### DIFF
--- a/flux/src/common/StatusLabel.tsx
+++ b/flux/src/common/StatusLabel.tsx
@@ -1,6 +1,4 @@
-import {
-  StatusLabel as HLStatusLabel,
-} from '@kinvolk/headlamp-plugin/lib/components/common';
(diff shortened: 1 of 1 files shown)


ashu8912:
Tested, Thanks @mgalesloot

Issue #122, opened 2024-12-03
Allow reconciliation --force 


ashu8912:
Allow force reconciliation for flux resources 

ref https://fluxcd.io/flux/cmd/flux_reconcile_helmrelease/#options

siegenthalerroger:
This doesn't quite seem to work as intended by #124 

1. when you select the button it asks to "disable force reconciliation", regardless of whether it is active or not
2. it doesn't have the same effect as the cli does when force reconciling a stalled (retries exhausted) helm release as the cli does. In this case it seems to do nothing actually.

PR #124, opened 2024-12-10
flux-plugin: Enable/Disable force reconciliation

ashu8912:


ashu8912:
fixes #122

flux/src/actions/index.tsx
```diff
@@ -3,6 +3,123 @@ import { KubeObject } from '@kinvolk/headlamp-plugin/lib/lib/k8s/cluster';
...
+            });
+        }}
+        title={'Enable Force Reconciliation'}
+        description={`Are you sure you want to enable force reconciliation?`}
+      />
+    </>
+  );
+}
+
+function DisableForceReconciliation(props) {
```
joaquimrocha: Seems like this component is the same as above, just with different logic + strings based on whether it's enable/disable. I think we should just have one single component.

joaquimrocha (changes requested):
Besides making just one component, the commit should be prefixed with the plugin name `flux:`, not `flux-plugin:`.

flux/src/actions/index.tsx
```diff
@@ -3,6 +3,66 @@ import { KubeObject } from '@kinvolk/headlamp-plugin/lib/lib/k8s/cluster';
 import { useSnackbar } from 'notistack';
 import React from 'react';
 
+function ForceReconciliationAction(props) {
+  const { enqueueSnackbar } = useSnackbar();
+  const [open, setOpen] = React.useState<boolean>(false);
+  const { resource } = props;
+  if(!resource.jsonData.spec.force) {
+    return null;
+  }
```
joaquimrocha: This will prevent the component from rendering if the force is false.

flux/src/actions/index.tsx
```diff
@@ -3,6 +3,66 @@ import { KubeObject } from '@kinvolk/headlamp-plugin/lib/lib/k8s/cluster';
...
+                force: false,
+              },
+            },
+            resource.jsonData.metadata.namespace,
+            resource.jsonData.metadata.name
+          )
+            .then(response => {
+              if (!response.spec.force) {
+                enqueueSnackbar(
+                  `Successfully ${forceReconciliationType}d force reconciliation for ${resource.metadata.name}`,
```
joaquimrocha: As a general rule, never generate words for human readable messages like this. Just use `Disabled force reconciliation for ${resource.metadata.name}` and `Enabled force reconciliation for ${resource.metadata.name}`

flux/src/actions/index.tsx
```diff
@@ -3,6 +3,66 @@ import { KubeObject } from '@kinvolk/headlamp-plugin/lib/lib/k8s/cluster';
...
+                  { variant: 'success' }
+                );
+              } else {
+                enqueueSnackbar(`Failed to ${forceReconciliationType} force reconciliation for ${resource.metadata.name}`, {
```
joaquimrocha: Same about not generating english words from variables.

flux/src/actions/index.tsx
```diff
@@ -3,6 +3,66 @@ import { KubeObject } from '@kinvolk/headlamp-plugin/lib/lib/k8s/cluster';
...
+                  variant: 'error',
+                });
+              }
+            })
+            .catch(error => {
+              enqueueSnackbar(
+                `Failed to ${forceReconciliationType} force reconciliation for ${resource.metadata.name} error ${error}`,
```
joaquimrocha: `Failed to enable force reconciliation for ${resource.metadata.name}: ${error}`
and
`Failed to disable force reconciliation for ${resource.metadata.name}: ${error}`

flux/src/actions/index.tsx
```diff
@@ -3,6 +3,66 @@ import { KubeObject } from '@kinvolk/headlamp-plugin/lib/lib/k8s/cluster';
...
+                { variant: 'error' }
+              );
+            });
+        }}
+        title={`${forceReconciliationType} Force Reconciliation`}
+        description={`Are you sure you want to ${forceReconciliationType} force reconciliation?`}
```
joaquimrocha: ditto

joaquimrocha (changes requested):

Issue #115, opened 2024-11-19
Flux plugin - Onboarding Issues


kingdonb:
I'm testing the Flux plugin and I see there are some open issues already, I'm just going to open one issue and write stream-of-consciousness before I forget all these small things I noticed along the way - first an issue that might be in Headlamp proper, the Plugins Catalog was missing from my UI entirely! Until I went to the Navigation menu and did "Reload"

Then I was able to install the Flux plugin. Connecting to a cluster with Flux installed, I found that everything seems to work, except for a few small issues:

* A Kustomization which was not suspended had the "Resume" button - clicking it didn't seem to do anything except for turn it into a Suspend button, as expected. Then, the modal behavior worked. Resume/Suspend behavior was OK. This happened on a VCluster. It didn't happen on every Flux kustomization. Not sure what about this specific customization triggered the issue.
* Any Git Repository on any of my clusters failed to load in the UI. The Git Repositories on the "Sources" UI looked OK, until I clicked on the git repository's name. Then I was presented with a blank screen (and thus not able to get to the Flux UI buttons)

<img width="1439" alt="Screenshot 2024-11-19 at 9 01 56 AM" src="https://github.com/user-attachments/assets/3b151eca-8f16-4efd-ab6f-a1276db21338">

I spent most of my time pre-release testing with a deployment via Helm and init container, I'm fielding some questions about this in the Flux channel on CNCF slack today. It looks like someone ran into this issue:

* https://github.com/headlamp-k8s/plugins/issues/112

(I don't have anything to add to this myself right now, as I'm testing the native app today)

joaquimrocha:
Thanks for the report @kingdonb , we will take a look.

cc/ @ashu8912

vijayansarathy:
Hi @kingdonb 
Which desktop version of Headlamp are you using? Are you running it on a Mac.
I am really interested in using the Flux plugin but I have not been able to install any plugins at all on the Mac version. 
I had [reported this issue](https://github.com/headlamp-k8s/plugins/issues/119) back in Nov. I tried again today with v0.27.0 and encounter the same issue.

joaquimrocha:
@ashu8912 Is this fixed? @kingdonb can you confirm either way?

kingdonb:
Is what fixed? I can't say for sure. I have not been able to access my vclusters with Headlamp native app, for some reason. It will take me some time to narrow down.

[...]

PR #125, opened 2024-12-11
Flux suspend,resume fix

ashu8912:


ashu8912:
fixes first part for #115

flux/src/actions/index.tsx
```diff
@@ -8,20 +8,14 @@ function SuspendAction(props) {
   const { resource } = props;
   const [open, setOpen] = React.useState<boolean>(false);
 
+  if(resource.jsonData.spec.suspend) {
```
joaquimrocha: Do you need to check whether you have the `suspend` prop there? Because before you were checking.
Maybe `resource?.jsonData?.spec?.suspend`

joaquimrocha (changes requested):

flux/src/actions/index.tsx
ashu8912: fixed

joaquimrocha (approved):

joaquimrocha (changes requested):
Seems like the PR is also changing how Flux is being checked for installation. Can you check whether this should go into different commits?

flux/src/sources/SourceList.tsx
```diff
@@ -43,9 +45,16 @@ export default function FluxSources() {
     },
   ];
 
+  if(isFluxInstalled === null) {
+    return null;
```
joaquimrocha: Should this be a spinner?

flux/src/image-automation/ImageAutomationList.tsx
```diff
@@ -73,7 +84,14 @@ export default function ImageAutomationWrapper() {
 
   if (!isImageAutomationControllerAvailable) {
     return (
-      <SectionBox>
+      <SectionBox  style={{
```
joaquimrocha: should be `sx={{`

flux/src/image-automation/ImageAutomationList.tsx
```diff
@@ -73,7 +84,14 @@ export default function ImageAutomationWrapper() {
...
+        padding: '1rem',
+        alignItems: 'center',
+        margin: '2rem auto',
+        height: '20vh',
```
joaquimrocha: Not needed AFAIU.

flux/src/image-automation/ImageAutomationList.tsx
```diff
@@ -73,7 +84,14 @@ export default function ImageAutomationWrapper() {
...
+        width: '50%',
```
joaquimrocha: This may mess the width when in narrow screens. Please either account for smaller screens, or use `maxWidth: '600px'` for example.

flux/src/image-automation/ImageAutomationList.tsx
```diff
@@ -73,7 +84,14 @@ export default function ImageAutomationWrapper() {
...
+        backgroundColor: theme.palette.background.paper,
```
joaquimrocha: Not needed AFAIU.

joaquimrocha (changes requested):
Check out the styling changes I added comments for. They happen in more than 1 SectionBox.

flux/src/checkflux/index.tsx
```diff
@@ -43,11 +28,9 @@ export default function CheckIfFluxInstalled() {
...
       </Box>
     );
-  }
 }
 
-export function useFluxControllerAvailableCheck(props: { name: string }) {
-  const { name } = props;
+export function useFluxInstallCheck() {
   const [deployments] = K8s.ResourceClasses.Deployment.useList({
     labelSelector: 'app.kubernetes.io/part-of=flux',
```
sniok: Seems like you could use label selector with both labels here, then `deployments?.find(...` won't be neccessary?
```tsx
labelSelector: 'app.kubernetes.io/part-of=flux,app.kubernetes.io/component=source-controller',
```

flux/src/helm-releases/HelmReleaseList.tsx
```diff
@@ -21,7 +24,14 @@ export function HelmReleases() {
 
   if (!isHelmReleasesControllerAvailable) {
     return (
-      <SectionBox>
+      <SectionBox  style={{
```
sniok: sx here as well

flux/src/helm-releases/HelmReleaseList.tsx
```diff
@@ -45,9 +55,18 @@ function HelmReleasesListWrapper() {
     return helmReleases?.makeCRClass();
   }, [helmReleases]);
 
+  const isFluxInstalled = useFluxInstallCheck();
```
sniok, line 55: not critical but this repeating part could be extracted to a separate component that can be used to wrap page components

then in index.ts it would look something like this: 
```tsx
registerRoute({
  path: '/flux/helmreleases',
  parent: 'flux',
  sidebar: 'helmreleases',
  component: () => <FluxInstalledOnly><HelmReleases /></FluxInstalledOnly>,
  exact: true,
  name: 'helmreleases',
});

sniok (changes requested):
looks good, just some minor comments

flux/src/checkflux/index.tsx
joaquimrocha: Good point!

flux/src/helm-releases/HelmReleaseList.tsx
joaquimrocha: Maybe for later.

PR #126, opened 2024-12-12
Improve Prometheus detection

caellach:
The Prometheus plugin's auto-detection wasn't working for me because the port wasn't being included in the queries. Manually specifying the service did work but I don't want to re-enter that in all my various browsers. Having it 'just work' is better.

This improves Prometheus detection by searching through pods and services, testing the open ports with an 'up' query, and selecting the first working instance (similar to how it currently works in main). It also adds a custom label (`headlamp-prometheus=true`) which allows marking a pod or service as the one that this plugin should be using for the case where multiple Prometheus instances are available; this case is outside of my usage, but I'm confident that it will work.

I'm currently running this in my personal cluster without issue.

Let me know if anything in this PR needs modification and I'd be happy to accommodate.

### DIFF
--- a/prometheus/src/request.tsx
+++ b/prometheus/src/request.tsx
@@ -2,19 +2,247 @@ import { ApiProxy } from '@kinvolk/headlamp-plugin/lib';
 
 const request = ApiProxy.request;
 
-export async function isPrometheusInstalled() {
-  const queryParams = new URLSearchParams();
-  queryParams.append('labelSelector', 'app.kubernetes.io/name=prometheus');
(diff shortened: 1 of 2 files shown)


joaquimrocha:
@caellach Can you sign-off on your commits? That's done with `git commit -s`, so DCO checks pass.

sniok:
Tested by using a prometheus helm chart that's not from prometheus-community and with this change it detects it correctly. on the main branch it errors out

sniok (approved):

sniok:
Merged manually to add signoffs

PR #127, opened 2024-12-12
Fixes for app catalog

joaquimrocha:
fixes https://github.com/headlamp-k8s/headlamp/issues/2618

## Steps to reproduce:
1. Try to install any chart -> now it works

### Do not crash when clearing the categories
1. Open the app-catalog and click the X next to the All category -> no crash, the All word goes away but the apps are still there

### Editor dialog setting the release name on version change
1. Click Install on a chart, then type a release name
2. Change the version
3. Realize that the values editor was reloaded but the release name is the same

### Update editor UX: spinner
1. Go to an installed version and click Update
2. Verify that a spinner is shown until the versions are fetched

### Update editor UX: insensitive update button
1. Go to an installed version and click Update
2. When the versions are loading -> the Update button is disabled

### Update editor UX: versions' sorting
1. Go to an installed version and click Update
2. After the versions are loaded -> check that the versions are shown with the most recent at the top

### DIFF
--- a/app-catalog/src/components/charts/EditorDialog.tsx
+++ b/app-catalog/src/components/charts/EditorDialog.tsx
@@ -76,10 +76,9 @@ export function EditorDialog(props: {
     setChartInstallDescription(`${chart.name} deployment`);
     fetchChartDetailFromArtifact(chart.name, chart.repository.name).then(response => {
       if (response.available_versions) {
-        setVersions(
-          response.available_versions.map(({ version }) => ({ title: version, value: version }))
-        );
-        setSelectedVersion(response.available_versions[0].version);
+        const availableVersions = response.available_versions.map(({ version }) => ({ title: version, value: version }));
+        setVersions(availableVersions);
+        setSelectedVersion(availableVersions[0]);
       }
     });
     handleChartValueFetch(chart);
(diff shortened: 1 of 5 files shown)


app-catalog/src/components/charts/EditorDialog.tsx
```diff
@@ -76,10 +76,9 @@ export function EditorDialog(props: {
     setChartInstallDescription(`${chart.name} deployment`);
     fetchChartDetailFromArtifact(chart.name, chart.repository.name).then(response => {
       if (response.available_versions) {
-        setVersions(
-          response.available_versions.map(({ version }) => ({ title: version, value: version }))
-        );
-        setSelectedVersion(response.available_versions[0].version);
+        const availableVersions = response.available_versions.map(({ version }) => ({ title: version, value: version }));
```
ashu8912, line 79: Why this change???

app-catalog/src/components/releases/EditorDialog.tsx
```diff
@@ -1,4 +1,4 @@
-import { Dialog } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import { Dialog, Loader } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
```
ashu8912, line 1: a linter is failing here

ashu8912 (approved):
some small comments but overall tested and it works for me

app-catalog/src/components/charts/EditorDialog.tsx
joaquimrocha: Because it was setting the versions for the selectedVersion wrong. We need to set the whole object, not just the actual version.

PR #128, opened 2024-12-12
plugin-catalog: Add a switch to show/hide non-official plugins quickly

joaquimrocha:
The plugin settings hide away the possibility of seeing more plugins. These changes add a switch to the plugin catalog to display/hide non-official plugins without going to the settings.

How to test:
1. Go to the plugin catalog. Notice the new switch next to the title
2. Switch it off -> See the warning/confirmation.
3. Press No -> The confirm dialog is closed and nothing happens
4. Press Yes -> The catalog is reloaded showing a lot more plugins


ashu8912 (approved):
Tested, thanks

joaquimrocha:
> Tried it out, works great. I wish it could remember the confirm choice so it doesn't pop up every time, but it is not essential

That's on purpose 😉

PR #117, opened 2024-11-20
plugin-catalog: Rework installed list to include all plugins

vyncent-t:
## Description

Solves Issue: [#2586](https://github.com/headlamp-k8s/headlamp/issues/2586)

This PR addresses the issue observed during our ContribuFest at KubeCon, where users expressed confusion about the `Installed` view in the Plugin Catalog. Specifically, users were expecting to find their development plugins in the `Installed` view, but only plugins installed via the Plugin Catalog were listed. 

To resolve this, this PR proposes the following updates:

- **Unified Installed View**: The `Installed` view in the Plugin Catalog will now display all installed plugins, regardless of their installation source.
- **Categorization**:
  - Plugins installed via the Plugin Catalog will appear under a section titled `Installed from the Plugin Catalog`.
  - Plugins installed from other sources will appear under a section titled `Other Installed Plugins`.
- **Link to Plugin Details**: Each plugin entry will include a link to its respective details/settings page to provide more information and configuration options.

These changes aim to improve user experience by consolidating plugin information in one view while maintaining clarity about their source and functionality.

---
## Image
image

---

## Changes Made

1. Updated the Plugin Catalog `Installed` view to:
   - Display all plugins (both catalog-installed and non catalog-installed).
   - Categorize plugins into `Installed from the Plugin Catalog` and `Other Installed Plugins`.
2. Added links to plugin details/settings for each listed plugin.

---

## Steps to Test

1. Open the Plugin Catalog in the Headlamp application.
2. Install a plugin via the Plugin Catalog.
3. Add a plugin manually to simulate a non-catalog installation.
4. Navigate to the `Installed` view and confirm:
   - Plugins are categorized under `Installed from the Plugin Catalog` or `Other Plugins`.
   - Each plugin has a working link to its details/settings page.

---

## Notes

- Plugins listed under `Other Plugins` will not include uninstall functionality if they were not installed via the Plugin Catalog.

PR #129, opened 2024-12-12
plugin-catalog: Add link to all plugins from installed list

joaquimrocha:
This is related to #117 , but just adds a link to the plugin settings view which shows all installed plugins.
Until #117 is in a state we can merge, I think this is a simple enough change to mitigate the confusion between the 2 installed views.


skoeva (approved):
I think the wording can be simplified but since these aren't the final changes, that can wait. Otherwise tested and LGTM

sniok (approved):

PR #133, opened 2024-12-14
node-shell: add node shell plugin

farodin91:
image
image
image

Fixes https://github.com/headlamp-k8s/headlamp/issues/996

Possible Improvements:
* ensure pod is running
* improve error handling
* ensure pod is stopped
* add testing
* Windows Support


farodin91:
@joaquimrocha Would you like to review?

joaquimrocha (commented):
I tested it and it works very well! Thanks @farodin91 .
I am still torn between having it as a plugin or having it in the core, but since it shouldn't work in certain environments, maybe it's better as a plugin.

I also realized that we should make it easier for plugins to declare cluster-based settings. These would show up both in the plugin settings, with the cluster selector already automatically done for them, and in the cluster's own settings, as sections.

Since I am leaving on vacation for 2 weeks, I will leave a deeper review of the code to @illume .

Issue #138, opened 2024-12-23
Plugin Catalog: Available updates not clearly indicated 


mgalesloot:
The Plugin Catalog shows a panel with available plugins. In this overview it shows the **available** version for each plugin. If the user has installed this plugin already, the catalog shows "Installed" in the right bottom corner. 

The phrase "Installed" gives the average user the impression that the latest version is already installed. However this may not be the case. The user has to navigate to plugin details to get information about "Available Version" and "Installed Version". 

The catalog could be improved by showing the phrase "Update Available" instead of "Installed" if the user has the option to upgrade to a newer version. Also an icon button (arrow top) could be added to allow the user to immediately start the update proces. 

In the plugin details page, the top arrow button is currently placed in the top right corner. It could be better placed directly in the field with the Available version to make it easier to understand the purpose of this button. 

![image](https://github.com/user-attachments/assets/c80d6df8-85e8-4ea5-8d0f-e06479a66683)

![image](https://github.com/user-attachments/assets/31487db2-326c-4dc3-9188-e94cb1fd2406)

PR #141, opened 2024-12-27
plugin-catalog: Add update available label to card

skoeva:
This change adds a label in the card for plugins with available updates. The "Installed" label now gets replaced with "Update available" in this case.

Fixes: #138 

### Testing
- [X] Open Headlamp and navigate to the plugin catalog
- [X] For any plugins with updates, ensure that the card displays the "Update available" label

image


joaquimrocha (approved):
Thanks. LGTM!

PR #144, opened 2025-01-08
prometheus: Added basic authorization

vlasov-y:
Hi! 

image

This PR adds support of basic authorization for prometheus plugin. Nevertheless, there is an issue on the backend side. I have decided to show changes and describe the issue in the video below.

Video review and description of the problem
https://github.com/user-attachments/assets/ee2f14ae-117f-47ce-be2b-bfa1472357b4

### DIFF
--- a/prometheus/package-lock.json
+++ b/prometheus/package-lock.json
@@ -8,6 +8,8 @@
       "name": "prometheus",
       "version": "0.3.0",
       "dependencies": {
+        "@iconify/react": "^5.2.0",
+        "@storybook/react": "^8.4.7",
         "recharts": "^2.7.3",
         "use-between": "^1.3.5"
       },
@@ -161,6 +163,7 @@
       "resolved": "https://registry.npmjs.org/@aw-web-design/x-default-browser/-/x-default-browser-1.4.126.tgz",
       "integrity": "sha512-Xk1sIhyNC/esHGGVjL/niHLowM0csl/kFO5uawBy4IrWwy0o1G8LGt3jP6nmWGz+USxeeqbihAmp/oVZju6wug==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "default-browser-id": "3.0.0"
       },
@@ -2287,10 +2290,11 @@
       }
     },
     "node_modules/@babel/register": {
-      "version": "7.24.6",
-      "resolved": "https://registry.npmjs.org/@babel/register/-/register-7.24.6.tgz",
-      "integrity": "sha512-WSuFCc2wCqMeXkz/i3yfAAsxwWflEgbVkZzivgAmXl/MxrXeoYFZOOPllbC8R8WTF7u61wSRQtDVZ1879cdu6w==",
+      "version": "7.25.9",
+      "resolved": "https://registry.npmjs.org/@babel/register/-/register-7.25.9.tgz",
+      "integrity": "sha512-8D43jXtGsYmEeDvm4MWHYUpWf8iiXgWYx3fW7E7Wb7Oe6FWqJPl5K6TuFW0dOwNZzEE5rjlaSJYH9JjrUKJszA==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "clone-deep": "^4.0.1",
         "find-cache-dir": "^2.0.0",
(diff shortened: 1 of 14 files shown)


joaquimrocha:
@yolossn It'd great to include this in our release next week. Please let me know if you can review or if someone else should.

vlasov-y:
> @yolossn It'd great to include this in our release next week. Please let me know if you can review or if someone else should.

It has an issue that I have mentioned in the video. Seems that backend does not pass authorization header to the prometheus.

yolossn:
@vlasov-y  I looked into the auth headers not being forwarded issue that you mentioned. Looks like the kubernetes api-server is assuming the authorization header is for itself and not forwarding it to the proxied service or pod. Since the Authorization header is being removed by the kubernetes api-server we should look for an upstream fix.
cc: @joaquimrocha 

Sharing the scripts that I used to debug this for your reference

Create a http echo service.
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpecho
spec:
  replicas: 1
  selector:
    matchLabels:
      app: httpecho
  template:
    metadata:
      labels:
        app: httpecho
    spec:
      containers:
      - name: httpecho
        image: mendhak/http-https-echo:latest
        ports:
        - containerPort: 8443
---
apiVersion: v1
kind: Service
metadata:
  name: httpecho
spec:
  ports:
  - port: 80
    targetPort: 8443
  selector:
    app: httpecho
```

Once the echo service is deployed create a proxy to api server by running

```bash
kubectl proxy
```

```bash
curl -H "Authorization: Basic $(echo -n 'username:password' | base64)" \
  http://localhost:8001/api/v1/namespaces/default/services/httpecho:80/proxy/api/v1/query_range?start=1737014987&end=1737101387&step=300&query=sum%28rate%28container_cpu_usage_seconds_total%7Bcontainer%21%3D%27%27%2Cnamespace%3D%27default%27%2Cpod%3D%27httpecho-56b9f47ccb-qss6m%27%7D%5B1m%5D%29%29+by+%28pod%2Cnamespace%29
```

image


```bash
curl -H "X-Forwarded-Authorization: Basic $(echo -n 'username:password' | base64)" \
  http://localhost:8001/api/v1/namespaces/default/services/httpecho:80/proxy/api/v1/query_range?start=1737014987&end=1737101387&step=300&query=sum%28rate%28container_cpu_usage_seconds_total%7Bcontainer%21%3D%27%27%2Cnamespace%3D%27default%27%2Cpod%3D%27httpecho-56b9f47ccb-qss6m%27%7D%5B1m%5D%29%29+by+%28pod%2Cnamespace%29
```

image

vlasov-y:
@yolossn thank you for your investigation. Wdyt we have to do with this PR while problem is not fixed in k8s API? By the way, have you created an issue in k8s repo for that already?

yolossn:
> @yolossn thank you for your investigation. Wdyt we have to do with this PR while problem is not fixed in k8s API? By the way, have you created an issue in k8s repo for that already?

I am looking for alternate ways to solve this. I haven't created any issues yet. For now I am looking into the prometheus codebase to check if it accepts any other header(like X-Forwarded-Authorization) other than Authorization. Ill keep you updated.

vlasov-y:
> > @yolossn thank you for your investigation. Wdyt we have to do with this PR while problem is not fixed in k8s API? By the way, have you created an issue in k8s repo for that already?
> 
> I am looking for alternate ways to solve this. I haven't created any issues yet. For now I am looking into the prometheus codebase to check if it accepts any other header(like X-Forwarded-Authorization) other than Authorization. Ill keep you updated.

Thanks! I have skimmed over issues, we are not first https://github.com/kubernetes/kubernetes/issues/38775#issue-195629876
I will read couple more issues, maybe they have built the solution already

vlasov-y:
@yolossn I think we are in a dead end. I have made couple more checks:

1. CORS headers are not returned in a reply through the kube-api, while there are if I query Caddy directly (both with and without authorization)
2. It is possible to add kube-apiserver extra arg `--requestheader-extra-headers-prefix=X-Remote-Extra-` that will probably help to resolve the issue with Authorization header (not CORS), but making changes to kube-apiserver is too much for such case.

yolossn:
> @yolossn It looks like we've hit a dead end. I've done a few more checks:

Yes,  seems like there isn't any easy workaround for this. Ill search if there are any other workarounds without having to make major changes to the plugin itself. If nothing works maybe we should consider creating a port-forward to the prometheus service and then use it for querying the metrics but this solution will make the plugin unusable for in-cluster deployments of headlamp.

joaquimrocha:
Moving to Draft since it needs some further investigation.

Issue #149, opened 2025-01-23
app-catalog: UI papercuts


sniok:
Here are some small UI issues found during testing

---

Settings button 'floats' and is not clear that it is a button

![Image](https://github.com/user-attachments/assets/ab1fc1ee-27d5-4d70-b087-490ac994d74b)

---

App icon overlaps the text

![Image](https://github.com/user-attachments/assets/8019d8bd-43ed-41f7-bfce-301041768ae0)

---

Hovering install button makes it unreadable

Normal
![Image](https://github.com/user-attachments/assets/d6aa1d9b-8e26-4b43-80fb-eb8cb8aea1cf)

Hover
![Image](https://github.com/user-attachments/assets/869e9f73-e06e-4070-be44-999eacff4d72)

---

Typing into search makes the whole UI 'bounce'

Also note how Categories dropdown rerenders

https://github.com/user-attachments/assets/56e7ce44-bd57-4a8b-ae96-ec57aa58e1a5

---

After clicking on Settings page doesn't have a back button

![Image](https://github.com/user-attachments/assets/c247333a-1d57-4244-8689-b65c3f37f32d)

---

After clicking on pagination page button focus always goes to the Search field

![Image](https://github.com/user-attachments/assets/6c420d34-45d4-4faa-8e64-515bd362558c)

---

skoeva:
can take a look at these

skoeva:
I can open an issue for the settings button and the back button, those will need a refactor to get fixed. the rest should be good in #150

vyncent-t:
![image](https://github.com/user-attachments/assets/111855d6-48bd-4ce5-a922-b83beda3c867)

not sure about the part for the box button styling, I had originally used a boxed button but had feedback to change it early on to what it looks like now, maybe we can check in w/ @joaquimrocha to see if we can convert him to be a box enjoyer too 😉

vyncent-t:
also worth mentioning there has been some shifts for the way styles are done for this plugin, mainly changing away from using the style prop to the sx prop (see the current install button for the list view and then the details view, list view uses sx and details is still on style, causing that color issue)

PR #150, opened 2025-01-23
app-catalog: Fix UI bugs

skoeva:
These changes address minor bugs in the app catalog UI.

Fixes: #149


sniok:
Thanks for taking a look into this!

The install button turns light-blue on hover, let's keep it consistent with other black buttons we have in the app.

image

For example here's how cluster chooser button looks while hovered

image

---

When typing into the search the whole page freezes up, it sends a request and rerenders everything on each key press

CPU profiler shows this 

image

I'd suggest adding some form debouncing to `search` state updates. Alternatively you could try [useDeferredValue](https://react.dev/reference/react/useDeferredValue0) hook, although usually for this case simple debouncing works a bit better

---

I know this wasn't in the original issue but I've noticed that sometimes images are missing, looking at the url there's an `undefined` at the end of the path so maybe it wouldn't be hard to not show image at all when it's missing

image

skoeva:
@sniok thanks for the input here, added those changes to the PR 😃

sniok (approved):
looks good, thanks!

sniok:
@skoeva let's rebase and I'll merge this

skoeva:
@sniok added an additional commit because the app would crash without setting the params, but should be good now

PR #160, opened 2025-02-06
app-catalog: Improve the README for app-catalog

vyncent-t:
## Description
This PR makes improvements to the current README for the App Catalog plugin, it expands on the details of the plugin's functionality and includes start up guide and visual references.

## Before
image

## After
image

Link to after changes
https://github.com/headlamp-k8s/plugins/tree/update-app-catalog-readme/app-catalog

## Changes
- [x] Includes "about" section to describe the plugin and how it works
- [x] Includes screenshots for references
- [x] Includes Settings up environment
- [x] Includes links to additional resources

### DIFF
--- a/app-catalog/README.md
+++ b/app-catalog/README.md
@@ -2,3 +2,52 @@
 
 This plugin adds an app catalog to Headlamp, powered by ArtifactHub and Helm.
 
+![Screenshot of the Headlamp UI with the App Catalog browse page](https://github.com/user-attachments/assets/eeee154a-9146-4132-9bb8-a9e57d863161 "Headlamp UI with App Catalog browse page")
+
+## About the App Catalog
+
+This plugin enables the ability to install Helm charts from artifacthub directly within Headlamp. 
+
+*Note: By default, the App Catalog plugin has its filter set to "Official". This can be changed by navigating to the settings page using the `Settings` button to the right of the title bar.*
+
+## Developing Headlamp Plugins
+
+### Setting up the Environment
+
+1. Run the Headlamp app from the main branch:
+
+   - In a terminal window use:
+
+   ```bash
+   cd headlamp-k8s/headlamp/app
+   npm install
+   npm start
+   ```
+
+2. Set up the App Catalog:
+
+   - In a new terminal window, use:
+
+   ```bash
+   cd headlamp-k8s/plugins/app-catalog
+   npm install
+   npm start
+   ```
+
+3. Ensure the App Catalog is enabled
+
+   - From within Headlamp, navigate to the `Settings` page via the sidebar or the gear icon in the top right corner.
+   - Select the `Plugins` tab located in the sidebar.
+   - Enable the App Catalog plugin by toggling the switch to the right of the menu.
+   - Navigate back to the main page of your cluster.
+
+
+
+After completing these steps, you'll see the App Catalog link in the sidebar.
+
+![Screenshot of the App Catalog link in the sidebar](https://github.com/user-attachments/assets/5ee65579-abfc-4820-bf83-bcc4e2bea0f5 "Screenshot of the App Catalog link in the sidebar")
+
+
+## Contributing
+
+We welcome contributions! If you have ideas for improvements or encounter any issues, please open an issue or submit a pull request on our [GitHub repository](https://github.com/headlamp-k8s/plugins).


app-catalog/README.md
```diff
@@ -2,3 +2,60 @@
...
 
+![image](https://github.com/user-attachments/assets/eeee154a-9146-4132-9bb8-a9e57d863161)
+
+## About the App Catalog
+
+This plugin enables the ability to browse for Helm charts directly within the Headlamp UI, using categories and filters that are consistent with those on ArtifactHub.
+The App Catalog provides detailed chart information and metadata directly within Headlamp, allowing users to make informed decisions about the resources they want to deploy. 
+This centralized access reduces the need for external tools and improves the management process.  
+
+*Note: By default, the App Catalog plugin has it's filter set to "Official". This can be changed by navigating to the settings page using the `Settings` button to the right of the title bar.*
```
skoeva:
```suggestion
*Note: By default, the App Catalog plugin has its filter set to "Official". This can be changed by navigating to the settings page using the `Settings` button to the right of the title bar.*
```

app-catalog/README.md
```diff
@@ -2,3 +2,60 @@
...
+
+1. Run the Headlamp app from the main branch:
+
+   ```bash
+   cd headlamp-k8s/headlamp/app
+   npm install
+   npm start
+   ```
+
+2. Set up the App Catalog:
```
skoeva, line 27: not sure if it's clear that this should be done in a separate terminal

skoeva (commented):
Looks solid! Added a comment and fixed a typo

app-catalog/README.md
```diff
@@ -2,3 +2,60 @@
 
 This plugin adds an app catalog to Headlamp, powered by ArtifactHub and Helm.
 
+![image](https://github.com/user-attachments/assets/eeee154a-9146-4132-9bb8-a9e57d863161)
+
+## About the App Catalog
+
+This plugin enables the ability to browse for Helm charts directly within the Headlamp UI, using categories and filters that are consistent with those on ArtifactHub.
```
sniok: Could you simplify this whole paragraph? I feel like it mostly just says "With this plugin you can view and install Helm charts from artifacthub"

vyncent-t: I'm looking to provide as much context as possible from the point of view of a user who is not familiar with plugins or helm charts. 

Personally when interacting with something new I would like to know what it does, what to expect it to look like, and why I would want to use it.

It could be simplified to the one liner of "With this plugin you can view and install Helm charts from artifacthub" but it still leaves some questions unanswered for those not already in the know

for example:

- is this a frontend plugin that I can see in that integrates into the UI or is this some sort of window for terminal compatibility?
- what can I expect it to look like? (assuming photos are unavailable or the reader does not care to make sense of some images)
- what is the benefit of using it? / why should I use this over doing it the default way?

vyncent-t:
I would like to apply as much context as possible for people who are newer to plugins or using helm charts. I mostly just want to provide some answers to potential questions without having to take a "try it to find out" approach.

app-catalog/README.md
vyncent-t: thanks!

app-catalog/README.md
vyncent-t: thanks!

app-catalog/README.md
```diff
@@ -2,3 +2,64 @@
...
+   - From within Headlamp, navigate to the `Settings` page via the sidebar or the gear icon in the top right corner.
+   - Select the `Plugins` tab located in the sidebar.
+   - Enable the App Catalog plugin by toggling the switch to the right of the menu.
+   - Navigate back to the main page of your cluster.
+
+
+
+After completing these steps, you'll see the App Catalog link in the sidebar.
+
+   ![image](https://github.com/user-attachments/assets/5ee65579-abfc-4820-bf83-bcc4e2bea0f5)
```
illume: Can you please add an image description here? `[image]`

app-catalog/README.md
```diff
@@ -2,3 +2,64 @@
 
 This plugin adds an app catalog to Headlamp, powered by ArtifactHub and Helm.
 
+![image](https://github.com/user-attachments/assets/eeee154a-9146-4132-9bb8-a9e57d863161)
```
illume: This one needs an alt tag as well instead of image.

app-catalog/README.md
illume: Providing lots of information for newbies who want it is a good thing.

Please consider the readability.
image

Having an executive summary seems a great way to convey 70% of the information. With headings providing the other 20%, and the text/images the rest.

app-catalog/README.md
```diff
@@ -2,3 +2,64 @@
...
+
+
+
+After completing these steps, you'll see the App Catalog link in the sidebar.
+
+![Screenshot of the App Catalog link in the sidebar](https://github.com/user-attachments/assets/5ee65579-abfc-4820-bf83-bcc4e2bea0f5 "Screenshot of the App Catalog link in the sidebar")
+
+
+
+### Additional Resources
```
illume: I think these are the defaults that come when making a new plugin and are here accidentally?

Probably they can be removed, since they are useful for people writing plugins... but don't really make sense here? We don't have this section on other plugins.

illume (changes requested):
Thanks @vyncent-t 

There's still the complex sentence @sniok pointed out, and the note about extra "Additional Resources" section I think may have been added accidentally.

app-catalog/README.md
vyncent-t: I borrowed this section from the plugin-catalog plugin, will remove the additional resources part though.

app-catalog/README.md
vyncent-t: Using what Olek had commented and making it shorter and easier to read

illume (approved):
👍 Great, thanks.

PR #163, opened 2025-02-17
flux: sort events on timestamp

mgalesloot:
Users are normally interested in the most recent events. Events should therefore be sorted on timestamp with the newest event on top.

### DIFF
--- a/flux/src/helpers/index.tsx
+++ b/flux/src/helpers/index.tsx
@@ -110,18 +110,13 @@ export function ObjectEvents(props: { events: any }) {
             muiTableBodyCellProps: {
               align: 'right',
             },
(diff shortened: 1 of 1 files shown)


flux/src/helpers/index.tsx
```diff
@@ -110,18 +110,13 @@ export function ObjectEvents(props: { events: any }) {
...
-              );
-            },
           },
         ]}
         data={events}
         initialState={{
           sorting: [
             {
-              id: 'Age',
+              id: 'age',
```
ashu8912, line 119: why is this change required??

ashu8912 (approved):

Issue #168, opened 2025-02-25
eslint is broken in headlamp plugins


mgalesloot:
Eslint is used in headlamp plugins to check for coding standards. 

However, the headlamp plugins have an issue with running eslint: 

```
Cannot find module '@kinvolk/eslint-config/prettier-config'`
ESLint couldn't find the config "@kinvolk" to extend from. Please check that the name of the config is correct.
```

This is an issue is in the local build environment, and also in the github actions. 

Check the last build of the main branch: 
- https://github.com/headlamp-k8s/plugins/actions/runs/13507684850
- Specific error for flux plugin: https://github.com/headlamp-k8s/plugins/actions/runs/13507684850/job/37740983185

I noticed the config for "prettier", "eslintConfig" and "devDependencies" in package.json. I am not sure this config is compatible with "@kinvolk/headlamp-plugin": "^0.10.0".

sniok:
Eslint along with prettier were fixed in the https://github.com/headlamp-k8s/plugins/pull/178
The CI now works properly too

PR #174, opened 2025-03-03
flux: fix eslint config

mgalesloot:
Fix eslint config in Flux plugin. See #168. 

In the local environment you may need to run `npm install --force`


skoeva (approved):

ashu8912 (commented):
I think we would need to add the lint fixes as well right? can we do that as well in this PR?

mgalesloot:
I am worried that an overall lint fix would cause issues for other PRs currently in progress. I am happy to do this task if there are no other open PRs.

PR #184, opened 2025-03-13
prometheus: fix doubled memory usage

viceice:
Currently memory usage is doubled because cadvisor is taking the values from cgroups which contains the parent group.
The parrent group accumulates the memory, so we need  to ignore the parent.

https://stackoverflow.com/a/69282328/10109857

- closes headlamp-k8s/headlamp#2920


sniok:
related issue https://github.com/headlamp-k8s/headlamp/issues/2920

PR #185, opened 2025-03-13
flux: fix source ref namespace

### LINKED ISSUE #177: Flux: HelmRelease sourceRef from different namespace causes 404
It seems the Flux plugin can't handle source refs of HelmRepository when it's located in a different namespace:

```yaml
apiVersion: source.toolkit.fluxcd.io/v1
kind: HelmRepository
metadata:
  name: bitnami-charts
  namespace: flux-system # needed for renovate
spec:
  type: oci
  url: oci://ghcr.io/visualon/bitnamicharts
---
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
metadata:
  name: opensearch
  namespace: services
spec:
  interval: 30m
  timeout: 30m
  chart:
    spec:
      chart: opensearch
      version: 1.6.2
      sourceRef:
        kind: HelmRepository
        name: bitnami-charts
        namespace: flux-system
      interval: 12h
```

Errors:

Image

viceice:
Respect that the source ref can point to a different namespace.

- closes #177

### DIFF
--- a/flux/src/common/Table.tsx
+++ b/flux/src/common/Table.tsx
@@ -123,13 +123,13 @@ export function Table(props: TableProps) {
                 return name;
               },
               Cell: ({ row }: any) => {
-                const { name, pluralKind } = getSourceNameAndPluralKind(row.original);
+                const { name, pluralKind, namespace } = getSourceNameAndPluralKind(row.original);
                 return (
                   <Link
                     routeName="source"
                     params={{
                       pluralName: pluralKind,
-                      namespace: row.original.jsonData.metadata.namespace,
+                      namespace: namespace ?? row.original.jsonData.metadata.namespace,
                       name: name,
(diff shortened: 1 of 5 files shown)


joaquimrocha (approved):
The .eslintcache is updated here but we should actually add it to .gitignore.

viceice:
> The .eslintcache is updated here but we should actually add it to .gitignore.

I can send a follow up PR to drop the file

viceice:
or should I drop it here and add to ignore?

viceice:
> The .eslintcache is updated here but we should actually add it to .gitignore.

moved to #192

illume (approved):
👍 thank you

viceice:
@joaquimrocha Any chances to also get a new flux plugin release?

Issue #198, opened 2025-03-22
Flux: Overview page shows OCI repositories as error


mgalesloot:
HelmRelease with type OCI does not have a status. They should be considered as ready to use. 

See https://fluxcd.io/flux/components/source/helmrepositories/#helm-oci-repository
> Because the OCI Helm repository is a data container, there’s nothing to report for READY and STATUS columns above. The existence of the object can be considered to be ready for use.

joaquimrocha:
Thanks @viceice . Looks like Ashu is taking a look.

PR #199, opened 2025-03-22
flux: fix status helmrepository type oci

mgalesloot:
Fixes https://github.com/headlamp-k8s/plugins/issues/198


ashu8912 (approved):
Thanks for this change, any knowledge share on why we have to handle ociHelmRepository seperately here???

mgalesloot:
The ociHelmRepository needs special handling because it is **always** ready, and does not have a status. See explanation at https://fluxcd.io/flux/components/source/helmrepositories/#helm-oci-repository

PR #201, opened 2025-03-22
Flux: imagepolicy reconcile should refer to imagerepository

### LINKED ISSUE #200: Flux: ImagePolicy reconcilation
Flux does not currently support reconcile method on the ImagePolicy. Instead, users should reconcile the ImageRepository. 

Required changes: 
- Hyperlink in ImagePolicy to the ref ImageRepository
- Remove reconcile buttons from ImagePolicy 
- Remove Force reconcile as it is not supported 
- Hide Reconcile status fields from ImagePolicy 

Also: 
- Show Tag Filter pattern for an ImagePolicy

mgalesloot:
Fixes #200

### DIFF
--- a/flux/src/image-automation/ImageAutomationSingle.tsx
+++ b/flux/src/image-automation/ImageAutomationSingle.tsx
@@ -9,12 +9,7 @@ import Editor from '@monaco-editor/react';
 import React, { ReactNode } from 'react';
 import { useParams } from 'react-router-dom';
 import YAML from 'yaml';
-import {
-  ForceReconciliationAction,
-  ResumeAction,
-  SuspendAction,
-  SyncAction,
-} from '../actions/index';
+import { ResumeAction, SuspendAction, SyncAction } from '../actions/index';
 import Flux404 from '../checkflux';
 import Link from '../common/Link';
 import RemainingTimeDisplay from '../common/RemainingTimeDisplay';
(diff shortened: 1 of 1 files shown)


mgalesloot:
The hyperlink in ImagePolicy to ImageRepository behaves strange in my test. After clicking on the link to the ImageRepository there is an in-place edit of the current page. I dont understand what is causing this. In attached screenshot you can see there resulting page.: 1) The URL in ths addressbar shows the target page `imagerepositories/bundle/api-gateway`, the title of the page is unchanged and is still ImagePolicy, in the events below we see in the message field the events from the repository.
image

flux/src/image-automation/ImageAutomationSingle.tsx
```diff
@@ -74,32 +69,44 @@ function CustomResourceDetails(props) {
...
       });
       extraInfo.push({
         name: 'Canonical Image Name',
-        value: resource?.jsonData?.status?.canonicalImageName || '-',
+        value: resource.jsonData?.status?.canonicalImageName || '-',
       });
     }
-    if (resource?.jsonData.kind === 'ImagePolicy') {
+
+    if (resource.jsonData.kind === 'ImagePolicy') {
```
joaquimrocha, line 95: Why this change? Seems like a risky change.

mgalesloot: I made this change because I felt the code would be more solid by replacing the multiple ? with a single check. This check is done on line 176. For me it works better because I tend to forget to put the ? all the time. 
>   if (!resource) {
    return <></>;
  }

Don't like it that way? Let me know and I revert it.

ashu8912: I think the problem with this is that if we pass null resource to table it shows a loader in that case but when we handle it like this we miss out on that

ashu8912 (commented):
For the url change but view not changing issue the line
 https://github.com/headlamp-k8s/plugins/pull/201/commits/eadefdc39e49734ecfc1a4f225baf674c4830704#diff-86b9680f03d3195c67e5ed23dc6a689c052e0daa55076ae25669d780da9d08c7R62
 needs to be updated to
  const [resource, error] = resourceClass.useGet(name, namespace);

flux/src/image-automation/ImageAutomationSingle.tsx
mgalesloot: This is reworked. I kept a single check on null resource in prepareExtraInfo(), but let the Table render always.

PR #205, opened 2025-03-26
github: Add workflow to build and publish container image

joaquimrocha:
This is based on #172 but is limited to building and publishing the image.
I believe we need to have what's in #172 but we need to do it in separate actions because, e.g. if the image fails, we shouldn't resort to doing things locally.
In the future we may trigger this workflow from a tag push.


skoeva (approved):
Makes sense, thanks for splitting things up here

PR #210, opened 2025-03-26
flux: Update ArtifactHub to use plugin version 0.2.0

joaquimrocha:


joaquimrocha:
This should be an automated step. Merging now since it's trivial.

PR #211, opened 2025-03-27
Fix multi arch image building

joaquimrocha:
Turns out that some plugins have dependencies that are arch specific and somehow this is failing for ARM.
I have added some changes but ultimately could only build the AMD64 version for the flux plugin (I had tested with the kompose plugin in my own clone).


l-nmch:
Hi, I was wondering if this PR would be merged someday. It could be really helpful to have an ARM version for the flux plugin !

kksudo:
https://github.com/headlamp-k8s/plugins/pull/290

PR #212, opened 2025-03-28
minikube: CommandCluster: Add warning if leaving and command is still running

illume:
image

### How to test

- in this branch minikube/ folder `npm start`
- Then in the app, go to load cluster.
- Start adding a cluster. 
- Then when it starts running... Try to navigate away from the page. You should see a warning message, click ok to leave (cancel to not leave).

( Uses [React Router Prompt component](https://v5.reactrouter.com/core/api/Prompt) )


joaquimrocha (approved):
I think for now this works, but in the future, we should have the process start from an outside component that's always visible, like a TopBarAction, so we can do whatever and it still runs on the background (we just display the state/output in the main view, etc.).

PR #213, opened 2025-03-28
minikube: Add default cluster name, and validation for existing cluster names

illume:
- Add generation of cluster name by default (minikube, minikube-1, minikube-2, etc)
- Add validation for existing cluster names

### How to test

- npm start
- Load Cluster
- Add minikube cluster
- See that the name is minikube-1 or so.
- Enter an existing cluster name, and see an error, 
  - prevented from using form to create with existing name

image

---

image


joaquimrocha:
> image

I see what you have done there. 😸

PR #232, opened 2025-05-02
prometheus: Add resolution input for metrics

adwait-godbole:
The official Prometheus Query UI accepts resolution as an input from the user:

image

It will be better if we also allow Headlamp to accept such an input. This will help us in calculating the step size at runtime based on `timeRange` and `resolution` rather than hard coding it like we are currently doing it at [this](https://github.com/headlamp-k8s/plugins/blob/main/prometheus/src/util.ts#L250-#L279) place. 

The inspiration for dynamic step size calculation has been taken from the [official Prometheus' UI code](https://github.com/prometheus/prometheus/blob/main/web/ui/mantine-ui/src/state/queryPageSlice.ts#L28-#L50)

image

image

cc @yolossn

### DIFF
--- a/prometheus/src/components/Chart/CPUChart/CPUChart.tsx
+++ b/prometheus/src/components/Chart/CPUChart/CPUChart.tsx
@@ -17,6 +17,7 @@ interface CPUChartProps {
   query: string;
   prometheusPrefix: string;
   interval: string;
+  resolution: string;
   autoRefresh: boolean;
   subPath: string;
 }
--- a/prometheus/src/components/Chart/DiskChart/DiskChart.tsx
+++ b/prometheus/src/components/Chart/DiskChart/DiskChart.tsx
@@ -18,6 +18,7 @@ interface DiskChartProps {
   usageQuery: string;
   capacityQuery: string;
   interval: string;
+  resolution: string;
   prometheusPrefix: string;
   autoRefresh: boolean;
   subPath: string;
--- a/prometheus/src/components/Chart/MemoryChart/MemoryChart.tsx
+++ b/prometheus/src/components/Chart/MemoryChart/MemoryChart.tsx
@@ -18,6 +18,7 @@ interface MemoryChartProps {
   query: string;
   prometheusPrefix: string;
   interval: string;
+  resolution: string;
   autoRefresh: boolean;
   subPath: string;
 }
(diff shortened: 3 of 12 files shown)


feym78:
@adwait-godbole 
Could you please explain where it will be configured? Dynamically in resource or same as now statically in plugin settings?

prometheus/src/components/Chart/CPUChart/CPUChart.tsx
```diff
@@ -17,6 +17,7 @@ interface CPUChartProps {
   query: string;
   prometheusPrefix: string;
   interval: string;
+  resolution: string;
```
yolossn, line 20: resolution is added to the props but it is not used in the CPUChart component.

prometheus/src/components/Chart/DiskChart/DiskChart.tsx
```diff
@@ -18,6 +18,7 @@ interface DiskChartProps {
   usageQuery: string;
   capacityQuery: string;
   interval: string;
+  resolution: string;
```
yolossn, line 21: similar to CPU chart the resolution prop is not used.

prometheus/src/components/Chart/MemoryChart/MemoryChart.tsx
```diff
@@ -18,6 +18,7 @@ interface MemoryChartProps {
   query: string;
   prometheusPrefix: string;
   interval: string;
+  resolution: string;
```
yolossn, line 21: The resolution prop is not used in the component.

prometheus/src/util.ts
```diff
@@ -240,42 +252,77 @@ export function formatBytes(bytes: number) {
 /**
  * Calculates the time range based on the given interval.
  * @param {string} interval - The time interval (e.g., '10m', '1h', '24h', 'week').
- * @returns {Object} An object containing the 'from' timestamp, 'to' timestamp, and 'step' in seconds.
+ * @returns {Object} An object containing the 'from' timestamp and 'to' timestamp in seconds.
  */
-export function getTimeRange(interval: string): { from: number; to: number; step: number } {
+export function getTimeRange(interval: string): { from: number; to: number } {
```
yolossn: IMHO we can get resolution as an arg and calculate the step and return the step here itself, instead of first fetching the time range and then finding the step size. WDYT?

prometheus/src/components/Chart/CPUChart/CPUChart.tsx
yolossn: My bad it is used in the chart component, resolving this

yolossn (approved):
Great work, LGTM.

PR #240, opened 2025-05-12
flux: Sanitize domain URL for validation

skoeva:
These changes to the flux source code replace the endsWith check with a more precise validation that ensures the groupName is either toolkit.fluxcd.io or a valid subdomain of it.


flux/src/helm-releases/Inventory.tsx
```diff
@@ -271,10 +271,12 @@ function inventoryNameLink(item): JSX.Element {
   const pluralName = PluralName(kind);
 
   // Flux types
-  if (groupName.endsWith('toolkit.fluxcd.io')) {
+  const allowedDomain = 'toolkit.fluxcd.io';
+  const groupNameParts = groupName.split('.');
+  if (groupNameParts.length >= 3 && groupNameParts.slice(-2).join('.') === allowedDomain) {
```
sniok: I think this code is essentially equivalent to `groupName.endsWith('toolkit.fluxcd.io')`? 
Could you provide a case where previous version fails? I'm not sure I understand what this fixes/improves

skoeva: I can simplify the change but essentially the current case does not entirely account for malicious urls like 'badtoolkit.fluxcd.io'

PR #243, opened 2025-05-13
github/workflows/headlamp-plugin-github-workflow: Limit token perms

skoeva:
This change limits the permissions of the GITHUB_TOKEN as desired.


illume (approved):
👍 thanks!

PR #254, opened 2025-05-24
KEDA: Add Map View for KEDA Resources

adwait-godbole:
`ScaledObject` View:

Screenshot from 2025-05-24 12-51-20

`ScaledJob` View:

Screenshot from 2025-05-24 13-34-51

Hi @yolossn, Hi @sniok.

I have the below queries that need to be addressed. I would appreciate your help so that I can fix them at the earliest.

1) On the Map View landing page grouped by `default` namespace, we don't see `ScaledObject` and/or `ScaledJob` resources even after selecting KEDA checkbox in the dropdown, they are hidden behind their respective `Deployment` and/or `Job` nodes. Is there a way of changing this behavior?

image

2) In the `ScaledObject` Map View page, I was trying to load the `HPA` resources using `HPA.useList()` functionality at the below line:

https://github.com/headlamp-k8s/plugins/blob/0c71913206d49536723ec2b541384833b16e6988/keda/src/mapView.tsx#L125

This is not working and the map view simply crashes on doing this. However, this does work for loading of other resources like `Deployment`, `StatefulSet`, etc.

3) Is there a way of configuring the on hover display text for the nodes?

for `ScaledObject` I currrently see the below which looks very cluttered and congested since we are displaying all the events:

Screenshot from 2025-05-24 15-41-24

but it would be much more UX friendly if we are able to show something similar to this:

image

### DIFF
--- a/keda/src/mapView.tsx
+++ b/keda/src/mapView.tsx
@@ -1,0 +1,254 @@
+import { useMemo } from 'react';
+import { Icon } from '@iconify/react';
+import { ScaledObject } from "./resources/scaledobject";
+import { ScaledObjectDetail } from './components/scaledobjects/Detail';
+import { K8s } from '@kinvolk/headlamp-plugin/lib';
+import { ScaledJob } from "./resources/scaledjob";
+import { ScaledJobDetail } from './components/scaledjobs/Detail';
+import Job from '@kinvolk/headlamp-plugin/lib/K8s/job';
+import { TriggerAuthentication } from "./resources/triggerAuthentication";
+import { TriggerAuthenticationDetail } from './components/triggerauthentication/Detail';
+import { ClusterTriggerAuthentication } from './resources/clusterTriggerAuthentication';
+import { ClusterTriggerAuthenticationDetail } from './components/clustertriggerauthentication/Detail';
+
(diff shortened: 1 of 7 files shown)


adwait-godbole:
@yolossn can you review?

keda/src/mapView.tsx
```diff
@@ -0,0 +1,254 @@
...
+      const [clusterTriggerAuthentications] = ClusterTriggerAuthentication.useList();
+      const [hpas] = K8s.ResourceClasses.HorizontalPodAutoscaler.useList();
+      
+      return useMemo(() => {
+        if (!scaledObjects) return null;
+        
+        const nodes = scaledObjects?.map(it => ({
+          id: it.metadata.uid,
+          kubeObject: it,
+          weight: 1100,
```
adwait-godbole, line 133: @yolossn, once kubernetes-sigs/headlamp/pull/3389 gets merged, having this weight will work as per our requirement of showing ScaledObject over HPA. Here's a demo after testing on my local on the same branch as the weights feature branch PR on Headlamp combined with our plugin Map View running in the background:

[Screencast from 30-05-25 04:53:52 PM IST.webm](https://github.com/user-attachments/assets/a992193b-3d70-4874-8ca2-ecd48b124e8e)

PR #255, opened 2025-05-26
Add the AI plugin 

ashu8912:
This PR introduces a new AI Assistant plugin for Headlamp that enhances Kubernetes observability and interaction using Large Language Models (LLMs). The plugin leverages LangChain to support multiple AI providers (e.g., OpenAI, Claude, Gemini) and provides a modular, extensible interface for intelligent Kubernetes assistance.

Features

- Multi-provider LLM Support:
  Seamlessly switch between different AI providers using LangChain's abstraction. Easily configurable via settings.

- Tool-augmented AI Prompts:
  Adds tool capabilities to prompts, enabling the assistant to execute predefined functions such as summarization,                 explanation, or cluster status checks.

- Kubernetes API Integration:
  Enables the AI assistant to include live Kubernetes API data in its responses. Supports direct requests to the API server to retrieve and explain cluster information.

- Interactive UI Integration:
  Integrated into the Headlamp UI as a plugin with a dedicated drawer for asking questions, receiving actionable insights, and running tool-enhanced queries.

Use Cases
    "Explain why this pod is in CrashLoopBackOff."
    "What does this HelmRelease do?"
    "Summarize all deployments in the prod namespace."
    "List all failing Kustomizations and suggest fixes."

Configuration
    Provider selection (OpenAI, Azure,Anthropic, etc.)

### DIFF
--- a/ai-plugin/src/index.tsx
+++ b/ai-plugin/src/index.tsx
@@ -11,11 +11,13 @@ import {
   Divider,
   FormControlLabel,
   IconButton,
-  Popover,
+  Paper,
+  Popper,
   Switch,
   ToggleButton,
   Tooltip,
   Typography,
+  useTheme
 } from '@mui/material';
 import React from 'react';
 import { useHistory } from 'react-router-dom';
@@ -129,6 +131,7 @@ function HeadlampAIPrompt() {
   const history = useHistory();
   const [popoverAnchor, setPopoverAnchor] = React.useState<HTMLElement | null>(null);
   const [showPopover, setShowPopover] = React.useState(false);
+  const theme = useTheme();
 
   const hasShownPopover = savedConfigs?.configPopoverShown || false;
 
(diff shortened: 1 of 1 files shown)


sniok:
there's a render loop in this plugin, you should add useMemo in `useDynamicPrompts` hook

Ollama doesn't work with tools, current ChatOllama is marked as deprecated and using `import { ChatOllama } from '@langchain/ollama';` fixes issue with tools

sniok:
For local models I'm getting this error in the console 

```
Error binding tools to local model: TypeError: this.model.bindTools is not a function
    at rJ.configureTools (LangChainManager.ts:206:36)
```

ai-plugin/src/helper/apihelper.tsx
```diff
@@ -0,0 +1,295 @@
...
+  if (method.toUpperCase() === 'GET') {
+    const cleanedUrl = cleanUrl(url);
+
+    try {
+      const requestOptions: any = {
+        method,
+        cluster,
+        body: body === '' ? undefined : body,
+      };
+      if (isLogRequest(url)) {
```
sniok, line 172: you should let LLM set headers so it can request text or json itself, then based on the response Content-Type header you can parse it as .text() or .json() without the need to hardcode check for logs request

sniok:
image
tools are being configured on each keypress when typing message

PR #268, opened 2025-06-05
KEDA Plugin

yolossn:
This PR contains the work done by @adwait-godbole during his LFX mentorship. As planned, the ongoing PRs were merged into a feature branch, and this final PR is intended for team-wide review before merging into the main branch and releasing the plugin.

### DIFF
--- a/keda/src/components/common/CommonComponents.tsx
+++ b/keda/src/components/common/CommonComponents.tsx
@@ -68,6 +68,21 @@ export function NotInstalledBanner({ isLoading = false }: NotInstalledBannerProp
   );
 }
 
+interface KedaInstallCheckProps {
+  children: React.ReactNode;
+  fallback?: React.ReactNode;
+}
+
+export function KedaInstallCheck({ children, fallback }: KedaInstallCheckProps) {
+  const { isKedaInstalled, isKedaCheckLoading } = useKedaInstalled();
+
+  if (!isKedaInstalled) {
+    return fallback || <NotInstalledBanner isLoading={isKedaCheckLoading} />;
+  }
+
+  return <>{children}</>;
+}
+
 interface BaseKedaAuthenticationProps {
   title?: string;
   resourceType: typeof TriggerAuthentication | typeof ClusterTriggerAuthentication;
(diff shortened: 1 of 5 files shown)


prometheus/src/components/Chart/KedaActiveJobsChart/KedaActiveJobsChart.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+interface KedaActiveJobsChartProps extends KedaChartProps {
+  prometheusPrefix: string;
+  interval: string;
+  resolution: string;
+  autoRefresh: boolean;
+  subPath: string;
+}
+
+export function KedaActiveJobsChart(props: KedaActiveJobsChartProps) {
```
sniok, line 21: I'm not sure if these charts belong in the prometheus plugin to be honest

keda/src/components/common/CommonComponents.tsx
```diff
@@ -0,0 +1,646 @@
...
+  namespace?: string;
+  name?: string;
+}
+
+export function BaseKedaAuthenticationDetail({
+  resourceType,
+  namespace,
+  name,
+}: BaseKedaAuthenticationProps) {
+  const { isKedaInstalled, isKedaCheckLoading } = useKedaInstalled();
```
sniok: This check is duplicated in BaseKedaAuthenticationDetail and BaseKedaAuthenticationList, I'd suggest extracting this logic into a composable component like so:

```tsx
<KedaAuthentication>
    {/* this only shows when keda is installed */}
    <KedaDetails {...} />
</KedaAuthentication>
```

sniok (approved):
Added a couple of comments but nothing major. Overall this looks great!

prometheus/src/components/Chart/KedaActiveJobsChart/KedaActiveJobsChart.tsx
yolossn: Yes this was discussed but we arrived at this solution because making prometheus plugin allow other plugins to register resource metrics handling will be a different task as a whole. The current plan is to eventually support a way for plugins to register resource metrics handling. Once that’s in place, this logic can be moved into the KEDA plugin itself.

PR #271, opened 2025-06-08
flux: add flux 2.0.0 CRD versions

### LINKED ISSUE #229: Flux: Overview page and HelmReleases page supports different API versions
Using version 0.2.0, the flux Overview page will lookup helm.toolkit.fluxcd.io/v2beta1 helmreleases, but the HelmReleases page will only look for helm.toolkit.fluxcd.io/v2 helmreleases

explsd:
Fixes #229 by adding some CRD versions used by older versions of Flux.

### DIFF
--- a/flux/src/notifications/NotificationList.tsx
+++ b/flux/src/notifications/NotificationList.tsx
@@ -16,7 +16,10 @@ const notificationVersion = 'v1beta3';
 
 export function alertNotificationClass() {
   return makeCustomResourceClass({
-    apiInfo: [{ group: notificationGroup, version: notificationVersion }],
+    apiInfo: [
+      { group: notificationGroup, version: notificationVersion },
+      { group: notificationGroup, version: 'v1beta2' },
+    ],
     isNamespaced: true,
     singularName: 'Alert',
     pluralName: 'alerts',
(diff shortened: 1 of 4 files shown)


skoeva (commented):
The failing test in the CI could be resolved by running prettier in the flux directory

skoeva (commented):
Squash the new commit into the old one and should be ready

illume (approved):
👍 looks good to me. I checked the flux release notes, and this seems to cover it.

Will leave it for @ashu8912 to review/merge.

illume:
I think gitRepositoryClass bucketRepositoryClass and helmRepositoryClass also need older/new API versions added. In https://github.com/headlamp-k8s/plugins/blob/639608f5f838c0b90308a9723bbfc10ffa8fab12/flux/src/sources/SourceList.tsx#L14

explsd:
My goal is Flux v2.0 support, which is already at GitRepository v1, but it's true that BucketRepository would still be v1beta2

explsd:
@joaquimrocha I'll add BucketRepository v1beta2, then I think it should be good to go

flux/src/notifications/NotificationList.tsx
```diff
@@ -34,7 +40,11 @@ export function providerNotificationClass() {
 
 export function receiverNotificationClass() {
   return makeCustomResourceClass({
-    apiInfo: [{ group: notificationGroup, version: notificationVersion }],
+    apiInfo: [
+      { group: notificationGroup, version: notificationVersion },
+      { group: notificationGroup, version: 'v1beta2' },
+      { group: notificationGroup, version: 'v1' },
```
ashu8912, line 46: do we have v1beta2 for notification??
https://fluxcd.io/flux/components/notification/api/

ashu8912 (approved):
Thank you so much for these changes, i have tested and it does cover all the old versions

illume (approved):
🎉

PR #290, opened 2025-06-22
Fix multiarch build

kksudo:
# Optimized Docker Builds, Workflow, and Plugin Updates

This PR improves the Docker build process, GitHub Actions workflow, and plugin dependencies for the Headlamp Kubernetes plugins repository. 

## Main Changes
1. **Fixed Flux Dependencies**: Resolved issues with the `flux` plugin dependencies to ensure stability.
2. **Fixed Multi-Architecture Builds**: Corrected multi-architecture build issues for broader platform compatibility.
3. **Bumped Flux Plugin Version**: Updated the `flux` plugin version in `flux/package.json` from `0.2.0` to `0.2.1`.

## Additional Improvements
1. **Multi-Plugin Builds**: Updated `.github/workflows/build-and-publish-image.yml` to support building multiple plugins in a single GitHub Actions run using a matrix strategy.
2. **Cleaned Up Orphaned Workflow Processes**: Removed lingering processes in the GitHub Actions workflow for improved reliability.
3. **Added GitHub Actions Cache**: Implemented caching to speed up workflow runs.
4. **Optimized Dockerfile**: Added logic to skip development dependencies in production builds when `ENVIRONMENT` is set to `production`.
5. **Added `.dockerignore`**: Excluded `node_modules`, `.git`, and test files to reduce Docker image size and improve build efficiency.
6. **Configurable Image Versions**: Introduced `BASE_IMAGE_VERSION` and `FINAL_IMAGE_VERSION` arguments in the `Dockerfile` and fixed unsupported Node base image versions.
7. **Added `.gitignore`**: Included a `.gitignore` file to prevent unnecessary files from being tracked.


illume:
Wow, thanks for all these improvements.

Would you mind breaking the changes up into separate commits? With one separate topic per commit.

I think the flux package.json version change can be removed from the PR. The release can come later.

kksudo:
> * **Fixed Flux Dependencies**: Resolved issues with the `flux` plugin dependencies to ensure stability.

Done, pls check attached PRs

kksudo:
> Wow, thanks for all these improvements.
> 
> Would you mind breaking the changes up into separate commits? With one separate topic per commit.
> 
> I think the flux package.json version change can be removed from the PR. The release can come later.


Fixed Flux Dependencies: Resolved issues with the flux plugin dependencies to ensure stability.
Done, pls check attached PRs

Dockerfile
```diff
@@ -1,5 +1,7 @@
-# Use the official Node.js 18 image as the base image for building the plugins
-FROM node:18@sha256:d0bbfdbad0bff8253e6159dcbee42141db4fc309365d5b8bcfce46ed71569078 AS builder
+ARG BASE_IMAGE_VERSION=18
```
joaquimrocha, line 1: For security reasons we need the full image version with the hash.

Dockerfile
```diff
@@ -1,5 +1,7 @@
...
+ARG FINAL_IMAGE_VERSION=3.20.3
```
joaquimrocha, line 2: For security reasons we need the full image version with the hash.

flux/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@headlamp-k8s/flux",
-  "version": "0.2.0",
```
joaquimrocha, line 3: Can you remove the flux commits from this PR? I think this should be just about the image building.

flux/README.md
joaquimrocha, line 1: The commit title needs to summarize what this commit does. It does update the docs but it should instead indicate what this is about.

flux/README.md
```diff
@@ -67,10 +67,15 @@ spec:
       - command:
           - /bin/sh
           - -c
-          - mkdir -p /build/plugins && cp -r /plugins/* /build/plugins/
+          - mkdir -p /build/plugins && cp -r /plugins/* /build/plugins/ && chown -R 100:101 /build
         image: ghcr.io/headlamp-k8s/headlamp-plugin-flux:latest
         imagePullPolicy: Always
         name: headlamp-plugins
+        securityContext:
```
joaquimrocha, line 74: Do we need to run this as root?

joaquimrocha (changes requested):

Dockerfile
kksudo: > For security reasons we need the full image version with the hash.

Would you like to get the [dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference?learn=dependency_version_updates&learnProduct=code-security) , which could watch updates and make a PR for the updated docker image versions?

joaquimrocha: I thought we had it. I will check it out. But in the current code, we need to keep the hash.

flux/README.md
kksudo: Not sure, I guess no.

illume:
Yeah, let's close this one now you have all the other ones :)

PR #291, opened 2025-06-22
flux: Readme: Improve documentation

### LINKED ISSUE #214: Flux: v0.2.0 can't create plugins/flux directory
Trying to update my flux plugin from v0.1.0-beta-2, and my init container is throwing errors stating it can't overwrite the .json files already existing.

I tried to add a command to delete them first, but got a permission denied, so I just blew away the PVC to start from scratch, and now I'm getting
`cp: can't create directory '/build/plugins/flux': Permission denied ` 

This is my updated non-working hr:
https://github.com/joryirving/home-ops/blob/8ea0eaf26d14816aad14266af941f7e4f8855198/kubernetes/apps/base/flux-system/headlamp/helmrelease.yaml

This is my reverted, working hr on the old version:
https://github.com/joryirving/home-ops/blob/ff723a4a7d3ddce3ab7723c8e0d545f9bab24fb5/kubernetes/apps/base/flux-system/headlamp/helmrelease.yaml

I can't find anything in the changelog where I need to do anything other than just bump the version.

kksudo:
## Description

Improve the Flux plugin [documentation](https://github.com/kksudo/plugins/tree/bugfix/fix-multiarch/flux)

Fixes #214
- https://github.com/headlamp-k8s/plugins/issues/214

---
Split [this](https://github.com/headlamp-k8s/plugins/pull/290) PR.
https://github.com/headlamp-k8s/plugins/pull/290#issuecomment-2994200788

### DIFF
--- a/flux/README.md
+++ b/flux/README.md
@@ -92,3 +92,32 @@ spec:
         persistentVolumeClaim:
           claimName: headlamp # The name of the Helm release
 ```
+
+As alternative, you can also use the Use EmptyDir (Ephemeral Shared Volume) to pass files from the init containers to the main container. 
+
+```yaml
+config:
+  pluginsDir: /build/plugins
+initContainers:
+  - command:
+      - /bin/sh
+      - -c
+      - mkdir -p /build/plugins && cp -r /plugins/* /build/plugins/ && chown -R 100:101 /build
+    image: ghcr.io/headlamp-k8s/headlamp-plugin-flux:latest
+    imagePullPolicy: Always
+    name: headlamp-plugins
+    securityContext:
+      runAsNonRoot: false
+      privileged: false
+      runAsUser: 0
+      runAsGroup: 0
+    volumeMounts:
+      - mountPath: /build/plugins
+        name: headlamp-plugins
+volumeMounts:
+  - mountPath: /build/plugins
+    name: headlamp-plugins
+volumes:
+  - name: headlamp-plugins
+    emptyDir: {}
+```


skoeva (approved):
LGTM

kksudo:
The security context is mandatory to copy changes, without it, you will get an error. 
Also, there is no statefulset app, that is why use the PV with PVC is not correct, as the alternative I added an option to use an empty dir.

> can you please explain this change, is this an issue?? Why would we have a permission issue here???

flux/README.md
```diff
@@ -67,10 +67,15 @@ spec:
       - command:
           - /bin/sh
           - -c
-          - mkdir -p /build/plugins && cp -r /plugins/* /build/plugins/
+          - mkdir -p /build/plugins && cp -r /plugins/* /build/plugins/ && chown -R 100:101 /build
```
illume, line 70: Oh... I just realised the 100:101 here is the headlamp user and group from the alpine headlamp Dockerfile container?

Can we use `chown -R headlamp:headlamp /build` here instead?

How about a comment explaining why the chown is necessary?

flux/README.md
```diff
@@ -67,10 +67,15 @@ spec:
...
         image: ghcr.io/headlamp-k8s/headlamp-plugin-flux:latest
         imagePullPolicy: Always
         name: headlamp-plugins
+        securityContext:
```
illume, line 74: This part here is required because of the `chown` used above? Or is it for something else?

Could you please add a comment to there explaining why it's necessary?

illume:
Added "Fixes https://github.com/headlamp-k8s/plugins/issues/214" to PR description.

flux/README.md
kksudo: The [securityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) overwrite the [USER headlamp](https://github.com/headlamp-k8s/plugins/blob/main/Dockerfile#L58C1-L59C1) non root user.
Without it, we don't have enough permissions to do the dirs with non `headlamp` owner.

flux/README.md
kksudo: > Oh... I just realised the 100:101 here is the headlamp user and group from the alpine headlamp Dockerfile container?

Yes
https://github.com/headlamp-k8s/plugins/blob/main/Dockerfile#L45
Actually we can improve also this line, to define exactly the number of the group to be more clear.

kksudo: > Can we use `chown -R headlamp:headlamp /build` here instead?

yes, but it's not a true way, because from image to image the number of the `headlamp` group could be diff.
It's a better to update the Dockerfile to set up a static number, smth like:

```
# Create a group with specific GID and user with specific UID
RUN addgroup -g 101 -S headlamp && \
    adduser -u 100 -S -G headlamp headlamp
```

I can update it, WDYT ?

kksudo:
> @ashu8912 @kksudo I added a few notes/questions at points where I think we can add comments to make this a bit easier to understand. Please check them to see what you think?

Answered

flux/README.md
illume: Thanks. I understand I think... so if someone uses headlamp:headlamp in say their own custom Debian image maybe it's different from an alpine image?

Yeah, I think it's a good change to make.

flux/README.md
illume: Ok, that makes sense now. Thanks.

Could you please add a comment about that in there for future readers?

kksudo: > Ok, that makes sense now. Thanks.
> 
> Could you please add a comment about that in there for future readers?

I can, it's not a big deal.
But, are you sure if it's necessary? Because it's a pretty simple setting. Moreover, we can't annotate every line. We should do it, only if it's some exception or TODO or workaround.

WDYT?

ashu8912: Umm i don't understand this change we already did chwon above so we would not want to run this as root right??

maybe this is what we need here

```
securityContext:
  runAsNonRoot: true
  runAsUser: 100
  runAsGroup: 101
  privileged: false
```

kksudo: Yes, it's true. What is the user who run this cmd? 
Without it, you will fail. 
Btw, you can try to run without these changes to check it.

kksudo: > Umm i don't understand this change we already did chwon above so we would not want to run this as root right??
> 
> maybe this is what we need here
> 
> ```
> securityContext:
>   runAsNonRoot: true
>   runAsUser: 100
>   runAsGroup: 101
>   privileged: false
> ```

@ashu8912 
Again, did you check the root cause of these changes [here ](https://github.com/headlamp-k8s/plugins/issues/214#issuecomment-2987831094)?

image

PR #298, opened 2025-06-25
KEDA: Pin golang dependency by hash

skoeva:
This change pins the golang dependency in the test-files Dockerfile by hash as desired.


illume (approved):
👍 thanks

PR #303, opened 2025-06-26
Bump @babel/core and @babel/runtime to 7.27.1

skoeva:
These changes bump the @babel/core and @babel/runtime dependencies to the desired 7.27.1 version.


illume (changes requested):
Would you mind rebasing?

PR #310, opened 2025-07-01
flux: Add types and fix unknown string function in Inventory

joaquimrocha:
Due to the types not being defined, a typo ("endswith" instead of "endsWith") was being covered and led to issues when running.


ashu8912 (approved):
LGTM!!

PR #324, opened 2025-07-11
flux: Fix pluralName helper and update how we show overview

ashu8912:
We want the failed status to be shown first thing in the overview. This PR takes the failed ones and shows them first and then the rest. We also fix the pluralName to call for functions when the kind could be undefined.


ashu8912:
Screenshot 2025-07-14 at 1 31 21 PM

joaquimrocha:
Tested and looks fine.

PR #325, opened 2025-07-12
Dockerfile: restored npm install to fix Docker build on main branch

stickshift:
Fixes the Dockerfile by reverting the npm ci change from @skoeva. npm ci relies on the package-lock.json that is explicitly deleted in the previous step. I agree npm ci is better but before restoring npm ci, the previous step that deletes package-lock.json needs to be fixed.

PR #352, opened 2025-08-01
Dockerfile: Use npm install instead of npm ci

skoeva:
This PR reverts the change done in #315 since we delete the `package-lock.json` in the previous step.

Follow-up from #325


illume (approved):
👍 

Thanks @stickshift and @skoeva

illume:
I guess this will need a more involved fix later to remove "npm install".

PR #358, opened 2025-08-08
ai-plugin: Fix update request and log errors

ashu8912:


ai-assistant/src/ContentRenderer.tsx
```diff
@@ -443,6 +443,54 @@ const ContentRenderer: React.FC<ContentRendererProps> = React.memo(
     const processedContent = useMemo(() => {
       if (!content) return null;
 
+      // First, check if content is a JSON response with error or success keys
+      try {
+        const parsedContent = JSON.parse(content.trim());
+
```
joaquimrocha: If this is the content you are checking the exception for, just wrap this one and use a conditional for the rest. Big try blocks may mask issues we're not catching for.

joaquimrocha (changes requested):
I think this PR is mixing many things. Maybe dividing in different commits would be better.

ai-assistant/src/ai/prompts.ts
```diff
@@ -22,6 +22,55 @@ RESOURCE CREATION GUIDELINES:
...
+1. Show: "Removing security context from container 'headlamp'"
+2. Use PUT with patch body: {"spec": {"template": {"spec": {"containers": [{"name": "headlamp", "securityContext": null}]}}}}
+
+To update an environment variable:
+1. Show: "Updating environment variable ENV_VAR to 'new-value'"
+2. Use PUT with patch body: {"spec": {"template": {"spec": {"containers": [{"name": "headlamp", "env": [{"name": "ENV_VAR", "value": "new-value"}]}]}}}}
+
+To add a label:
+1. Show: "Adding label 'environment: production'"
+2. Use PUT with patch body: {"metadata": {"labels": {"environment": "production"}}}
```
joaquimrocha: Maybe this belongs to the Kubernetes tool? We cannot have everything in the base prompt. The tools should provide the solutions for specific things like these.

ai-assistant/src/ai/prompts.ts
```diff
@@ -22,6 +22,55 @@ RESOURCE CREATION GUIDELINES:
...
+
+CONTROLLER-SPECIFIC CONSIDERATIONS:
+- Deployment: Most common, creates ReplicaSets which create Pods
+- StatefulSet: For stateful applications, pods have persistent identities
+- DaemonSet: One pod per node, can't change replica count
+- ReplicaSet: Usually managed by Deployments, rarely updated directly
+- Job: For one-time tasks, completed jobs shouldn't be updated
+- CronJob: For scheduled tasks, update the jobTemplate spec
```
joaquimrocha: Is this really needed? Seems like general k8s knowledge which LLMs should have, but if they don't then this is very limited?

ai-assistant/src/components/common/LogsButton.tsx
```diff
@@ -29,7 +29,6 @@ const LogsButton: React.FC<LogsButtonProps> = ({
     parts.push('Logs');
     return parts.join(' ');
   };
-
```
joaquimrocha: Do not remove this.

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -48,6 +49,8 @@ export const handleActualApiRequest = async (
   resourceInfo: string,
   targetCluster?: string // Allow specifying a specific cluster
 ) => {
+  console.log('url', 'method');
+  console.log(url, method, body, resourceInfo);
```
joaquimrocha: Left over?

ai-assistant/src/components/common/LogsButton.tsx
```diff
@@ -61,13 +60,15 @@ const LogsButton: React.FC<LogsButtonProps> = ({
...
-      />
+      {dialogOpen && (
+        <LogsDialog
+          open={dialogOpen}
+          onClose={() => setDialogOpen(false)}
+          logs={logs}
+          title={getTitle()}
+          resourceName={resourceName}
+        />
+      )}
```
joaquimrocha: Doesn't seem related to this commit.

ai-assistant/src/components/common/LogsButton.tsx
joaquimrocha, line 1: Many changes in this commit do not seem related to what the title says it does.

ai-assistant/src/ContentRenderer.tsx
```diff
@@ -215,16 +215,43 @@ const ContentRenderer: React.FC<ContentRendererProps> = React.memo(
           (className === 'language-json' || isJsonKubernetesResource(children));
 
         if (isYamlBlock && onYamlDetected && typeof children === 'string') {
-          // Try to parse as Kubernetes YAML
-          const parsed = parseKubernetesYAML(children);
-          if (parsed.isValid) {
+          // Handle multi-document YAML (documents separated by ---)
+          const yamlDocuments = children.split(/^---+$/m).filter(doc => doc.trim());
```
joaquimrocha: Why do you split the documents? If we have them in one doc maybe it's easier to apply? Or is this because the apply dialog you use has a limitation there?

ai-assistant/src/ai/prompts.ts
```diff
@@ -12,9 +12,25 @@ MANDATORY TOOL USAGE:
...
+- NEVER attempt to fetch logs from a multi-container pod without specifying the container name
+- Examples of user requests that specify containers:
+  - "get logs from pod-name container nginx" → kubernetes_api_request(url="/api/v1/namespaces/default/pods/pod-name/log?container=nginx", method="GET")
+  - "show logs for container sidecar in pod-name" → kubernetes_api_request(url="/api/v1/namespaces/default/pods/pod-name/log?container=sidecar", method="GET")
+- If you encounter an error about "container name must be specified", the error handler will automatically provide guidance to the user
+- IMPORTANT: Parse user requests carefully to detect if they're specifying a container name in their request:
+  - Look for patterns like "container [name]", "from container [name]", "[pod-name] [container-name]"
+  - If user specifies a container name, use it directly in the log URL
+  - If user doesn't specify a container and the pod has multiple containers, fetch pod details first to list available containers
```
joaquimrocha: Again, seems like you are assuming the kubernetes tool is enabled and this seems like something that should mostly belong there?

joaquimrocha (changes requested):
I think this PR mixes many things from different commits. They should be made atomic. Check also my comments in the different way we should approach some of the problems.

ai-assistant/src/ContentRenderer.tsx
ashu8912: umm so the below code is dependent on this parse that is why the entire block is in try

ai-assistant/src/ai/prompts.ts
ashu8912: moved this k8s tool

ai-assistant/src/ContentRenderer.tsx
joaquimrocha: Right. So in this case you set `let parsedContent = '';` or similar, and then check if the parse succeeded. Having such a big block in the try can cover issues you were not expecting to be firing an exception from.

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -327,37 +410,47 @@ export const handleActualApiRequest = async (
...
         aiManager.history.push({
+          success: true,
           role: 'tool',
           content: formattedResponse,
         });
       }
 
       return formattedResponse ?? 'ok';
     } catch (error) {
+      console.log(error);
```
joaquimrocha: Remove or properly log.

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -65,9 +93,16 @@ export const handleActualApiRequest = async (
...
+        error: true,
+        role: 'assistant',
+        content: `Error creating resource: ${error.message}`,
+      });
+
       return JSON.stringify({
         error: true,
-        message: `Error creating resource: ${error.message}`,
+        role: 'assistant',
+        content: `Error creating resource: ${error.message}`,
```
joaquimrocha: Seems like the return value is the same we are pushing to the history. Can you check if we should be instead handling the return value or check if we already are?

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -76,85 +111,118 @@ export const handleActualApiRequest = async (
...
+        role: 'tool',
+        success: true,
+        content: `${resourceIdentifier.kind} ${resourceIdentifier.name} deleted successfully.`,
+      });
+
+      return JSON.stringify({
+        success: true,
+        role: 'tool',
+        content: `${resourceIdentifier.kind} ${resourceIdentifier.name} deleted successfully.`,
+      });
```
joaquimrocha: Same comment about a return that is the same as the one we are pushing to history.

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -76,85 +111,118 @@ export const handleActualApiRequest = async (
...
       aiManager.history.push({
+        error: true,
+        role: 'assistant',
+        content: `Error deleting resource: ${error.message}`,
+      });
+      return JSON.stringify({
+        error: true,
         role: 'assistant',
         content: `Error deleting resource: ${error.message}`,
       });
```
joaquimrocha: Same comment about a return that is the same as the one we are pushing to history.

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -18,10 +18,11 @@ function deepMerge(target: any, source: any): any {
...
         // For objects, recursively merge
-        result[key] =
-          target[key] && typeof target[key] === 'object'
-            ? deepMerge(target[key], source[key])
-            : { ...source[key] };
+        // Initialize target[key] as empty object if it doesn't exist or isn't an object
+        if (!result[key] || typeof result[key] !== 'object' || Array.isArray(result[key])) {
+          result[key] = {};
+        }
+        result[key] = deepMerge(result[key], source[key]);
```
joaquimrocha, line 25: Doesn't belong to this commit.

ai-assistant/src/components/common/LogsButton.tsx
joaquimrocha: Still a lot not that doesn't belong to the commit.

ai-assistant/src/components/settings/ModelSelector.tsx
```diff
@@ -444,6 +447,14 @@ export default function ModelSelector({
...
   };
 
+  // Effect to handle opening provider selection after terms acceptance
+  useEffect(() => {
+    if (shouldOpenProviderSelection && hasAcceptedTerms()) {
+      setProviderSelectionOpen(true);
+      setShouldOpenProviderSelection(false);
+    }
+  }, [savedConfigs?.termsAccepted, shouldOpenProviderSelection]);
+
```
joaquimrocha: I see the description of the issue in the commit but I am a bit confused as to why this is needed, or how this is preventing the problem you described. Can you double check what the exact problem is? When the user accepts the terms -> provider selection should be opened, which seems to be what the code was doing before. Maybe there's a moment where this is not the case but I find it hard to see a solution that has a setProviderSelectionOpen and setShouldOpenProviderSelection.

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -75,25 +77,64 @@ export const handleActualApiRequest = async (
...
-        },
-        {
-          startMessage: `Creating ${resource.kind || 'resource'} in cluster ${cluster}...`,
-          cancelledMessage: `Cancelled creating resource in cluster.`,
-          successMessage: `${resource.kind || 'Resource'} created successfully.`,
-          errorMessage: `Failed to create resource.`,
-        }
-      );
+      const response = await new Promise(resolve => {
+        clusterAction(
```
joaquimrocha, line 98: Could we use await clusterAction?

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -267,6 +331,39 @@ export const handleActualApiRequest = async (
         });
       } catch (error) {
         console.log('Error in clusterRequest:', error);
+
+        // Check if this is a multi-container pod logs error
+        if (
+          isLogRequest(cleanedUrl) &&
+          error.message &&
+          error.message.includes('a container name must be specified')
```
joaquimrocha: error.message?.includes(....)

joaquimrocha: Also, could this be caught earlier by us? Certainly the request will be missing a parameter.  (before the request is sent, but sorry if this is already the case)

ai-assistant/src/helper/apihelper.tsx
```diff
@@ -267,6 +331,39 @@ export const handleActualApiRequest = async (
...
+              error: false,
+              role: 'assistant',
+              content: errorContent,
+            });
+
+            return JSON.stringify({
+              error: false,
+              role: 'assistant',
+              content: errorContent,
+            });
```
joaquimrocha: This is becoming a pattern. We need to see if we can optimize it.

ai-assistant/src/modal.tsx
```diff
@@ -353,6 +354,66 @@ export default function AIPrompt(props: {
...
+            operation = 'operation';
+        }
+      }
+
+      // Extract error details
+      const errorMessage = error?.message || error?.error || 'Unknown error occurred';
+      const statusCode = error?.status || error?.statusCode;
+
+      // Build error content
+      let errorContent = `Resource ${operation} failed: ${errorMessage}`;
```
joaquimrocha: This will be complicated to handle when we introduce translations for plugins.

ai-assistant/src/components/common/LogsButton.tsx
ashu8912: fixed now it only contains log related fixes

ai-assistant/src/modal.tsx
ashu8912: updated

ai-assistant/src/helper/apihelper.tsx
ashu8912: updated

ai-assistant/src/helper/apihelper.tsx
ashu8912: fixed

ai-assistant/src/helper/apihelper.tsx
ashu8912: I think clusterAction doesn't return a promise

ai-assistant/src/components/settings/ModelSelector.tsx
ashu8912: fixed

ai-assistant/src/helper/apihelper.tsx
ashu8912: fixed

ai-assistant/src/helper/apihelper.tsx
ashu8912: fixed

ai-assistant/src/helper/apihelper.tsx
ashu8912: fixed

ai-assistant/src/helper/apihelper.tsx
ashu8912: fixed

ai-assistant/src/ContentRenderer.tsx
```diff
@@ -9,6 +9,79 @@ import { LogsButton, YamlDisplay } from './components';
...
+    const logsData = JSON.parse(jsonString);
+
+    // Validate required fields
+    if (!logsData.data || !logsData.data.logs) {
+      return { success: false, error: 'Invalid logs data structure: missing required fields' };
+    }
+
+    return { success: true, data: logsData };
+  } catch (error) {
+    if (error instanceof SyntaxError) {
```
joaquimrocha: This will catch errors if you have a syntax error in the Js code.
I think you need to try for just the JSON.parse .

ai-assistant/src/ContentRenderer.tsx
```diff
@@ -312,7 +311,6 @@ const ContentRenderer: React.FC<ContentRendererProps> = React.memo(
             );
           }
         }
-
```
joaquimrocha: Do not remove this.

ai-assistant/src/langchain/LangChainManager.ts
```diff
@@ -963,6 +972,7 @@ Format your response to make the errors prominent and actionable.`,
       role: 'system',
       content:
         'REMINDER: Never suggest kubectl or command line tools. Always use the kubernetes_api_request tool or explain UI actions. The user is using a web dashboard and cannot access the command line.',
+      isDisplayOnly: true, // Mark as display-only to prevent sending to LLM
```
joaquimrocha: You are marking this as display-only but the message looks like it should be shared with the LLM...

ashu8912: updated

ai-assistant/src/ContentRenderer.tsx
ashu8912: updated

ai-assistant/src/ContentRenderer.tsx
ashu8912: updated

ai-assistant/src/ContentRenderer.tsx
ashu8912: updated

ai-assistant/src/ai/prompts.ts
ashu8912: updated

ai-assistant/src/components/common/LogsButton.tsx
ashu8912: updated

ai-assistant/src/helper/apihelper.tsx
ashu8912: fixed

ai-assistant/src/components/common/LogsButton.tsx
ashu8912: fixed

ai-assistant/src/ContentRenderer.tsx
ashu8912: fixed

ai-assistant/src/ai/prompts.ts
ashu8912: fixed

ai-assistant/src/helper/apihelper.tsx
ashu8912: updated

Issue #359, opened 2025-08-08
A11y issues in the Plugin Catalog


joaquimrocha:
I have noticed the following issues when using the Plugin Catalog, which should be fixed:
- [ ] When clicking install on a plugin, for a brief moment, we see the tooltip as "null:null"
- [ ] After a plugin is installed, the snackbar showing that it's installed will display its buttons as dark and on a dark background, making them almost impossible to read.
- [ ] The spinner in the button while installing, also lacks contrast IIRC

I recommend doing other checks for a11y in this plugin.

PR #360, opened 2025-08-08
LoadingButton: fix poor contrast on loading install button

vyncent-t:
This PR fixes the poor contrast of the loading spinner when clicking install on a plugin. The color should now default to the visible contrast for all themes.

Related to issue: https://github.com/headlamp-k8s/plugins/issues/359

##  After

image

image


illume (commented):
"plugin-catalog:" is missing at start of the context

joaquimrocha (approved):

PR #369, opened 2025-08-16
karpenter: Add Map View for Resources

SinghaAnirban005:
## Map View for Karpenter Resources

[Screencast from 2025-08-23 03-07-31.webm](https://github.com/user-attachments/assets/8d7d35bd-e014-4824-bfe8-fe245adfdc65)

Part of https://github.com/kubernetes-sigs/headlamp/issues/3231

### DIFF
--- a/karpenter/src/index.tsx
+++ b/karpenter/src/index.tsx
@@ -14,7 +14,14 @@
  * limitations under the License.
  */
 import { addIcon } from '@iconify/react';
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks!

Instead of fixing the commit problems in the last commit, can you please rebase the earlier commits so that the checks pass?

PR #400, opened 2025-10-02
flux: Define all resource classes that extend KubeObject, refactor to use useGet and useList methods

sniok:


illume (changes requested):
Can you please provide a PR description/steps to test/issue link?

PR #406, opened 2025-10-08
flux: add external artifacts

### LINKED ISSUE #405: Flux: add `ExternalArtifact` views
Flux v2.7 added new CRD's which should be supported by this plugin.

- <https://fluxcd.io/blog/2025/09/flux-v2.7.0/#artifact-generators>
- #408

viceice:
- closes #405
- https://fluxcd.io/blog/2025/09/flux-v2.7.0/#artifact-generators
 
have not yet tested

### DIFF
--- a/flux/src/overview/index.tsx
+++ b/flux/src/overview/index.tsx
@@ -35,6 +35,7 @@ import { alertNotificationClass } from '../notifications/NotificationList';
 import { receiverNotificationClass } from '../notifications/NotificationList';
 import {
   bucketRepositoryClass,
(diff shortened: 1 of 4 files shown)


viceice:
ok, need to clone to wsl as the commands don't work on windows 🙃

viceice:
works 💪

viceice:
Would be nice to have this released soon, so i can continue using `ExternalArtifacts`.
Without this `Kustomizations`  which use `ExternalArtifact` as source ref will crash the UI.

Issue #424, opened 2025-10-20
app-catalog: description display issues


illume:
There are display issues with the app catalog.

Some parts are cut off.

There's more issues when the zoom isn't default (an a11y requirement is it works with 2x zoom).

PR #441, opened 2025-11-03
app-catalog: Fix card styling on zoom

skoeva:
This change adjusts the card styling so there are no longer issues present on high zoom levels.

Fixes: #424 

### Testing
- [x] Open App Catalog and play around with the zoom
- [x] Note that the description and install button don't get clipped

image


illume (approved):
🎉 thanks!

PR #457, opened 2025-11-25
app-catalog: Replace plugin settings with verified toggle

skoeva:
This change replaces the settings button in the app catalog with the toggle for displaying only verified apps (similar to the plugin catalog). The AppCatalogSettings component is thus no longer necessary and removed here.

### Screenshot

image


illume (approved):
🎉 thanks!

PR #458, opened 2025-12-04
Bump mdast-util-to-hast to 13.2.1

skoeva:
This change bumps the mdast-util-to-hast dependency in the plugins to 13.2.1.


gambtho (approved):
lgtm

Issue #468, opened 2025-12-14
Proposal: Donate Knative Plugin to headlamp-k8s/plugins


kahirokunn:
## Summary
I’d like to donate a **Knative plugin for Headlamp** to this repository (`headlamp-k8s/plugins`) so it can be maintained and released as an official/recommended plugin.

- Source (current implementation): https://github.com/kahirokunn/headlamp-plugins/tree/main/knative

## What this plugin adds
This plugin adds a new sidebar item **“Knative”** and provides UI features for **Knative Services (KService)** operations, including:
- List & details view for KServices
- Traffic splitting updates
- Concurrency settings updates
- Redeploy / restart actions
- Networking references (e.g., HTTPRoutes)

## Demo
- Video: https://github.com/user-attachments/assets/a5c9cac3-f711-4306-84cd-83178fc876e0

## License
Apache-2.0

## Proposed next step
If maintainers agree, I can open a PR that:
- Adds the `knative/` plugin directory to this repo
- Adds an entry for `knative` in the root README “Current Plugins” table

PR #473, opened 2025-12-23
Init Knative plugin

kahirokunn:
fixes: https://github.com/headlamp-k8s/plugins/issues/468

## What this plugin adds
This plugin adds a new sidebar item **“Knative”** and provides UI features for **Knative Services (KService)** operations, including:
- List & details view for KServices
- Traffic splitting updates
- Concurrency settings updates
- Redeploy / restart actions
- Networking references (e.g., HTTPRoutes)

## Demo
- Video: 

https://github.com/user-attachments/assets/a5c9cac3-f711-4306-84cd-83178fc876e0

## Testing this plugin locally

To test this plugin with a local Knative environment, I recommend following this setup guide:
[Extending Knative Service with Envoy Gateway Integration - DEV Community](https://dev.to/kahirokunn/extending-knative-service-with-envoy-gateway-integration-56ak)

This guide will help you:
- Set up a local Knative environment
- Configure the necessary components to test Knative Services

### DIFF
--- a/knative/src/components/kservices/List.tsx
+++ b/knative/src/components/kservices/List.tsx
@@ -65,20 +65,34 @@ function getTags(svc: KService): string[] {
   return tags;
 }
 
-function getLatestRevisionShort(svc: KService): string {
-  const name = svc.metadata.name;
-  const latestRevisionFull =
-    svc.status?.latestCreatedRevisionName ?? svc.status?.latestReadyRevisionName ?? '';
-
-  if (!latestRevisionFull) {
+function getRevisionShort(svc: KService, revisionName: string | undefined): string {
+  if (!revisionName) {
     return '';
   }
 
-  if (latestRevisionFull.startsWith(`${name}-`)) {
-    return latestRevisionFull.slice(name.length + 1);
+  const name = svc.metadata.name;
+  if (revisionName.startsWith(`${name}-`)) {
+    return revisionName.slice(name.length + 1);
   }
 
-  return latestRevisionFull;
+  return revisionName;
+}
+
+function getReadyCondition(svc: KService): { status: string; reason?: string } | null {
+  const conditions = svc.status?.conditions;
+  if (!conditions) {
+    return null;
+  }
+
+  const readyCondition = conditions.find(c => c.type === 'Ready');
+  if (!readyCondition) {
+    return null;
+  }
+
+  return {
+    status: readyCondition.status || 'Unknown',
+    reason: readyCondition.reason,
+  };
 }
 
 function getServiceUrls(svc: KService, domainByServiceKey: Record<string, string[]>): string[] {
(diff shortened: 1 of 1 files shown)


knative/src/components/kservices/detail/header/KServiceLogsHeaderButton.tsx
```diff
@@ -0,0 +1,866 @@
...
+    // Pod set changes (scale up/down) are handled incrementally so logs don't fully refresh.
+    const streamConfigKey = [
+      activity.id,
+      serviceName,
+      namespace,
+      cluster,
+      selectedPodName,
+      selectedContainers.join('|'),
+      showTimestamps ? 'ts1' : 'ts0',
+      follow ? 'f1' : 'f0',
```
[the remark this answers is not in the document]
kahirokunn: That's not true.

knative/src/components/kservices/List.tsx
```diff
@@ -0,0 +1,450 @@
...
+  );
+
+  const ingressClassLabel = React.useMemo(() => {
+    if (!ingressClassData || ingressClassData.length <= 1) {
+      return 'Ingress class';
+    }
+    return 'Ingress classes';
+  }, [ingressClassData]);
+
+  const columns = React.useMemo<
```
yolossn, line 266: We usually try to keep the list table columns similar to the `kubectl get` response. 
image

kahirokunn:
@yolossn Yes, exactly. I'm not aiming to submit something perfect in the initial commit. I'd like to start with the core components like services and networking first.
For adding support for the other CRDs, I think it would be great to discuss the best approach through Issues. That way we can determine the right way to extend the plugin going forward. 🙂

knative/src/components/kservices/List.tsx
kahirokunn: Thanks for your review 🙂 
If we were only outputting the raw results of `kubectl get`, it would just be a matter of developing an alias for the CRD menu built into the headlamp. However, I don't believe any user would install a plugin just for that. Therefore, I think we should provide added value. What do you think?
Or does this mean you want to display additional columns that are shown in `kubectl get` but aren't displayed here?

kahirokunn:
> 
> > Yes, exactly. I'm not aiming to submit something perfect in the initial commit. I'd like to start with the core components like services and networking first. 
> > For adding support for the other CRDs, I think it would be great to discuss the best approach through Issues. That way we can determine the right way to extend the plugin going forward. 🙂
> 
> Cool. Yeah, this is an initial version and we can discuss and make changes in future iterations. Let's collect these ideas and notes then we can keep discussing and getting further feedback as well from others using it?

Yes, let's do that!😆

knative/src/components/kservices/List.tsx
illume: Without speaking for him, I think he means we try to be consistent with the columns that kubectl uses when possible. Because there's usually been a lot of thinking and years of tweaks to which columns are displayed. As well, there's a big group of people who are used to that. But of course where are very good reasons we have diverged from the columns kubectl chooses.

illume:
>  (I want to be able to run CI myself without bothering everyone.)

After your first PR is merged it runs the CI for you without approval.

PR #481, opened 2026-01-16
opencost plugin: fix correct endpoint for OpenCost API

guidemetothemoon:
After kubecost became opencost there have been changes to its API, where model and compute were removed from the allocation endpoint. Ref. https://github.com/opencost/opencost/issues/2845#issuecomment-2249815631
Currently opencost plugin doesn't populate any data with the original endpoint.
This PR fixes the endpoint according to the opencost API and testing confirms that the plugin works as expected after this change and data is populated correctly.

### DIFF
--- a/opencost/src/request.tsx
+++ b/opencost/src/request.tsx
@@ -29,7 +29,7 @@ export function fetchOpencostData(
   resource: string,
   accumulate: boolean
 ) {
-  const url = `/api/v1/namespaces/${namespace}/services/${serviceName}/proxy/model/allocation/compute?window=${window}&aggregate=${resource}&step=1d&accumulate=${accumulate.toString()}`;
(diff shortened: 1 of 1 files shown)


yolossn:
Hey, I just installed opencost in my dev cluster (using helm) and checked by installing the plugin from plugin store. The plugin is working for me. Which version of opencost are you using?

image

image

guidemetothemoon:
@yolossn even with opencost version 2.5.3 it doesn't work for me, still hangs in both desktop and in-cluster version of headlamp and I see 404 requests in network logs. If I update the endpoint as done in this PR and run headlamp desktop it works as expected.

How does your opencost plugin configuration look like when you're testing?
And are you able to see how the opencost API endpoint look like during your test?

PR #492, opened 2026-01-22
Knative: Fixed package name

### LINKED ISSUE #487: Knative Plugin Install Button Not Working
## Description
The Install button for the Knative plugin is not functioning properly.

Image

mudit06mah:
## Summary
Fixes installation button by changing to appropriate package name

## Related Issue
Fixes #487 

## Changes
Changed name in `package.json` from `knative-headlamp-plugin` to `knative`, this should generate correct download link.

### DIFF
--- a/knative/package.json
+++ b/knative/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "knative-headlamp-plugin",
+  "name": "@headlamp-k8s/knative",
   "version": "0.1.0",
   "description": "A Headlamp plugin for visualizing and managing Knative Services. Provides GUI functionality to list and view service details, edit traffic splitting, update concurrency settings, perform redeploy/restart operations, and reference external/internal HTTPRoutes.",
   "scripts": {


knative/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "knative-headlamp-plugin",
+  "name": "knative",
```
kahirokunn: Thank you!
However, based on other package.json files like keda or karpenter, the correct code is likely as follows:

```suggestion
  "name": "@headlamp-k8s/knative",
```

kahirokunn (changes requested):

knative/package.json
mudit06mah: @kahirokunn  yes you're correct, I took prometheus' `package.json` as a reference as it is just `"name": prometheus`

Should that also be changed in the future to match the convention?

kahirokunn:
LGTM 😊

PR #493, opened 2026-01-22
fix: use error color for failed reconciliations

### LINKED ISSUE #401: Flux: The colors for failed reconciliations are incorrect
I have just noticed that a chart showing that 1/1 failed resources is showing as orange instead of red. This gives the feeling that the issue is not that problematic. We need to change it to the error color.

Screenshot showing the chart

Tusharjamdade:
Fixes #401 

This PR fixes the color used for failed reconciliations in the chart UI. Previously, a fully failed state (e.g., 1/1 failed resources) was displayed using the warning (orange) color, which downplayed the severity of the issue.

The chart now correctly uses the error (red) color for failed reconciliations, making critical states more visually distinct and easier to identify.

### DIFF
--- a/flux/src/overview/index.tsx
+++ b/flux/src/overview/index.tsx
@@ -539,7 +539,7 @@ function FluxOverviewChart({ resourceClass }) {
         {
           name: 'failed',
           value: adjustedFailedPercent,
-          fill: '#DC7501',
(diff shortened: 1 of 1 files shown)


illume (approved):
thanks!

Issue #438, opened 2025-10-29
ai-assistant: Add local provider API key property


pantherra:
Hello!
When choosing local AI provider - the API key field is missing. Sometimes local AI API providers can be configured with API key protection of the endpoint.

PR #495, opened 2026-01-24
ai-assistant: Added API Key for Local Provider

aadhil2k4:
fixes: #438 

This PR adds an optional API key field for the Local AI Provider. Some local AI endpoints are protected by API Keys, and the absence of this field prevents users from authenticating. 
This change makes the Local provider consistent with other providers while keeping the API key optional

image

### DIFF
--- a/ai-assistant/src/config/modelConfig.ts
+++ b/ai-assistant/src/config/modelConfig.ts
@@ -189,6 +189,13 @@ export const modelProviders: ModelProvider[] = [
     description: 'Integration with locally hosted models (Ollama or similar)',
     fields: [
       {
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for this!

Can you please squash the commits and change the git commit message to match [the style we use](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines)?

Something like this:
> ai-assistant: modelConfig: Add API Key config for Local Provider

Please see the open review comment?

aadhil2k4:
@illume I have implemented the suggested changes. Can you please review it. Thank you!!

PR #514, opened 2026-02-05
Update dependencies

skoeva:


skoeva (commented):
Let's wait on #527 first

PR #519, opened 2026-02-08
fix(app-catalog): prevent EditorDialog crash on invalid YAML input

Darshika11b:
### What this PR does
- Prevents EditorDialog from crashing on invalid YAML input
- Adds null-safety for optional chart and release fields
- Does not change existing UI or behavior

### Why
In real-world clusters, Helm releases may have missing metadata or invalid values.yaml.
This caused EditorDialog to crash. The fix ensures graceful handling.

### Testing
- Opened EditorDialog with valid and invalid YAML
- Verified no UI regressions


illume:
Hello. 

I'll mark this as draft to make it easier on reviewers. If you want to continue this, please mark it as ready to review after pushing changes.

We can't accept PRs without the DCO.

PR #520, opened 2026-02-08
fix(app-catalog): guard against missing chart metadata in EditorDialog

Darshika11b:
### Summary
Prevents EditorDialog from crashing when Helm chart metadata is missing.

### What changed
- Added defensive checks for `release.chart.metadata`
- Prevents runtime crash when metadata is undefined
- No UI changes

### Why
Some Helm releases do not include complete chart metadata.
Accessing `metadata.name` without checks caused EditorDialog to crash.

### Testing
- Verified EditorDialog opens normally with valid metadata
- Verified graceful handling when metadata is missing

### DIFF
--- a/app-catalog/src/components/releases/EditorDialog.tsx
+++ b/app-catalog/src/components/releases/EditorDialog.tsx
@@ -48,20 +48,25 @@ export function EditorDialog(props: {
     isUpdateRelease,
     handleUpdate,
   } = props;
+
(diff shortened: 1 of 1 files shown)


illume:
Hello. 

I'll mark this as draft to make it easier on reviewers. If you want to continue this, please mark it as ready to review after pushing changes.

We can't accept PRs without the DCO.

PR #571, opened 2026-03-22
Plugin catalog UX improvements

joaquimrocha:
These changes show an indicator in the plugin installation list.

Image showing the UI change

It also updates the headlamp-plugin + adds a confirmation dialog if the user attempts to delete a plugin.


illume (changes requested):
Thanks for these changes.


Can you please address the open review comments?

Issue #536, opened 2026-02-24
ai-assistant: Update available models in the UI


skoeva:
I believe these are hardcoded in the UI. The list should be reviewed and updated to accurately reflect the available offerings from each provider, as some older models may have been deprecated by now (e.g., older versions of GPT-3.5) and newer models have not yet been added (e.g., newest Opus/Sonnet models from Anthropic).

PR #576, opened 2026-03-25
ai-assistant: Update list of available models

tboerger:
Remove deprecated old models and add current regular models. Fixes https://github.com/headlamp-k8s/plugins/issues/536.


tboerger:
@illume I have executed it and amended my changes.

PR #579, opened 2026-03-28
Flux block changing owned resources

joaquimrocha:
Following a chat with @stefanprodan at KubeCon, it was clear that one improvement for the Flux in Headlamp UX would be for Headlamp not to show options that make the Flux-managed resources drift from the configured state in gitops.
So, these changes do that by removing the edit, scale, delete, etc. actions from resources that have a label showing they are managed by Flux.

Here are some screenshots:

**Namespace that's managed by Flux**
Namespace that's managed by Flux

**Namespace not managed by Flux**
Namespace not managed by Flux


stefanprodan (approved):
LGTM

Great improvement! Thanks @joaquimrocha 🏅

skoeva (changes requested):
following up - just needs to be signed and formatted with prettier

joaquimrocha:
on it

ashu8912 (approved):
LGTM!!!

illume (approved):
Yeah, nice changes. Thanks.

PR #580, opened 2026-03-28
Flux improve empty states

joaquimrocha:
I noticed that the Flux plugin still showed that all CRDs were fine when Flux was not even installed in the cluster.
Also, there is no need for all the sidebar entries if Flux is not installed, so these changes also hide all sidebar entries but the Overview.

### DIFF
--- a/flux/src/common/Resources.tsx
+++ b/flux/src/common/Resources.tsx
@@ -119,6 +119,13 @@ export class ImageRepository extends KubeObject {
   static isNamespaced = true;
 }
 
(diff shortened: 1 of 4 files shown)


skoeva (changes requested):
following up - just needs to be signed and formatted with prettier

joaquimrocha:
checking now

skoeva (commented):
fixed the lint error in the CI

illume (approved):
Awesome. Thanks!!

PR #583, opened 2026-03-30
kyverno: ComplianceBadge: UI improvements and navigation fix

itvi-1234:
**Summary**

This PR improves the UX of the Kyverno Compliance badge popup by fixing interaction issues and navigation bugs found during local testing.

Changes:

1. Overlay fix: Added `ClickAwayListener` to ensure the compliance popup closes when clicking anywhere outside of it.

2. Navigation fix: Updated the "View all violations" link to use Router.createRouteURL with the correct route name (KyvernoViolations). Previously it was using a hardcoded path which caused a 404 in multi-cluster environments because the cluster prefix was missing.

3. Branding UI: Added a yellow underline to the popup title to match Headlamp's visual aesthetic.

### DIFF
--- a/kyverno/src/components/ComplianceBadge.tsx
+++ b/kyverno/src/components/ComplianceBadge.tsx
@@ -15,6 +15,7 @@
  */
 
 import { Icon } from '@iconify/react';
+import { K8s } from '@kinvolk/headlamp-plugin/lib';
 import {
   Badge,
   Box,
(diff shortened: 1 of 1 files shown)


kyverno/src/components/ComplianceBadge.tsx
```diff
@@ -85,32 +93,59 @@ export function ComplianceBadge() {
...
-          </Typography>
-          <Box sx={{ display: 'flex', flexDirection: 'column', gap: 0.5 }}>
-            <Typography variant="body2" color={compliancePct >= 90 ? 'success.main' : compliancePct >= 70 ? 'warning.main' : 'error.main'}>
-              {compliancePct}% compliant
+        <ClickAwayListener onClickAway={handleClose}>
+          <Paper elevation={8} sx={{ p: 2, minWidth: 220 }}>
+            <Typography
+              variant="subtitle2"
+              sx={{
+                borderBottom: '2px solid #f9da4c',
```
joaquimrocha: I would skip this change. It's not related to fixing the closing of the dialog + also doesn't take into account that you can have different themes and then the hardcoded color here will not look good.

joaquimrocha (changes requested):
I would not introduce any hardocoded color changes. Please make the PR about closing the dialog on clicking out of the, and about fixing the links. Two commits.

PR #590, opened 2026-04-02
fix: re-detect Kyverno CRDs on cluster switch

itvi-1234:
**Summary**
The useKyvernoCRDs hook was only running once on plugin mount. This meant that if the user switched between different Kubernetes clusters (where Kyverno might or might not be installed), the plugin's status (and sidebar visibility) would remain stale from the first detected cluster.

**Changes**
1.Integrated K8s.useCluster() hook to track active cluster changes.
2.Added cluster as a dependency to the detection useEffect.
3.Implemented a loading state reset on cluster switch to ensure fresh CRD detection.

### DIFF
--- a/kyverno/src/hooks/useKyvernoCRDs.ts
+++ b/kyverno/src/hooks/useKyvernoCRDs.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { ApiProxy } from '@kinvolk/headlamp-plugin/lib';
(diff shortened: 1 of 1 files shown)


joaquimrocha:
Hi @itvi-1234 . Thanks. Unfortunately I don't have any more time today for Headlamp maintenance.
Happy if others interested in this plugin chime in on your changes.

itvi-1234:
No worries @joaquimrocha :rocket:

PR #604, opened 2026-04-06
Prometheus: gate Job metrics/actions on batch/v1

### LINKED ISSUE #597: Volcano: Prometheus metrics section incorrectly appears on Volcano Jobs
**Problem:**
The Prometheus metrics section, intended for Kubernetes batch/v1 Job, is also showing up on Volcano Job details view. This is because the check only matches on kind: Job without verifying the apiVersion, causing a false match with Volcano's batch.volcano.sh/v1alpha1 Job.

Image

**Fix:**
In the Prometheus plugin, update the condition to check both kind and apiVersion, not just kind alone. This [check](https://github.com/headlamp-k8s/plugins/blob/fa4f70f77645192f4dd533a6b2c7aacc694bfaa7/prometheus/src/index.tsx#L26) in prometheus plugin has to be updated.

cc: @mahmoudmagdy1-1

mahmoudmagdy1-1:
## Summary
Volcano uses `kind: Job` with a different API version, so kind only checks can incorrectly show Prometheus UI on Volcano Job details. (mentioned in #597)
## Changes
- add a shared supportsPrometheusMetrics helper in prometheus/src/util.ts
- use a small API-version rule map so Job is treated as supported only when apiVersion === 'batch/v1'
- reuse the shared helper in:
  - `prometheus/src/index.tsx` (metrics + header action)
  - `prometheus/src/components/VisibilityButton/VisibilityButton.tsx` (button rendering)
- add focused unit tests in prometheus/src/util.test.ts covering:
  - Kubernetes Job
  - Volcano Job
  - unsupported kinds
  - missing resource input

Closes #597

### DIFF
--- a/prometheus/src/index.tsx
+++ b/prometheus/src/index.tsx
@@ -11,8 +11,12 @@ import { GenericMetricsChart } from './components/Chart/GenericMetricsChart/Gene
 import { KedaChart } from './components/Chart/KedaChart/KedaChart';
 import { Settings } from './components/Settings/Settings';
 import { VisibilityButton } from './components/VisibilityButton/VisibilityButton';
-import { ChartEnabledKinds, PLUGIN_NAME } from './util';
-import { getNodeClaimChartConfigs, getNodePoolChartConfigs } from './util';
+import {
+  getNodeClaimChartConfigs,
+  getNodePoolChartConfigs,
+  PLUGIN_NAME,
+  supportsPrometheusMetrics,
+} from './util';
 
 type SectionWithId = { id: string };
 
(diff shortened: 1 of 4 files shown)


prometheus/src/index.tsx
```diff
@@ -160,7 +160,7 @@ registerDetailsViewHeaderActionsProcessor(function addPrometheusMetricsButton(re
   }
 
   // If the action is not supposed to be added, we do nothing.
-  if (!ChartEnabledKinds.includes(resource?.jsonData?.kind)) {
+  if (!ChartEnabledKinds.includes(resource?.jsonData?.kind) || isUnsupportedBatchJob(resource)) {
```
yolossn: Instead of adding extra check for unsupported batch job, we should change the way the actual check happens here.

mahmoudmagdy1-1: Done, can you please check again?

prometheus/src/util.ts
```diff
@@ -154,6 +154,42 @@ export function getPrometheusResolution(cluster: string): string {
...
+
+  const requiredApiVersion = resourceApiVersionRules[kind];
+  if (requiredApiVersion) {
+    return getResourceApiVersion(resource) === requiredApiVersion;
+  }
+
+  return true;
+}
+
 export const ChartEnabledKinds = [
```
yolossn: now `ChartEnabledKinds` is only used by `supportsPrometheusMetrics` so we can drop the export here.

mahmoudmagdy1-1: Done, Thanks for review

PR #620, opened 2026-04-11
fix: use strategic-merge PATCH instead of GET + deepMerge + PUT in ai-assistant

atharrva01:
I found a pretty bad silent data-loss bug in the ai-assistant's PUT handler. It was GETting the live resource, running it through a homegrown `deepMerge`, and PUTting the whole thing back. The array branch just did a verbatim replace, so something as innocent as "update the image to v2" would wipe every `env`, `volumeMount`, sidecar container, and volume from the Deployment. The API server accepted it without complaint, pods restarted broken, and the UI showed a green success toast the whole time.

I fixed this by switching to a single PATCH with `application/strategic-merge-patch+json` and removing `deepMerge` entirely. Now the raw LLM patch goes straight to the server, which handles array merging correctly by merge key (`name` for containers/env, `mountPath` for volumeMounts, etc.). This also gets rid of the TOCTOU window that existed between the GET and PUT.

Happy path is unchanged, the snackbar, callbacks, and overall flow all work the same way.

### DIFF
--- a/ai-assistant/src/helper/apihelper.tsx
+++ b/ai-assistant/src/helper/apihelper.tsx
@@ -4,34 +4,6 @@ import { getCluster } from '@kinvolk/headlamp-plugin/lib/Utils';
 import YAML from 'yaml';
 import { isLogRequest, isSpecificResourceRequestHelper } from './index';
 
-// Deep merge function to merge patch with current resource
-function deepMerge(target: any, source: any): any {
-  const result = { ...target };
-
-  for (const key in source) {
-    if (source[key] === null) {
-      // If source value is null, remove the property
-      delete result[key];
-    } else if (source[key] !== undefined) {
-      if (Array.isArray(source[key])) {
-        // For arrays, replace entirely
-        result[key] = [...source[key]];
-      } else if (typeof source[key] === 'object' && source[key] !== null) {
-        // For objects, recursively merge
-        // Initialize target[key] as empty object if it doesn't exist or isn't an object
-        if (!result[key] || typeof result[key] !== 'object' || Array.isArray(result[key])) {
-          result[key] = {};
-        }
-        result[key] = deepMerge(result[key], source[key]);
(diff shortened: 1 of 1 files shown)


yolossn:
@atharrva01  looks like there are some formatting issues, can you fix them?

yolossn:
Steps to test:

1. Create a dummy deployment for testing this

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: patch-test-config
  namespace: default
data:
  app.conf: |
    greeting=hello
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: patch-test
  namespace: default
  labels:
    app: patch-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: patch-test
  template:
    metadata:
      labels:
        app: patch-test
    spec:
      containers:
        - name: web
          image: nginx:1.25
          ports:
            - containerPort: 80
          env:
            - name: LOG_LEVEL
              value: info
            - name: FEATURE_FLAG
              value: "true"
          volumeMounts:
            - name: config
              mountPath: /etc/app-config
        - name: sidecar
          image: busybox:1.36
          command: ["sh", "-c", "while true; do echo sidecar alive; sleep 30; done"]
      volumes:
        - name: config
          configMap:
            name: patch-test-config

```

2. Make sure the AI assistant is configured and try these prompts
>  Update the web container image in deployment patch-test in namespace default to nginx:1.27.

The apply patch pop up will appear as below
image


3. click apply and go check the deployment. Only the image should be changed

yolossn (changes requested):
tested this and works as expected, once the formatting issues are fixed good to merge.

atharrva01:
hey @yolossn, thanks for testing! Fixed the formatting issue, it was an extra blank line left after removing the deepMerge function. Prettier check passes now.

yolossn (approved):

PR #627, opened 2026-04-14
kubeflow: Release 0.1.0-alpha

illume:
Signed-off-by: René Dudfield <renedudfield@microsoft.com>

### DIFF
--- a/kubeflow/0.1.0-alpha/artifacthub-pkg.yml
+++ b/kubeflow/0.1.0-alpha/artifacthub-pkg.yml
@@ -1,0 +1,34 @@
+version: 0.1.0-alpha
+name: "headlamp_kubeflow"
+displayName: Kubeflow
+createdAt: "2026-04-14T10:19:00Z"
+logoURL: "https://raw.githubusercontent.com/kubeflow/manifests/refs/heads/master/common/oauth2-proxy/base/kubeflow-logo.svg"
+description: A UI for viewing and managing Kubeflow resources in Headlamp.
+changes:
+  - kind: added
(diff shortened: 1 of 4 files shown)


kubeflow/0.1.0-alpha/artifacthub-pkg.yml
```diff
@@ -0,0 +1,34 @@
...
+  - kind: added
+    description: "Initial release of the Headlamp plugin for Kubeflow."
+  - kind: added
+    description: "Auto-detect installed Kubeflow components by probing cluster CRDs."
+  - kind: added
+    description: "Notebooks support: Notebook, Profile, and PodDefault resources (kubeflow.org/v1)."
+  - kind: added
+    description: "Notebook log viewer in detail side-pane."
+  - kind: added
+    description: "Pipelines support: Pipeline and PipelineVersion resources (pipelines.kubeflow.org/v2beta1)."
```
alokdangre: it does not support for pipeline yet, just checks if it is persent,

alokdangre: same for all below @illume

alokdangre (commented):
also in yml file didn't mentioned much about the notebooks component,
like about notebooks dashboard

kubeflow/0.1.0-alpha/artifacthub-pkg.yml
illume: updated. @alokdangre is it ok?

alokdangre (approved):

illume:
Thanks @alokdangre

PR #628, opened 2026-04-14
fix(kyverno): fix silent matching failure for namespaced Policy Reports

itvi-1234:
### Summary
This PR fixes a critical bug where Headlamp's Kyverno plugin failed to display report results for Namespaced Policies (e.g. `Policy` kind). Kyverno prefixes namespaced policy results with the namespace string internally (e.g., `default/require-labels`), causing the strict string comparison in the frontend to silently fail and mistakenly render `Associated Report Results (0)`.

### Changes
- Updated `AssociatedReportsSection` in `src/components/PolicyViewer.tsx` to conditionally accept a `namespace` prop.
- Added a smart-match helper (`matchesPolicy`) that checks both the raw `policyName` strings and `"namespace/policyName"` strings to appropriately capture the varying reporting formats from Kyverno's backend.
- Result payload in UI now accurately represents all validation scans rather than improperly skipping namespaced targets.

### Steps to Test
1. Make sure to have Kyverno deployed with at least one violated `Policy` resource tied to a specific namespace (e.g. `default`).
2. Generate a failed/passed state using a test pod so that a `PolicyReport` gets propagated.
3. Observe that navigating via `Headlamp -> Kyverno -> Policies -> Your Namespaced Policy` correctly lists `Associated Report Results (X)` rather than an incorrect string mismatch of `(0)`.

### Screenshots (Before)

image

### After

image

### DIFF
--- a/kyverno/src/components/PolicyViewer.tsx
+++ b/kyverno/src/components/PolicyViewer.tsx
@@ -159,15 +159,18 @@ function RulesTable({ rules }: { rules: PolicyRule[] }) {
   );
 }
 
-function AssociatedReportsSection({ policyName }: { policyName: string }) {
+function AssociatedReportsSection({ policyName, namespace }: { policyName: string; namespace?: string }) {
   const { items: policyReports } = PolicyReport.useList();
   const { items: clusterPolicyReports } = ClusterPolicyReport.useList();
 
   const matchingResults: (PolicyReportResult & { reportName: string; reportNamespace?: string })[] = [];
 
+  const qualifiedName = namespace ? `${namespace}/${policyName}` : policyName;
+  const matchesPolicy = (resultPolicy: string) => resultPolicy === policyName || resultPolicy === qualifiedName;
+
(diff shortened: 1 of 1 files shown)


itvi-1234:
Hi @illume  @sniok  @joaquimrocha  @yolossn 👋

This PR fixes a silent bug in the PolicyViewer where namespaced policies always showed "Associated Report Results (0)" even when real violations existed.

**Root cause:** Kyverno stores policy names in reports with a namespace prefix (e.g. `default/require-labels`), but the UI was doing a strict match against just `require-labels` — so nothing ever matched.

**Fix:** Pass the namespace down to `AssociatedReportsSection` and check both formats (`policyName` and `namespace/policyName`) before filtering results.

ClusterPolicies were unaffected since they don't get a namespace prefix.

Happy to make any changes if needed. Thanks!

itvi-1234:
Hey @joaquimrocha , Any direction regarding it

PR #630, opened 2026-04-15
fix(flux): resolve empty Helm inventory and speed up resource loading

ayushmaan-16:
### Description
I found that the `fetchResources` function was returning the data list immediately, without actually waiting for the API calls to finish. On top of that, it was firing these requests one-by-one in a loop, which was making the process quite slow.

image

### Changes
I have refactored the function to use `Promise.all` with a `.map()` logic. Now, the code correctly waits for all the data to arrive before showing it in the UI. Also, instead of making separate calls one by one, it now fetches everything at once in parallel, which makes it much more efficient.

image

### DIFF
--- a/flux/src/common/StatusLabel.tsx
+++ b/flux/src/common/StatusLabel.tsx
@@ -1,12 +1,13 @@
 import { StatusLabel as HLStatusLabel } from '@kinvolk/headlamp-plugin/lib/components/common';
-import { KubeObject } from '@kinvolk/headlamp-plugin/lib/lib/k8s/KubeObject';
+import { KubeCRD } from '@kinvolk/headlamp-plugin/lib/k8s/crd';
+import { KubeObject } from '@kinvolk/headlamp-plugin/lib/k8s/KubeObject';
 import { Tooltip, Typography } from '@mui/material';
 
 interface StatusLabelProps {
   item: KubeObject;
 }
 
-export function getStatusText(item: KubeCRD): string {
+export function getStatusText(item: KubeObject | KubeCRD): string {
   const ready = item?.jsonData?.status?.conditions?.find((c: any) => c.type === 'Ready');
   if (!ready) return '-';
   if (item?.jsonData?.spec?.suspend) return 'Suspended';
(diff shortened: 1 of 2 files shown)


illume (changes requested):
Thanks for these changes.


Can you please have a look at the git commits to see if they meet the contribution guidelines? We use a Linux kernel style of git commits. See the [contributing guide](https://headlamp.dev/docs/latest/development/contributing/#commit-guidelines) for general context, and please see previous git commits with git log for examples.

<details>
<summary>Commit guidelines</summary>

- Use atomic commits focused on a single change.
- Use the title format `<area>: <description of changes>`.
- Keep the title and body lines under 72 characters.
- Explain the intention and why the change is needed.
- Make commit titles meaningful and describe what changed.
- Do not add code that a later commit rewrites; squash or reorder commits instead.
- Do not include `Fixes #NN` in commit messages.

Good examples:

- `frontend: HomeButton: Fix so it navigates to home`
- `backend: config: Add enable-dynamic-clusters flag`

</details>

ayushmaan-16:
@illume Done with the changes.

ayushmaan-16:
@illume Just wanted to lyk about the second commit in my PR (`flux: resolve TS2304 compilation error for KubeCRD`).

While working on the `HelmInventory` fix, I noticed that the CI typecheck for the flux plugin was failing with a `TS2304: Cannot find name 'KubeCRD'` error in `StatusLabel.tsx`.

I looked into it a bit, and it seems this happened because two recent updates on the `main` branch slightly overlapped:
- One PR updated `StatusLabelProps` to use `KubeObject` instead, and removed the `KubeCRD` import.
- Another PR extracted the new `getStatusText(item: KubeCRD)` helper function.

When both of these got merged, the file was still using `KubeCRD` for the helper function, but the actual import was missing. This is what was causing the build check to fail.

Just to fix the CI, I have gone ahead and added a small fix for this. I updated the types to safely match the new `KubeObject` usage, re-added the required imports, and made sure the import paths are properly resolved.

Please let me know if any further changes are required from my side. 

Thanks!

illume:
@ayushmaan-16 thanks for the explanation and the fix!

illume:
@ayushmaan-16 it looks like it needs a npm run lint— —fix

ayushmaan-16:
@illume done with the  npm run lint— —fix. Could you please review it.

illume (approved):
🎉 thanks!

PR #631, opened 2026-04-18
kyverno : Dashboard : Fix incomplete isLoading state for policies

itvi-1234:
### Summary

This PR fixes a race condition where the `isLoading` state in the Kyverno Dashboard was dropping too early. 
Previously, `isLoading` only waited for `clusterPolicies` and `policyReports`. If these two hooks fetched instantly (e.g., returning empty arrays), the dashboard component would assume loading was completely finished and mount early, while `policies` and `clusterPolicyReports` were still pending (`null`). This caused the dashboard stats to temporarily flicker or show incomplete/zero data on initial load.


itvi-1234:
Hi @joaquimrocha  @yolossn @illume @sniok , could you please review this PR

PR #633, opened 2026-04-19
fix(chart): guard against undefined data point in multi-series merge

atharrva01:
When two Prometheus `query_range` responses return different sample counts, which happens routinely on fresh or restarted pods  the merge loop in `Chart.tsx` throws `TypeError: Cannot read properties of undefined (reading 'y')` and kills the chart panel via ErrorBoundary.

The fix is one line: optional-chain `data[index]` and fall back to `null`. Recharts already handles null dataKey values gracefully, so the shorter series just shows a gap instead of crashing.

Tested against NetworkChart (rx + tx) on a freshly scheduled pod , both series render cleanly now.

### DIFF
--- a/prometheus/src/components/Chart/Chart/Chart.tsx
+++ b/prometheus/src/components/Chart/Chart/Chart.tsx
@@ -149,7 +149,7 @@ export default function Chart(props: ChartProps) {
         (element, index) => {
           const mergedElement = { timestamp: element.timestamp };
           for (const plotName of Object.keys(fetchedMetrics)) {
-            mergedElement[plotName] = fetchedMetrics[plotName].data[index].y;
+            mergedElement[plotName] = fetchedMetrics[plotName].data[index]?.y ?? null;
           }
           return mergedElement;
         }


atharrva01:
hi @yolossn @illume guarding data[index] since secondary Prometheus series can have fewer samples than the first, was crashing the chart render.

illume (changes requested):
Thanks for these changes.


Can you please have a look at the git commits to see if they meet the contribution guidelines? We use a Linux kernel style of git commits. See the [contributing guide](https://headlamp.dev/docs/latest/development/contributing/#commit-guidelines) for general context, and please see previous git commits with git log for examples.

<details>
<summary>Commit guidelines</summary>

- Use atomic commits focused on a single change.
- Use the title format `<area>: <description of changes>`.
- Keep the title and body lines under 72 characters.
- Explain the intention and why the change is needed.
- Make commit titles meaningful and describe what changed.
- Do not add code that a later commit rewrites; squash or reorder commits instead.
- Do not include `Fixes #NN` in commit messages.

Good examples:

- `frontend: HomeButton: Fix so it navigates to home`
- `backend: config: Add enable-dynamic-clusters flag`

</details>

atharrva01:
Hey @illume, thanks! I’ve updated the commit message to Linux kernel style: `prometheus: Chart: Guard merge against out-of-bounds series access` with a short explanation. Let me know if it looks good!

illume (approved):
🎉 thanks!

PR #637, opened 2026-04-21
Kyverno : Make CEL Policies clickable with detailed Activity viewer

itvi-1234:
### Summary
This PR enhances the UX by allowing users to click on any CEL policy (Validating, Mutating, Generating, and Deleting) directly from the `CELPolicyList`. Previously, policy names were rendered as plain text strings, preventing users from seeing the internal specifications and expressions. 

Clicking the policy name now launches a `split-right` Activity overlay that beautifully renders the policy details and configurations using `SimpleTable`.

### Changes
- **Added:** `CELPolicyViewer.tsx` component to act as the detailed viewer for CEL policies. It dynamically supports `ValidatingPolicy`, `MutatingPolicy`, `GeneratingPolicy`, and `DeletingPolicy` using `instanceof` checks.
- **Updated:** `CELPolicyList.tsx` to render the policy `Name` as a `MuiLink` (button component) which intercepts clicks to trigger `Activity.launch()` opening the new detailed side-panel overlay.
- **Implemented:** Robust `SimpleTable` mapping within the viewer to parse and display real policy properties like `Expressions`, `Messages`, `Patch Types`, and `Schedules`.

### Steps to Test
1. Build the kyverno plugin using `npm run build` in the `plugins/kyverno` directory.
2. Ensure you have a CEL policy in your active cluster (e.g., `ValidatingPolicy`, `MutatingPolicy`).
3. Navigate to **Kyverno -> Validating Policies** (or corresponding pages).
4. Observe that the policy names in the table are now blue clickable links.
5. Click a policy name to open the detailed drawer on the right side and verify that the metrics/tables render accurate data.

### Screenshots
image


itvi-1234:
Hey @joaquimrocha , Could you please review it

itvi-1234:
Hi @joaquimrocha , Any directions regarding this PR :rocket:

skoeva (changes requested):
hi, could you address the open comments?

illume (changes requested):
Thanks for these changes.


Can you please address the open review comments?

itvi-1234:
Hi @illume  @skoeva Thanks for the review!  , I have Addressed all points in the latest commit:

1. ID collision - included resourceClass.kind in the Activity id
2. Stale data - viewer now receives name + resourceClass and calls useGet internally, matching the PolicyViewer pattern
3. Import path - SectionBox moved to lib/components/common
| any - removed from prop type; only cast at the useGet call site, with instanceof narrowing for type-safe field access in JSX

illume:
For the commit message, just having the kyverno part there at the start is better.

(Also for the pr title)

illume (changes requested):
Thanks for this PR.


The open review comments from Copilot still need attention — can you have a look?

itvi-1234:
Hi @illume I have Addressed the latest review comments:

-> ready without any — Added a CELPolicy union type and a structural CELPolicyResourceClass interface (with useGet), so policy.ready is now accessed type-safely without any casts

-> messageExpression fallback — Message column now renders item.message ?? item.messageExpression ?? '-' to handle both fields from CELValidation

illume (changes requested):
Thanks for this PR.


The open review comments from Copilot still need attention — can you have a look?

PR #644, opened 2026-04-25
minikube: Use readFileSync instead of cat

skoeva:
This change replaces the `execSync('cat /etc/os-release').toString()` call in `askRestartLibvirtUbuntu24()` with `fs.readFileSync('/etc/os-release', 'utf8')`.

Specifically in `minikube/manage-minikube.js`:
- Add `readFileSync` from Node’s built-in `fs` module near the top-level requires.
- Update line 67 region to use `readFileSync('/etc/os-release', 'utf8')`.
- Keep surrounding logic unchanged so behavior remains the same (string content check for Ubuntu 24).


illume (approved):
🎉

PR #662, opened 2026-05-05
Fix local model tool calls

ashu8912:


skoeva (changes requested):
looks like there's a formatting issue, run `npm run format -- --check`

PR #665, opened 2026-05-05
app-catalog: Fix early returns before hooks in EditorDialog components

### LINKED ISSUE #525: app-catalog: EditorDialog: Early returns should be moved after hooks

AshutoshSharma-pixel:
Fixes #525

## Problem

Both `EditorDialog` components had early returns before hook calls,
violating React's Rules of Hooks which require hooks to be called
unconditionally on every render.

- `releases/EditorDialog.tsx`: `if (!release) return null` at line 52,
  before 10+ hook calls
- `charts/EditorDialog.tsx`: `if (!props.chart) return null` at line 27,
  before all hook calls

## Fix

Moved both guards to after all hooks. State initializers in
`releases/EditorDialog.tsx` made null-safe with optional chaining
(`release?.chart?.values`, `release?.config`).

## Testing

`npm run tsc`, `npm run lint`, and `npm run format` all pass clean.

### DIFF
--- a/app-catalog/src/components/releases/EditorDialog.tsx
+++ b/app-catalog/src/components/releases/EditorDialog.tsx
@@ -49,15 +49,12 @@ export function EditorDialog(props: {
     isUpdateRelease,
     handleUpdate,
   } = props;
-  if (!release) return null;
   const { t } = useTranslation();
   const themeName = localStorage.getItem('headlampThemePreference');
   const { enqueueSnackbar } = useSnackbar();
-  const [valuesToShow, setValuesToShow] = useState(
-    Object.assign({}, release.chart.values, release.config)
-  );
-  const [values, setValues] = useState(Object.assign({}, release.chart.values, release.config));
-  const [userValues, setUserValues] = useState(release.config);
+  const [valuesToShow, setValuesToShow] = useState<Record<string, unknown>>({});
+  const [values, setValues] = useState<Record<string, unknown>>({});
+  const [userValues, setUserValues] = useState<Record<string, unknown>>({});
   const [isUserValues, setIsUserValues] = useState(false);
   const [releaseUpdateDescription, setReleaseUpdateDescription] = useState('');
   const [upgradeLoading, setUpgradeLoading] = useState(false);
@@ -72,6 +69,7 @@ export function EditorDialog(props: {
   const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
+    if (!release) return;
     let isMounted = true;
 
     if (isUpdateRelease) {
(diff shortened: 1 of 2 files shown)


app-catalog/src/components/releases/EditorDialog.tsx
```diff
@@ -49,15 +49,14 @@ export function EditorDialog(props: {
     isUpdateRelease,
     handleUpdate,
   } = props;
-  if (!release) return null;
   const { t } = useTranslation();
   const themeName = localStorage.getItem('headlampThemePreference');
   const { enqueueSnackbar } = useSnackbar();
   const [valuesToShow, setValuesToShow] = useState(
-    Object.assign({}, release.chart.values, release.config)
+    Object.assign({}, release?.chart?.values, release?.config)
```
skoeva: this seems unrelated to the issue? maybe separate this into a different commit

AshutoshSharma-pixel: This change is necessary as part of the hooks fix - after moving the if (!release) return null guard to after the hooks, release can be null when useState initializers run. Without optional chaining here, it would throw at render time. Happy to split into two commits if preferred though.

skoeva: ah okay, feel free to keep it in the one commit!

skoeva (approved):
lgtm!

AshutoshSharma-pixel:
@illume Addressed all 3 Copilot comments in the latest commit:

- releases/EditorDialog.tsx: state initializers now use empty object defaults {} instead of dereferencing release directly; added useEffect to sync valuesToShow, values, and userValues when release loads; added if (!release) return; inside the useEffect that dereferences release.chart.metadata

- charts/EditorDialog.tsx: moved if (!props.chart) return null to after ALL useEffect hooks (line 130, after the last useEffect at line 123)

AshutoshSharma-pixel:
@ashu8912 All Copilot comments have been addressed in the latest commit (214296f). The guards are now after all hooks in both files. @skoeva has approved ready for merge when you get a chance.

AshutoshSharma-pixel:
@skoeva addressed the remaining Copilot comment in the latest commit (e39f537) added release to the useEffect dependency array so chart versions re-fetch correctly when release loads asynchronously.

AshutoshSharma-pixel:
Addressed all 4 Copilot comments in the latest commit (f3767b7):

- valuesToShow and values now store independent copies using spread ({...merged}) to avoid shared reference mutation
- userValues also stores an independent copy ({...release.config})
- All three useState initializers typed as Record<string, unknown>
- chart added to selectedVersion useEffect dependency array
- Guard changed from props.chart to chart for consistency

AshutoshSharma-pixel:
Thanks for the feedback. The current implementation is sufficient for the scope of this fix (moving guards after hooks). The lazy initializer suggestion adds complexity beyond what's needed here, and Object.assign({}, ...) already creates a new object so the spread is redundant but harmless. I'll leave these as potential follow-up improvements. The core Rules of Hooks violation is fixed.

PR #666, opened 2026-05-06
strimzi: Migrate to kafka.strimzi.io/v1, drop ZooKeeper support

krrish-sehgal:
## Summary

Strimzi 1.0.0 (April 2026) promotes all CRDs to the `v1` API and
removes `v1beta2` entirely. ZooKeeper-based Kafka clusters are no
longer supported; KRaft is the only control plane going forward.

This PR updates the plugin to target the new API.

**What changed:**

- All five resource classes (`Kafka`, `KafkaTopic`, `KafkaUser`,
  `KafkaNodePool`, `StrimziPodSet`) bumped from `v1beta2` to `v1`
- Every hardcoded API path in `KafkaTopicList`, `KafkaUserList`, and
  `KafkaClusterTopology` updated to `/apis/kafka.strimzi.io/v1/...`
  and `/apis/core.strimzi.io/v1/...`
- Removed `spec.zookeeper` field and getter from the `Kafka` resource
  class; removed `isKRaftMode`, `getClusterMode`, and the `KafkaLike`
  helper type (ZooKeeper detection is no longer needed)
- Dropped the `Mode` column from `KafkaList` and the `ZooKeeper replicas`
  row from the Kafka detail view
- Simplified `KafkaClusterTopology`: removed the entire ZooKeeper branch
  from layout calculation and node generation; only two paths remain
  (KafkaNodePools and a minimal fallback for clusters where node pools
  have not yet reconciled)
- Storybook mocks updated to `v1` and renamed the ZK cluster fixture to
  `dev-cluster`
- Deleted ZK-only test files: `zookeeper-ephemeral.yaml`,
  `topic-basic-zk.yaml`, `user-scram-zk.yaml`
- Remaining test YAMLs bumped to `v1`; KRaft examples updated to
  Kafka 4.0.0 with `KafkaNodePool`

**README prerequisite:** Strimzi 1.0.0 or later is now required.

## Why

The v1beta2 API is gone in Strimzi 1.0.0. Any cluster running 1.0.0+
with this plugin at its old version would get `404` on every API call
and see no resources in the UI.

## Steps to test

1. Run a cluster with Strimzi 1.0.0 installed
2. Apply `test-files/kafka/kraft-ephemeral.yaml` (creates a KRaft
   cluster with a KafkaNodePool)
3. Open Headlamp with the plugin loaded; confirm Kafka cluster, topics,
   and users all list and detail correctly
4. Confirm the Topology modal renders without a ZooKeeper group

## Notes

[...]

PR #668, opened 2026-05-06
strimzi: api: Add runtime v1/v1beta2 API version detection

krrish-sehgal:
## Summary

Strimzi 1.0.0 graduated all CRDs to `kafka.strimzi.io/v1` and
`core.strimzi.io/v1`, retiring the `v1beta2` API. Rather than a hard
cutover that would break clusters still running Strimzi 0.x, this PR
makes the plugin work with **both** generations by detecting which
version the connected cluster serves.

Key changes:

- **`src/utils/strimziApiVersion.ts`** — probes the API group discovery
  endpoints (`/apis/kafka.strimzi.io` and `/apis/core.strimzi.io`) once
  on plugin load and caches the result. A single shared promise prevents
  duplicate requests when multiple components mount simultaneously.
- **`src/hooks/useStrimziApiVersions.ts`** — React hook that returns the
  cached `{ kafka, core }` version strings and triggers a re-render when
  the probe resolves.
- **V1 subclasses** (`KafkaV1`, `KafkaTopicV1`, `KafkaUserV1`,
  `KafkaConnectV1`, `KafkaConnectorV1`) extend the existing v1beta2
  classes with a single `static apiVersion` override. No logic is
  duplicated.
- **`KafkaConnectorV1`** also overrides `desiredState()` to drop the
  deprecated `spec.pause` fallback that was removed from the v1 API.
- All six list / topology components call `useStrimziApiVersions()`,
  select the right class for `ResourceListView`, and build API paths
  from the resolved version string.

## Why this is needed

- Clusters running Strimzi 0.x only serve `v1beta2`; pointing them at
  `v1` returns 404s.
- Clusters running Strimzi 1.0+ only serve `v1`; pointing them at
  `v1beta2` returns 404s.
- A single release should cover both, with no user configuration
  required.

## Steps to test

1. Apply Strimzi 0.x on a cluster (serves v1beta2). Install the plugin.
   Verify all list views load and CRUD operations work.
2. Apply Strimzi 1.0+ on a cluster (serves v1 only). Install the plugin.
   Verify all list views load and CRUD operations work.
3. In both cases, check the browser console. There should be no 404
   errors on `/apis/kafka.strimzi.io/...` or `/apis/core.strimzi.io/...`.

## Notes

- The probe defaults to `v1beta2` before it resolves (safe fallback).
  On a Strimzi 1.0 cluster there is a brief initial flash before the
  correct v1 class is used; this resolves in one re-render cycle.
- PRs #666 and #667 (hard v1-only migration) can be superseded by this
  approach and closed.

### DIFF
--- a/strimzi/README.md
+++ b/strimzi/README.md
@@ -41,7 +41,7 @@ A Headlamp plugin for managing Strimzi (Apache Kafka on Kubernetes) resources di
 ## 📋 Prerequisites
 
 - [Headlamp](https://headlamp.dev/) installed
-- A Kubernetes cluster with [Strimzi operator](https://strimzi.io/) deployed
+- A Kubernetes cluster with [Strimzi operator](https://strimzi.io/) deployed (0.x with `v1beta2` APIs, or 1.0+ with `v1` APIs; the plugin detects and uses whichever version the cluster serves)
 
 ## 🚀 Quick Start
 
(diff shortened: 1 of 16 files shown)


krrish-sehgal:
Superseded by #667 which now contains the same dual-API support. Closing this to keep things clean.

PR #669, opened 2026-05-06
strimzi: Add Error Boundary to prevent plugin crashes breaking Headlamp

krrish-sehgal:
## Summary

When the Strimzi plugin is installed on a cluster that does not have
the Strimzi operator deployed, any navigation to a Strimzi view causes
API calls that return 404. In some cases this leads to a React render
error that propagates all the way up to Headlamp's root React tree,
blanking the entire UI and making cluster connections appear broken on
the home page. This was reported by Jakub.

**Root cause:** The plugin had no React Error Boundary. An uncaught
error inside any plugin component escapes the plugin and crashes
Headlamp's own rendering.

**Fix:** Two changes, zero new dependencies:

1. `StrimziErrorBoundary` — a React class component (Error Boundaries
   must be class components per the React spec). Catches any render
   error in its subtree and shows a scoped error message inside the
   Strimzi view, leaving the rest of Headlamp completely unaffected.

2. `safe()` helper in `index.tsx` — wraps every registered route
   component with the boundary. The wrapper is a named function
   component so React DevTools shows a useful display name.

## Steps to test

1. Install the plugin in Headlamp.
2. Connect to a cluster that does **not** have Strimzi installed.
3. Navigate to **Strimzi > Kafka Clusters** in the sidebar.
4. **Before this fix:** Headlamp home page shows clusters as
   disconnected / the entire UI may go blank.
5. **After this fix:** Only the Strimzi view shows an error message;
   the rest of Headlamp continues to work normally.

### DIFF
--- a/strimzi/src/components/StrimziErrorBoundary.tsx
+++ b/strimzi/src/components/StrimziErrorBoundary.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { Box, Typography } from '@mui/material';
+import { SectionBox } from '@kinvolk/headlamp-plugin/lib/components/common';
+
+interface State {
+  hasError: boolean;
+  message: string;
+}
+
+/**
+ * React Error Boundary for the Strimzi plugin.
+ *
+ * Wrapping every plugin route in this boundary ensures that uncaught
+ * render errors inside any Strimzi component are contained here and
+ * never propagate up to Headlamp's root React tree. Without this, a
+ * render crash inside the plugin can blank the entire Headlamp UI,
+ * including the home-page cluster connection indicators.
+ */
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 thanks!

krrish-sehgal:
Closing as obsolete: the Error Boundary functionality this PR adds already landed in main via #667 (commit 5598407 "strimzi: Handle missing Strimzi CRDs gracefully"), which introduced StrimziErrorBoundary and the safe() route wrapper in strimzi/src/index.tsx. The remaining diff is only cosmetic (comment wording, line wrapping, import order), so there's nothing left to merge here.

PR #672, opened 2026-05-07
knative: Align list columns with kubectl output

### LINKED ISSUE #479: knative: kubectl column consistency
This discussion from the PR https://github.com/headlamp-k8s/plugins/pull/473#discussion_r2668514618

> We usually try to keep the list table columns similar to the `kubectl get` response. 
> image

>> Thanks for your review 🙂
>> If we were only outputting the raw results of kubectl get, it would just be a matter of developing an alias for the CRD menu built into the headlamp. However, I don't believe any user would install a plugin just for that. Therefore, I think we should provide added value. What do you think?
Or does this mean you want to display additional columns that are shown in kubectl get but aren't displayed here?

Without speaking for him, I think he means we try to be consistent with the columns that kubectl uses when possible. Because there's usually been a lot of thinking and years of tweaks to which columns are displayed. As well, there's a big group of people who are used to that. But of course where are very good reasons we have diverged from the columns kubectl chooses.

AshutoshSharma-pixel:
Fixes #479

## Problem

The Revisions and DomainMappings list views were not consistent with
`kubectl get` column output.

`kubectl get revisions`:
- Shows IMAGE before READY, and has a separate REASON column
- Plugin had READY before IMAGE, and no REASON column

`kubectl get domainmapping`:
- Has a separate REASON column after READY
- Plugin had no REASON column

## Changes

**revisions/List.tsx:**
- Moved `image` column before `ready` to match kubectl order
- Added `reason` column after `ready`

**domainmappings/List.tsx:**
- Added `reason` column after `ready`

## Testing

`npm run tsc`, `npm run lint`, and `npm run format` all pass clean.

### DIFF
--- a/knative/src/components/domainmappings/List.tsx
+++ b/knative/src/components/domainmappings/List.tsx
@@ -82,6 +82,28 @@ export function useDomainMappingColumns(
         },
       },
       {
+        id: 'reason',
+        label: 'Reason',
+        gridTemplate: 'auto',
+        disableFiltering: true,
+        getValue: item => {
+          const dm = item as KnativeDomainMapping;
+          return getReadyCondition(dm).reason?.toLowerCase() || '';
+        },
+        render: item => {
+          const dm = item as KnativeDomainMapping;
+          const reason = getReadyCondition(dm).reason;
+          if (!reason) {
+            return (
+              <Typography variant="body2" color="text.secondary">
+                -
+              </Typography>
+            );
+          }
+          return <Typography variant="body2">{reason}</Typography>;
+        },
+      },
+      {
         id: 'clusterdomainclaim',
         label: 'ClusterDomainClaim',
         gridTemplate: 'auto',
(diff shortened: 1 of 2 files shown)


yolossn:
@mudit06mah The changes look good to me. Would like to get your inputs since you are actively working on it. Thoughts?

mudit06mah:
Hey @yolossn !
This issue was actually resolved in #473 (or a subsequent PR by @kahirokunn) so it should have been closed already.
The reason the plugin does not have a reason column is because the reason is shown as a tooltip when hovered over the Ready column:
image

The only valid change in the PR is placing the image column before the Ready label, which too might not be that necessary?

yolossn:
@AshutoshSharma-pixel Looks like this is already fixed. Closing this.

kahirokunn:
Thanks for your contribution 🙏

AshutoshSharma-pixel:
Thanks for the context @mudit06mah and @yolossn .I missed the tooltip implementation. Appreciate the review.

mudit06mah:
> Thanks for the context @mudit06mah and @yolossn .I missed the tooltip implementation. Appreciate the review.

Sorry for still having the issue open and thank you for your contribution :)

PR #683, opened 2026-05-09
fix(flux): resolve hardcoded URLs and maintain cluster context

Joshna907:
### **Summary**
This PR refactors the Flux plugin's navigation logic to eliminate hardcoded URL strings and direct window.location mutations. It migrates all internal links to use Headlamp's Routing API, ensuring compatibility with multi-cluster and sub-path deployments.

### **Key changes**
`flux/src/overview/index.tsx`:

1. Migrated the "Flux Settings" action to use Utils.Router.createRouteURL instead of a hardcoded path.
2. Updated the prepareLink helper to return internal route names (e.g., 'sources', 'kustomizations') which allows Headlamp to dynamically generate the correct URLs.
3. Ensured FluxOverviewChart links pass the activeCluster context to prevent accidental cluster switching.

`flux/src/helpers/index.tsx`:

1. Updated the NameLink helper to propagate row.original.cluster to the Link component, ensuring resource links maintain the correct cluster context in multi-cluster environments.
2. Why this is needed
3. Broken Navigation: Hardcoded paths starting with / fail when Headlamp is served behind a reverse proxy or has a base path (e.g., /headlamp/).
4. Cluster Context Loss: Without the activeCluster property, navigating through the Flux plugin can cause the UI to drop the user back into the default cluster, creating confusion and potential errors when managing multiple environments.

### **Steps to test**

1. In a multi-cluster Headlamp environment, navigate to the Flux -> Overview page of a non-default cluster.
2. Click on any resource category legend (e.g., "Kustomizations").
3. Verify that you are taken to the correct list view on the same cluster (check the URL prefix).
4. Click the "Settings" (cog) icon in the Flux Overview header and verify it correctly opens the plugin settings page without a 404 error.

### DIFF
--- a/flux/src/helpers/index.tsx
+++ b/flux/src/helpers/index.tsx
@@ -195,6 +195,7 @@ export function NameLink(resourceClass: KubeObjectClass) {
           namespace: row.original.jsonData.metadata.namespace,
           pluralName: pluralName,
         }}
+        activeCluster={row.original.cluster}
       >
         {cell.getValue()}
       </Link>
--- a/flux/src/index.tsx
+++ b/flux/src/index.tsx
@@ -244,6 +244,7 @@ registerRoute({
   sidebar: 'notifications',
   component: () => <Notifications />,
   exact: true,
+  name: 'notifications',
 });
 
 registerRoute({
--- a/flux/src/overview/index.tsx
+++ b/flux/src/overview/index.tsx
@@ -1,5 +1,5 @@
 import { Icon } from '@iconify/react';
-import { K8s } from '@kinvolk/headlamp-plugin/lib';
+import { K8s, Router, Utils } from '@kinvolk/headlamp-plugin/lib';
 import {
   ActionButton,
   Link,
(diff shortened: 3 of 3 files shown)


Joshna907:
Can you check this @joaquimrocha ? and @yolossn , @illume ?

joaquimrocha:
Hey @Joshna907 . Can you change the prefix in the commit message from fix(flux) to "flux: ". We don't normally use fix/feat, even if some commits added it in the past.

About the code, let's wait for Copilot reviews. I am a bit surprised that you turn `/flux/notifications` to `notifications` and it still works like before.

Joshna907:
> Hey @Joshna907 . Can you change the prefix in the commit message from fix(flux) to "flux: ". We don't normally use fix/feat, even if some commits added it in the past.
> 
> About the code, let's wait for Copilot reviews. I am a bit surprised that you turn `/flux/notifications` to `notifications` and it still works like before.

Done! I’ve updated the commit prefix to flux:  and refactored the navigation to use Headlamp's Router and useHistory

Joshna907:
@joaquimrocha you could check it now everything should work as expected

Joshna907:
@joaquimrocha no new copilot comment generated I think we could merge it now

PR #708, opened 2026-05-13
radius: standardize empty-state health reporting and preserve cluster context

Joshna907:
### **Summary:**
This PR standardizes the Radius plugin's Overview dashboard by ensuring that resource types with no configured items report 100% health instead of an empty state. It also implements a "Not Installed" empty state and ensures that multi-cluster context is preserved during navigation, aligning the plugin with Headlamp's core visual and architectural standards.

### **Key changes**

`radius/src/radius-overview/index.tsx`:

1. Updated ResourceStatusChart to render a healthy (100%) chart when no resources exist, instead of returning a plain text "0" indicator.
2. Implemented a professional "Radius is not installed" empty state banner with installation links, triggered when Radius API calls fail.
3. Refactored all table Link components to propagate activeCluster context using Utils.getCluster(), preventing accidental cluster switching during navigation.

### **Why this is needed**

1. Misleading Metrics: Currently, the Radius dashboard shows an "empty" state for zero resources, which doesn't clearly signify a healthy environment. Aligning this with the "100% Success" pattern used in Flux and Kyverno provides a more positive and accurate UX.
2. User Consistency: A standardized "Not Installed" screen helps new users understand why the dashboard is empty and provides clear next steps for installation.
3. Cluster Context Loss: Without passing activeCluster to resource links, navigating through the Radius plugin in a multi-cluster setup can cause the UI to drop the user back into the default cluster.

### **Steps to test**

1. Connect to a cluster where Radius is not installed. Navigate to the Radius -> Overview page and verify the "Radius is not installed" banner appears.
2. Connect to a cluster where Radius is installed but no Applications or Environments exist. Verify that the status charts show 100% health instead of just the number "0".
3. In a multi-cluster environment, navigate to the Radius Overview of a non-default cluster and click on an Application or Environment name. Verify that you remain on the same cluster (check the URL prefix).

### DIFF
--- a/radius/src/models/radius.ts
+++ b/radius/src/models/radius.ts
@@ -504,6 +504,9 @@ export function useRadiusResources(): [UCPResource[] | null, Error | null, boole
         // Build promises for all resource types across all providers
         const promises: Promise<UCPResource[]>[] = [];
 
+        let successCount = 0;
+        let lastError: Error | null = null;
+
         // Helper function to fetch resources for a provider
         const fetchProviderResources = async (
           provider: string,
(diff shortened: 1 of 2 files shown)


joaquimrocha:
@filipevrevez , can you check this one?

Joshna907:
ptal @illume @skoeva @joaquimrocha

PR #726, opened 2026-05-18
prometheus: Add Kafka cluster metrics charts

krrish-sehgal:
Adds a Prometheus metrics section for Strimzi `Kafka` clusters, shown on the Kafka detail page.

This started out as part of #709 (Strimzi map integration). Per review feedback it has been split out and reworked to follow the prometheus plugin's chart pattern instead of the standalone metrics client it originally used. #709 is now map-only.

## Approach

Same pattern as the existing Karpenter and KEDA charts:

- `Kafka` is added to `ChartEnabledKinds` in `util.ts`.
- `PrometheusMetrics()` in `index.tsx` returns a `StrimziChart` for the Kafka kind.
- The existing `registerDetailsViewSectionsProcessor` injects the section into the Kafka detail page, so the Strimzi plugin itself needs no changes.

Because it reuses the plugin's cluster config and endpoint discovery, both manual-address and autoDetect Prometheus setups work, and metrics appear automatically for anyone who already has the Prometheus plugin configured.

## What it shows

`StrimziChart` offers four metric views via a toggle, all scoped to the Kafka cluster's namespace:

- Throughput: bytes in and out per sec
- Messages: messages in per sec
- Partition health: under-replicated and offline partitions
- Consumer lag: total consumer group lag

Queries target the kafka-exporter and JMX exporter scrape outputs that ship with the standard Strimzi monitoring stack.

## Notes

- Queries are namespace-scoped. A Strimzi deployment normally runs one Kafka cluster per namespace, and namespace scoping keeps the queries working regardless of whether the `strimzi_io_cluster` label survives Prometheus relabeling.
- One new translation key (`metric chooser`) was extracted via `npm run i18n`; the rest of the strings already existed.

## Test plan

- [ ] `npm run tsc`, `npm run lint`, `npm test`, and `npm run build` all pass
- [ ] On a Kafka detail page with the Prometheus plugin configured and Strimzi monitoring deployed, confirm the four metric views populate and refresh
- [ ] With Prometheus not configured for the cluster, confirm the setup banner shows instead of an error

### DIFF
--- a/prometheus/locales/cs/translation.json
+++ b/prometheus/locales/cs/translation.json
@@ -57,6 +57,7 @@
   "memory": "",
   "receive": "",
   "transmit": "",
+  "metric chooser": "",
   "Invalid Address Format": "",
   "Testing Connection": "",
   "Connection successful!": "",
--- a/prometheus/locales/de/translation.json
+++ b/prometheus/locales/de/translation.json
@@ -57,6 +57,7 @@
   "memory": "",
   "receive": "",
   "transmit": "",
+  "metric chooser": "",
   "Invalid Address Format": "",
   "Testing Connection": "",
   "Connection successful!": "",
--- a/prometheus/locales/en/translation.json
+++ b/prometheus/locales/en/translation.json
@@ -57,6 +57,7 @@
   "memory": "memory",
   "receive": "receive",
   "transmit": "transmit",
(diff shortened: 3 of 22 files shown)


illume (approved):
🎉 thanks!

PR #744, opened 2026-05-25
minikube: Update dependencies

yolossn:


minikube/src/index.tsx
```diff
@@ -13,7 +13,7 @@ import React from 'react';
 import ClusterStatus from './ClusterStatus';
 import CommandCluster from './CommandCluster/CommandCluster';
 import CreateClusterPage from './CreateClusterPage';
-import MinikubeIcon from './minikube.svg';
+import MinikubeIcon from './minikube.svg?react';
```
skoeva, line 16: this should be squashed into the headlamp-plugin update commit

ashu8912: This file doesn't seem to be part of the diff, is this removed now?

ashu8912 (approved):
LGTM!!

PR #746, opened 2026-05-25
plugin-catalog: Update dependencies

yolossn:


plugin-catalog/src/components/plugins/PluginCard.tsx
```diff
@@ -12,7 +12,7 @@ import {
   Typography,
 } from '@mui/material';
 import { PluginPackage } from './List';
-import PluginIcon from './plugin-icon.svg';
+import PluginIcon from './plugin-icon.svg?react';
```
skoeva, line 15: this should be squashed into the headlamp-plugin update commit

ashu8912: Is there a fix for this?? We saw this breaks the plugin catalog

yolossn: @ashu8912 the commit that adds this was removed

Issue #486, opened 2026-01-20
Improve Knative Headlamp plugin: Serverless Workload Management UI - Linux Foundation paid mentorship project


illume:
This issue is for a Linux Foundation mentorship project. See https://mentorship.lfx.linuxfoundation.org/project/b6178852-f18d-41e2-b352-c839b8570eae This project is ongoing by @mudit06mah

  **Knative** is a Kubernetes-based platform for deploying and managing serverless workloads, providing automatic scaling (to zero), traffic splitting, and simplified app deployment. This project will continue development of a Headlamp plugin that brings **first-class UI support for Knative** (focusing on Knative Serving resources) directly into Headlamp’s dashboard. The plugin will allow Kubernetes operators and developers to **visualize, inspect, and tweak Knative Services and their components** (Services, Revisions, Configurations, Routes/traffic rules) without leaving the Headlamp UI. It builds on an initial [Knative plugin](https://github.com/headlamp-k8s/plugins/473). Our goal is to complete the remaining work and polish the experience. By doing so, we empower users to manage serverless applications (like deploying new revisions or adjusting traffic splits) more easily, complementing the `kn` CLI with an intuitive dashboard interface.

## Expected Outcome

- **Complete and fully functional Knative plugin** integrated into the official Headlamp plugin repository, appearing as a new sidebar item “Knative”. The plugin should **list Knative Services (KService)** across namespaces, showing key info (name, URL, traffic percentage, latest revision status, etc.) in a table. The list columns will be adjusted to closely match what `kubectl get kservice` displays for familiarity.
- **Detail pages for Knative Services** that present both **status and configuration**: e.g., the service’s URL, current traffic split between revisions, concurrency and scaling settings, and condition statuses (Ready/Not Ready). Users should be able to perform common actions from the UI, such as adjusting traffic percents between revisions, updating configuration values (like environment variables or concurrency limits), and triggering a new revision (rolling out new image or configuration) via an “Update” or “Redeploy” action. These actions would be backed by form inputs or modals, and changes can be saved to the cluster with proper feedback.
[...]

PR #752, opened 2026-05-26
knative: kserivce: Fix validation for revision name

mudit06mah:
## Description

This PR fixes validation error by adding name to the metadata in resource definition, and allowing it to be null.
Error was earlier being raised as the newest revision was being edited instead of a new revision being made, which threw a validation error.

## Related Issue
#486 (LFX)


illume (approved):
🎉

PR #765, opened 2026-05-28
kyverno: Restore dropped fixes for ViolationsView scope and Policy ready status

itvi-1234:
### Description
This PR restores a couple of fixes that were accidentally dropped during recent force-pushes/rebases on the `add-kyverno-plugin` branch.

**Restored Fixes:**
1. **ViolationsView Missing Pod Names:** 
   When Kyverno generates per-resource policy reports, the resource information is stored in the report's `scope` rather than the individual result's `resources` array. This PR restores the fallback to the report scope so that pod names, kinds, and namespaces are displayed correctly in the Violations table instead of `-`. (Originally fixed in #677)

2. **Policy Ready Status Logic:**
   During the refactoring into `KyvernoPolicyBase`, the logic to compute the `ready` status from policy conditions was overwritten with the older `this.status?.ready` implementation. This restores the logic to correctly check for the `Ready` condition in the `status.conditions` array. (Originally fixed in #678)

### DIFF
--- a/kyverno/src/components/ViolationsView.tsx
+++ b/kyverno/src/components/ViolationsView.tsx
@@ -32,6 +32,7 @@ import { useMemo, useState } from 'react';
 import {
   ClusterPolicyReport,
   PolicyReport,
+  PolicyReportInterface,
   PolicyReportResult,
   PolicyResultStatus,
 } from '../resources/policyReport';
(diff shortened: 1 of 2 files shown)


itvi-1234:
cc @joaquimrocha

joaquimrocha (approved):

Issue #755, opened 2026-05-27
karpenter: do i18n


illume:
See https://headlamp.dev/docs/latest/development/plugins/i18n

PR #769, opened 2026-05-28
karpenter: implement i18n

ayush-that:
Implements i18n for the karpenter plugin per #755.

- wires `useTranslation` from `@kinvolk/headlamp-plugin/lib` into every component that renders user-facing strings (NodePool list/detail, NodeClass list/detail, Scaling list/detail, PendingPods list, EventList, StatusLabel, EditConfigButton, resourceEditor dialog)
- the CPU/Memory tooltip helpers take `t` as a parameter since they are invoked as plain functions from PercentageBar's tooltipFunc, not as JSX components
- adds `"headlamp": { "i18n": ["en"] }` config in package.json (the `npm run i18n` script already existed)
- adds the generated `locales/en/translation.json` (~108 keys)

Module-level strings (registerSidebarEntry labels in index.tsx, map source labels in mapView.tsx, cloudProviders.ts column configs) are left as raw English — they run before any React context exists, matching the precedent in flux/prometheus/kubeflow plugins.

Verified locally: `npm run tsc`, `npm run lint`, `npm test`, and `npm run build` all pass. Build output confirms the headlamp-plugin useTranslation transform runs and `locales/` is copied to `dist/`.

Refs #755


illume (approved):
🎉 thanks!

PR #783, opened 2026-06-02
plugin-catalog: Allow version selection in Detail view

### LINKED ISSUE #782: Plugin Catalog: no way to select which version to install
### Summary

The Plugin Catalog plugin does not let users choose **which version** of a plugin to install. ArtifactHub exposes multiple published versions for many plugins, but the catalog only ever installs the latest one, with no UI to pick a different version.

### Current behavior

- The plugin detail view (`src/components/plugins/Detail.tsx`) shows a single read-only **Available Version** field (the latest).
- The install flow builds a versionless ArtifactHub URL — `https://artifacthub.io/packages/headlamp/{repo}/{plugin}` — so the desktop backend always resolves to the latest version.
- The ArtifactHub API response already includes an `available_versions` array, but it is never surfaced in the UI.

### Expected behavior

Users should be able to see all versions ArtifactHub has published for a plugin and select which one to install (e.g. to pin to a known-good version or roll back).

### Example

The Flux plugin (`headlamp-plugins/headlamp_flux`) has 7 published versions on ArtifactHub:

`0.6.0` (latest), `0.5.0`, `0.4.0`, `0.3.0`, `0.2.0`, `0.1.0-beta-2`, `0.1.0-beta-1`

None of these (other than the latest) are selectable from the catalog today.

### Proposed solution

- Add a version selector (dropdown) on the plugin detail view, populated from `available_versions`, defaulting to the latest.
- Pass the selected version through to install by appending it as a path segment: `https://artifacthub.io/packages/headlamp/{repo}/{plugin}/{version}` (ArtifactHub serves per-version metadata at this path; the desktop backend resolves it without changes).
- Update the rendered README to reflect the selected version (fetched from `/api/v1/packages/headlamp/{repo}/{plugin}/{version}`).

### Notes

- Plugins with a single published version (e.g. `headlamp_cluster-api`, which only has `0.1.0-alpha`) will simply show that one version.

yolossn:
this patch allows the user to select a version when multiple version of the plugin is available and install that specific version

fixes: https://github.com/headlamp-k8s/plugins/issues/782

https://github.com/user-attachments/assets/eed83225-75e7-42f7-a750-26d2c4df06e4

### DIFF
--- a/plugin-catalog/src/components/plugins/Detail.stories.tsx
+++ b/plugin-catalog/src/components/plugins/Detail.stories.tsx
@@ -80,10 +80,19 @@ Default.args = {
   currentActionState: null,
   currentActionMessage: null,
   currentActionProgress: 0,
-  onInstall: () => alert('Install action triggered'),
+  selectedVersion: '1.0.0',
+  onVersionChange: version => alert(`Version changed to ${version}`),
+  onInstall: version => alert(`Install action triggered for version ${version}`),
   onUpdate: pluginName => alert(`Update action triggered for ${pluginName}`),
   onUninstall: pluginName => alert(`Uninstall action triggered for ${pluginName}`),
   onCancel: () => alert('Cancel action triggered'),
+};
+
+export const Installable = Template.bind({});
+Installable.args = {
+  ...Default.args,
+  pluginDetail: { ...samplePluginDetail, isInstalled: false, version: '1.1.0' },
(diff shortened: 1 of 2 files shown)


ashu8912 (commented):
Overall LGTM!!, just when we select a version and press install the version selector should show the selected version

ashu8912 (approved):
LGTM!!

PR #785, opened 2026-06-04
flux: Fix HelmRelease inventory missing since v0.5

### LINKED ISSUE #448: Flux: No more Inventory for HelmReleases in v0.5
In v0.4 the Flux plugin would list resources associated with the HelmRelease in Inventory. With v0.5, we are no longer seeing them listed. That was an awesome feature and makes it so that we don't want to upgrade from v0.4.

AshutoshSharma-pixel:
Closes #448

HelmRelease inventory stopped showing in v0.5. The custom HelmInventory component in Inventory.tsx was no longer being used, leaving the inventory section empty.
Fix: Updated HelmReleaseSingle.tsx to use GetResourcesFromInventory from kustomizations/Inventory.tsx, mirroring how Kustomization inventory is displayed. Removed the now-unused helm-releases/Inventory.tsx.
All checks passing: tsc, lint, format.


flux/src/helm-releases/Inventory.tsx
ashu8912, line 1: @AshutoshSharma-pixel why was this file removed?? What was the issue with previous approach?

ashu8912 (commented):
Added a comment

flux/src/helm-releases/Inventory.tsx
AshutoshSharma-pixel: The Inventory.tsx file in helm-releases/ contained a custom HelmInventory component that was no longer being used anywhere HelmReleaseSingle.tsx stopped using it between v0.4 and v0.5 causing the inventory to vanish.
Rather than revert to the previous approach, I used GetResourcesFromInventory from kustomizations/Inventory.tsx, which already handles the same pattern correctly. The old file was dead code at that point, so I deleted it.
if you want i can restore it and adapt it instead.

PR #786, opened 2026-06-05
app-catalog: Show error feedback when helm install fails

### LINKED ISSUE #225: app-catalog: When failing to install there is no indication given to user
1. Install something that fails.
2. Notice how it doesn't show you there is a problem, or tell you what the problem is.

(You can eventually see it failed to install by clicking the "installed" page).

Also there is no indication of how long something is taking to install. I think taking logs from helm and displaying them might be the way to go.

AshutoshSharma-pixel:
Closes #225

When a helm chart installation failed, the UI gave no feedback - the dialog would silently stay in a loading state or close without any error message.

Root causes fixed:
- checkInstallStatus had no .catch() on the getActionStatus polling call - network errors were silently swallowed.

- Both .catch() blocks in installAndCreateReleaseHandler called handleEditor(false) but never called setInstallLoading(false) leaving the install button stuck in "Installing..." indefinitely after failure.

- Dialog was being closed prematurely before the status polling completed.

- Error snackbars now include the specific error message and autoHideDuration: 5000

### DIFF
--- a/app-catalog/src/components/charts/EditorDialog.tsx
+++ b/app-catalog/src/components/charts/EditorDialog.tsx
@@ -129,26 +129,49 @@ export function EditorDialog(props: {
 
   function checkInstallStatus(releaseName: string) {
     setTimeout(() => {
-      getActionStatus(releaseName, 'install').then((response: any) => {
-        if (response.status === 'processing') {
-          checkInstallStatus(releaseName);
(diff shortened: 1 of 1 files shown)


AshutoshSharma-pixel:
Simplified the condition to response.status !== 'success' as suggested.

PR #795, opened 2026-06-08
backstage: Add info to plugin settings and fix lint

yolossn:
this patch adds an info alert to the plugin settings and fixes a linting error


ashu8912 (approved):
LGTM!!!

PR #808, opened 2026-06-11
fix(app-catalog): normalize chart list for empty state

Suhani95:
## Description

The App Catalog supports chart data from both Vanilla Helm repositories and Artifact Hub.

For Artifact Hub, charts are stored as an object keyed by `package_id`, while some parts of the UI were relying on array length checks. This could lead to incorrect empty-state and pagination behavior when chart data was returned in an object-shaped format.

This PR introduces a normalized` chartList` array derived from the chart data and uses it consistently for:

- Empty-state checks

- Chart rendering

- Pagination visibility

By normalizing the data once and reusing it throughout the component, the logic becomes consistent and works correctly regardless of the underlying chart data structure.

Testing

## Testing

* `npm run test`
* `npm run lint`
* `npm run tsc`

### DIFF
--- a/app-catalog/src/components/charts/List.tsx
+++ b/app-catalog/src/components/charts/List.tsx
@@ -302,7 +302,11 @@ export function ChartsList({ fetchCharts = fetchChartsFromArtifact }) {
       fetchIcons();
     }
   }, [charts]);
-
+  const chartList = charts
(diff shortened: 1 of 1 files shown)


Suhani95:
@illume, could you please take a look when you have a chance? Thanks!

PR #831, opened 2026-06-17
cert-manager: Fix ready status getters and add condition tests

vtushar06:
## What

Three cert-manager resource getters derive readiness from status conditions. `Certificate.ready` correctly returned `status === 'True'`, but `Issuer.ready` and `CertificateRequest.ready` returned the raw condition status string. Their consumers use a truthy check (for example `issuer.ready ? 'Ready' : 'Not Ready'` in `issuers/List.tsx`, and the same in `certificateRequests/Detail.tsx`), so a `False` or `Unknown` status, being a non-empty string, rendered a not-ready resource as **Ready**.

This pulls the condition lookup into a small `conditions.ts` helper (`getConditionStatus`, `isConditionTrue`), points the getters at it, and adds unit tests. `approved` and `denied` keep returning the raw status, since their consumers already compare against `'True'`.

## Why

A not-ready Issuer or CertificateRequest showing "Ready" is misleading. Sharing the lookup also removes the duplicated condition-finding across the three resources and makes the logic testable; the plugin had no unit tests before this.

## Testing

`cd cert-manager && npm test` (also ran `tsc`, `lint`, `format --check` and `build`, all clean):

```
 src/resources/conditions.test.ts (7 tests) 4ms

 Test Files  1 passed | 1 skipped (2)
      Tests  7 passed | 1 skipped (8)
```

cc @illume

### DIFF
--- a/cert-manager/src/resources/certificate.ts
+++ b/cert-manager/src/resources/certificate.ts
@@ -1,6 +1,7 @@
 import { KubeObject } from '@kinvolk/headlamp-plugin/lib/k8s/cluster';
 import { KubeObjectInterface } from '@kinvolk/headlamp-plugin/lib/k8s/cluster';
 import { Condition, IssuerReference, SecretKeySelector } from './common';
+import { isConditionTrue } from './conditions';
 
 export type KeyUsage =
   | 'signing'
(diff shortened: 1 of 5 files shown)


vtushar06:
@illume PR is ready for final review.
On the boolean `ready`: this intentionally matches `Certificate.ready`, which already returned `status === 'True'`. The change brings `Issuer.ready` and `CertificateRequest.ready` in line with it, and the only consumers are truthy checks (`issuers/List.tsx`, `certificateRequests/Detail.tsx`), so there's no behavior change beyond fixing the not-ready-shows-as-Ready bug. I kept `approved`/`denied` as the raw status since their consumers compare against 'True' to distinguish the states, so I'd rather not split them into extra getters. `Condition['type']` is currently just `string`, so I left the helper signature as is.

vtushar06:
@illume @ashu8912 gentle follow-up, this fixes a real ready-status bug plus adds tests. Ready for review whenever.

vtushar06:
@illume, would appreciate a review here whenever you get a chance.

vtushar06:
@illume gentle follow-up, this fixes a real ready-status bug plus adds tests, threads resolved and CI green. Ready for review whenever you have time.

PR #847, opened 2026-06-20
Update Kyverno name to match the other plugins

joaquimrocha:
So it matches other plugins.


joaquimrocha:
Reviewed by Summit Goyal who gave his approval on Slack.

PR #856, opened 2026-06-26
Kyverno : Add kyverno storybooks 

itvi-1234:
Added storybooks for the Headlamp kyverno plugin

### DIFF
--- a/kyverno/package.json
+++ b/kyverno/package.json
@@ -10,7 +10,7 @@
     "lint-fix": "headlamp-plugin lint --fix",
     "package": "headlamp-plugin package",
     "tsc": "headlamp-plugin tsc",
-    "storybook": "headlamp-plugin storybook",
+    "storybook": "storybook dev -p 6006 -c .storybook",
     "storybook-build": "headlamp-plugin storybook-build",
     "test": "headlamp-plugin test",
     "i18n": "headlamp-plugin i18n"
@@ -35,6 +35,8 @@
     "typescript": "5.6.2"
   },
   "devDependencies": {
-    "@kinvolk/headlamp-plugin": "^0.13.1"
+    "@kinvolk/headlamp-plugin": "^0.13.1",
+    "@types/react-i18next": "^7.8.3",
+    "tty-browserify": "^0.0.1"
   }
 }
--- a/kyverno/src/components/PolicyList.tsx
+++ b/kyverno/src/components/PolicyList.tsx
@@ -17,10 +17,107 @@
 import { Icon } from '@iconify/react';
 import { Activity, useTranslation } from '@kinvolk/headlamp-plugin/lib';
 import { ResourceListView } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import {
+  DateLabel,
+  SectionHeader,
+  SimpleTable,
+} from '@kinvolk/headlamp-plugin/lib/components/common';
 import { Box, Chip, Link as MuiLink, Typography } from '@mui/material';
 import { usePolicyResultCounts } from '../hooks/usePolicyResultCounts';
 import { KyvernoPolicy } from '../resources/kyvernoPolicy';
 import { PolicyViewer } from './PolicyViewer';
+
+// ── Pure component for Storybook (no API calls, accepts props directly) ───
+export interface PolicyRow {
+  name: string;
+  namespace?: string;
+  ready: boolean;
+  validationFailureAction: string;
+  background: boolean;
+  ruleTypes: string[];
+  ruleCount: number;
+  creationTimestamp?: string;
+  results?: { fail: number; total: number } | null;
+}
+
+export function PurePolicyTable({
+  items,
+  onNameClick,
+}: {
+  items: PolicyRow[];
+  onNameClick?: (item: PolicyRow) => void;
+}) {
+  return (
+    <Box>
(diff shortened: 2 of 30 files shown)


srajang1805:
@itvi-1234 I noticed two regressions in `PureComplianceBadge` while reviewing the Storybook changes, so I prepared a fix.

The fix:

* Restores `useTranslation()` and the original `t()` wrappers (preserving i18n).
* Restores the original no-data empty state instead of showing `100% compliant` when there are no policy reports.

I've pushed the changes to my fork:

* Fork: `srajang1805/plugins`
* Branch: `feat/kyverno-storybooks-v3`
* Commit: `d0db780`

If you'd like to include it, you can cherry-pick it with:

```bash
git remote add srajang1805 https://github.com/srajang1805/plugins
git fetch srajang1805 feat/kyverno-storybooks-v3
git cherry-pick d0db780
```

itvi-1234:
Hey , Could you please - 
Test the branch , by running it locally , check all the edge cases , check if any good practices can be done and what left , also test if the kyverno plugin is running nice in the headlamp UI too after testing storybooks , you have to install kyverno in ur cluster and made some .yaml files to test it.after that if you feel that there might be any scope of improvement there then you can just make a PR on top of this branch.

Pragati5-DEBUG:
Hi @itvi-1234 — tested the Kyverno plugin (branch feat/kyverno-storybooks-v3) on minikube + Kyverno v1.13.

**Environment**
- Windows, Headlamp dev (frontend + backend), minikube v1.35.1
- ClusterPolicy `require-app-label` (Audit mode)
- Test resources: `kyverno-test-good`, `kyverno-test-bad`, and later `allow-test-bad-pod` PolicyException

**Headlamp UI**
- **Violations:** Working — failing pods show correct messages and pod names (`kyverno-test-bad`, volcano-job-* pods).
- **Policy Reports:** Working — `kyverno-test-good` shows Pass:1, `kyverno-test-bad` shows Fail:1.
- **Policy Exceptions:** Empty when no resources exist (expected). After applying a test PolicyException (`allow-test-bad-pod`), the row appears in Kyverno → Exceptions with policy name and detail view.
- **Plugin load / cluster connection:** OK.

**Storybook**
- Ran locally on port 6006 — Kyverno stories render correctly (ComplianceBadge, ViolationsView, PolicyReportList, etc.).

**Minor note (non-blocking)**
- Policy Exceptions empty state in live UI can look completely blank; Storybook uses “No policy exceptions found” — might be worth aligning empty-state messaging
- . Happy to help with more testing if useful.

Pragati5-DEBUG:
Hi @itvi-1234 — here's the demo video of Kyverno plugin testing in Headlamp UI + Storybook (PR #856).
Thanks for the opportunity to help test!

https://github.com/user-attachments/assets/6e920567-a9fc-4692-815f-127751cfe2b7

itvi-1234:
cc @joaquimrocha

kyverno/package.json
```diff
@@ -35,6 +35,8 @@
     "typescript": "5.6.2"
   },
   "devDependencies": {
-    "@kinvolk/headlamp-plugin": "^0.13.1"
+    "@kinvolk/headlamp-plugin": "^0.13.1",
+    "@types/react-i18next": "^7.8.3",
+    "tty-browserify": "^0.0.1"
```
joaquimrocha, line 40: Why are these deps needed?

kyverno/package.json
```diff
@@ -10,7 +10,7 @@
     "lint-fix": "headlamp-plugin lint --fix",
     "package": "headlamp-plugin package",
     "tsc": "headlamp-plugin tsc",
-    "storybook": "headlamp-plugin storybook",
+    "storybook": "storybook dev -p 6006 -c .storybook",
```
joaquimrocha, line 13: Why do we need to override this command?

kyverno/src/components/PolicyReportList.tsx
```diff
@@ -17,11 +17,94 @@
...
+  scope?: string;
+  pass: number;
+  fail: number;
+  warn: number;
+  error: number;
+  skip: number;
+  creationTimestamp?: string;
+}
+
+export function PurePolicyReportTable({
```
joaquimrocha, line 43: This is added and exported but we keep exporting PolicyReportList without using PolicyReportRow.
It seems like this is a case in other (most?) files here?

kyverno/src/components/PolicyList.tsx
```diff
@@ -40,6 +137,7 @@ function openPolicyActivity(item: KyvernoPolicy) {
 export function PolicyList() {
   const { t } = useTranslation();
   const counts = usePolicyResultCounts();
+
```
joaquimrocha, line 140: Don't add changes (newlines, in this case) that are unrelated to the commits in question.

joaquimrocha (changes requested):
I don't understand this PR. Having Pure* components is usually for replacing the actual logic within components that are not easy to test directly. This was a pattern emerging from the fact that we had components using URLs and making requests, so we'd keep these "dynamic" parts in the root components and abstract their inner logic in a Pure* component that'd just take whatever were values coming from those dynamic parts.

AFAIU, this PR is adding Pure* components sometimes when it shouldn't need to, and it's adding them without replacing the original components' inner logic. So essentially it is not testing anything that the user sees in the end.

PR #858, opened 2026-06-26
flux: generic controller support

cschindlbeck:
**Problem**

The Controllers section in Flux Checks was built on a hardcoded allowlist of pod label selectors covering only the six standard Flux controllers plus source-watcher. Any third-party controller deployed alongside Flux — such as terraform-runner-tofu-controller (Tofu/OpenTofu) — was silently omitted.

Additionally, the pod list was scoped to fluxCheck.namespace, which defaults to an empty string on non-bootstrapped clusters, causing Headlamp to fall back to listing pods across all namespaces.

**Changes**

- Switch from pods to deployments as the data source for the Controllers table. Deployments give a cleaner picture of intent (desired vs. ready replicas) compared to individual pod phase.
- Generic name-based filter: include any deployment in the flux namespace whose name contains controller or watcher. This picks up all standard Flux controllers, source-watcher, terraform-runner-tofu-controller, and any future additions without code changes.
- Namespace fix: scope the deployment list to fluxCheck.namespace || 'flux-system' so it never accidentally queries all namespaces on non-bootstrapped clusters.
- Health checks updated: useFluxHealth and the accordion status badge now compare readyReplicas === replicas instead of phase === 'Running'.
- Controllers table updated: name links to the deployment detail page, status column shows X/Y ready replicas, image path corrected to spec.template.spec.containers[0].image.

### DIFF
--- a/flux/src/overview/index.tsx
+++ b/flux/src/overview/index.tsx
@@ -155,12 +155,16 @@ function useFluxHealth(
       }
     }
 
-    // Check controller pod status.
+    // Check controller deployment readiness.
     if (controllers && controllers.length > 0) {
-      const notRunning = controllers.filter(c => c.jsonData?.status?.phase !== 'Running');
-      if (notRunning.length > 0) {
+      const notReady = controllers.filter(c => {
+        const desired = c.jsonData?.spec?.replicas ?? 1;
+        const ready = c.jsonData?.status?.readyReplicas ?? 0;
+        return ready !== desired;
+      });
+      if (notReady.length > 0) {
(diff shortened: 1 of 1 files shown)


illume:
Please check the git commit message follows the format... something like:
> flux: Support generic controllers in flux plugin within flux-system
> ...

cschindlbeck:
@illume tested in the cluster, works. also changed commit message and implemented ghcp suggestions

illume (approved):
Awesome, thanks! 🎉

PR #860, opened 2026-06-26
cluster-api: README.md: Fix screenshots not rendering in Headlamp and Artifact Hub

PixelRobots:
## Problem

The README currently references screenshots via github.com/user-attachments/assets/ URLs (images uploaded via GitHub issue/PR drag-and-drop). These URLs are not rendered by Headlamp's plugin catalog or Artifact Hub, so users see broken images instead of the actual screenshots.

image
image

## Solution

- Move the two screenshots into the repository under cluster-api/docs/
- Update README image links to use 
aw.githubusercontent.com URLs, which are reliably rendered everywhere

## After

Both images now display correctly in Headlamp's plugin catalog and on Artifact Hub.

## Changes

- \cluster-api/README.md\ — updated image URLs
- \cluster-api/docs/dashboard.png\ — added screenshot (Clusters dashboard view)
- \cluster-api/docs/map-view.png\ — added screenshot (Map view visualization)


illume (changes requested):
Thanks for this!!!

I think it needs to be updated in the 0.x.x/README.md files as well.
Please change the message to start with something like "cluster-api: README.md:"

illume (approved):
🎉 thanks!

PR #877, opened 2026-07-06
velero: Add backup coverage panels (Phase 1)

Pragati5-DEBUG:
Introduce a read-only Velero plugin that shows schedule coverage on Deployment, StatefulSet, PVC, and Namespace detail views, with install detection, plugin settings, tests, and manual-test fixtures.

Tracks kubernetes-sigs/headlamp#5198.


itvi-1234:
Could you please break down the commits ?

velero/src/hooks/useVeleroData.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface VeleroDataState {
+  loading: boolean;
+  error: Error | null;
+  schedules: ScheduleCoverageInput[];
+  backups: BackupCoverageInput[];
+  getCoverageForWorkload: (target: WorkloadTarget) => ScheduleCoverageResult[];
+  getSchedulesForNamespace: (namespace: string) => ScheduleCoverageResult[];
+}
+
+export function useVeleroData(): VeleroDataState {
```
itvi-1234, line 50: could you document them?

velero/src/coverage.ts
```diff
@@ -0,0 +1,149 @@
+import { formatNextScheduledRun } from './cron';
+import type { VeleroBackupTemplate } from './resources/velero';
+
+export interface WorkloadTarget {
+  namespace: string;
+  labels: Record<string, string>;
+  resourceKind: 'deployments' | 'statefulsets' | 'persistentvolumeclaims';
+}
+
+export interface ScheduleCoverageInput {
```
itvi-1234, line 19: documenting them too?

velero/src/coverage.test.ts
```diff
@@ -0,0 +1,257 @@
...
+          name: 'backup-1',
+          scheduleName: 'daily',
+          phase: 'Completed',
+          completionTimestamp: '2026-07-02T10:00:00Z',
+        },
+      },
+    ]);
+  });
+
+  test('partial namespace coverage only matches labeled workloads', () => {
```
itvi-1234, line 411: have you test them locally ?

itvi-1234 (commented):
Could you please take a look?

Pragati5-DEBUG:
@itvi-1234 Thanks for the review!



**Documentation:** Added JSDoc to the coverage types/helpers and `useVeleroData` / `VeleroDataState`.

**Testing:** `npm test` passes (13 unit tests). Manually verified on minikube with Velero + fixtures from `velero/test-files/` — covered (`nginx-demo`, `web` StatefulSet) and not-covered cases. Details in `velero/test-files/README.md`.

itvi-1234:
@Pragati5-DEBUG , I think we can also document other files too and add some details

velero/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@headlamp-k8s/velero",
+  "version": "0.1.0",
+  "description": "Velero backup coverage panels for Headlamp",
```
itvi-1234: I think It should cover the whole idea of the velero plugin , not just about this PR

Pragati5-DEBUG:
@itvi-1234 Thanks — want to make sure I match what you have in mind:

1. **package.json `description`** — expand to describe the full Velero plugin (coverage panels on Deployment/STS/PVC/Namespace detail views, read-only Phase 1). OK?

2. **Other file docs** — planning JSDoc on the remaining public modules:
   - `resources/velero.ts` (Schedule/Backup wrappers)
   - `useVeleroInstalled.ts`
   - `config.ts`
   - `cron.ts`
   
   Anything else you had in mind (e.g. `index.tsx`, panel components, README)?

Happy to push once you confirm scope. Thanks!

Pragati5-DEBUG:
@itvi-1234 Done — pushed an update:

- **package.json** `description` now describes the full Velero plugin (coverage on Deployment/STS/PVC/Namespace detail views, read-only).
- **JSDoc** added on `resources/velero.ts`, `config.ts`, `useVeleroInstalled.ts`, `cron.ts`, `index.tsx`, `BackupCoveragePanelPure`, and `settings.tsx` (in addition to the earlier `coverage.ts` / `useVeleroData.ts` docs).

Please take another look when you have time. Thanks!

itvi-1234:
cc @illume

Pragati5-DEBUG:
https://github.com/user-attachments/assets/13c86b64-b03e-4644-afcd-6afaf5431efe

reasonerjt:
@Pragati5-DEBUG 

Some resources may be covered by velero via the resource policy:
https://velero.io/docs/main/resource-filtering/#resource-policies
Can this plugin capture these resources ?

Pragati5-DEBUG:
@reasonerjt Thanks!

For this plugin (Phase 1): no — it does not capture resources covered via resource policies.

It only evaluates Schedule template filters (included/excluded namespaces & resources, and label selectors) for Deployment / StatefulSet / PVC / Namespace.

Resource-policy awareness isn’t implemented here; we can add it as a follow-up if useful.

velero/src/coverage.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  const included = template.includedNamespaces ?? [];
+  if (included.length === 0) {
+    return true;
+  }
+
+  return included.includes(namespace);
+}
+
+function labelsMatch(template: VeleroBackupTemplate, labels: Record<string, string>): boolean {
```
kaovilai: labelsMatch() only evaluates matchLabels. A valid selector using matchExpressions therefore appears empty and matches every resource. Velero also supports orLabelSelectors, which is absent from the normalized type. This can incorrectly report an uncovered workload as protected—the most dangerous failure mode for this plugin. Velero’s Schedule API supports both ordinary label selectors and OR selectors.

Pragati5-DEBUG:
@kaovilai Thanks — good catch. You’re right that Phase 1 only handled `matchLabels`, so selectors with `matchExpressions` (and `orLabelSelectors`) can look empty and incorrectly report a workload as covered.

I’ll fix this for Phase 1:
1. Evaluate `matchLabels` + `matchExpressions`
2. Support `orLabelSelectors` (match if any selector matches)
3. Fail closed when a selector can’t be evaluated (no false “covered”)
4. Add unit tests for those cases

Will push an update shortly.

velero/src/resources/velero.ts
```diff
@@ -0,0 +1,88 @@
...
+  orLabelSelectors?: LabelSelector[];
+  includedResources?: string[];
+  excludedResources?: string[];
+}
+
+/** Velero Schedule CRD (velero.io/v1). */
+export interface VeleroScheduleSpec {
+  schedule?: string;
+  template?: VeleroBackupTemplate;
+}
```
kaovilai, line 32: P1: Account for paused schedules before reporting coverage

Could we carry spec.paused into ScheduleCoverageInput and either exclude paused schedules from coverage or clearly display them as paused?

A paused schedule will not create new backups, but the current implementation still reports the resource as covered and computes a “Next run.” That can give operators the false impression that the resource has active, ongoing backup protection. Please also add a unit test for a paused schedule.

velero/src/coverage.ts
```diff
@@ -0,0 +1,230 @@
...
+export function scheduleCoversWorkload(
+  schedule: ScheduleCoverageInput,
+  target: WorkloadTarget
+): boolean {
+  const template = schedule.template ?? {};
+  return (
+    namespaceIncluded(template, target.namespace) &&
+    resourceIncluded(template, target.resourceKind) &&
+    labelsMatch(template, target.labels)
+  );
```
kaovilai, line 234: P1: Honor the resource-level Velero exclusion label

Before reporting coverage, please check whether the target has velero.io/exclude-from-backup: "true". Velero skips an individually labeled resource even when it otherwise matches the schedule’s namespace, resource, and label filters.

Without this check, the panel can report a resource as covered even though Velero will deliberately omit it from the backup. Please add tests for this on at least a workload and a PVC.

velero/src/coverage.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  const aliases = resourceAliases(kind);
+  return included.some(name => aliases.has(name) || included.includes(kind));
+}
+
+function namespaceIncluded(template: VeleroBackupTemplate, namespace: string): boolean {
+  const excluded = template.excludedNamespaces ?? [];
+  if (excluded.includes(namespace)) {
+    return false;
+  }
```
kaovilai, line 123: P1: Apply Velero’s namespace glob matching semantics

These exact includes() checks do not handle the glob patterns supported by Velero namespace filters, such as app-*. In particular, an excludedNamespaces pattern such as prod-* will not match prod-api, causing the plugin to report coverage even though Velero excludes that namespace.

Could this use matching compatible with Velero’s namespace-glob behavior? Please add tests for both globbed inclusion and exclusion, especially the false-positive exclusion case.

kaovilai (approved):

itvi-1234:
Great work @Pragati5-DEBUG ,
cc @illume

Pragati5-DEBUG:
Hi @illume  Phase 1 is open as PR #877 and I've rebased it onto main. Would love your feedback.

PR #907, opened 2026-07-15
kmesh: components: Implement Authorization Policy viewer

itvi-1234:
## Description
This PR implements the **Authorization Policy Viewer** (Section 11.7) for the Kmesh Headlamp plugin. It introduces a new read-only dashboard to view security policies active in the daemon when running in workload / dual-engine mode.

### Changes Included
- **New API Hook (`useAuthzPolicies`)**: Extracts the `policies[]` array directly from the `/debug/config_dump/dual-engine` config dump endpoint.
- **AuthzPolicies Component**: Renders a `SimpleTable` detailing active policies, including their `Name`, `Namespace`, `Scope`, and `Action` (e.g., ALLOW/DENY).
- **Rules Inspector (Dialog)**: Replaced a bulky inline JSON view with a clean "View Rules" button. Clicking this opens a Material-UI Dialog displaying the raw evaluated `rules` (from `security.proto`) in a formatted JSON block.
- **Routing**: Registered a new sidebar entry under Kmesh (`Auth Policies`) and mapped it to the `/kmesh/authz-policies` route.

### DIFF
--- a/kmesh/src/components/daemon/AuthzPolicies.tsx
+++ b/kmesh/src/components/daemon/AuthzPolicies.tsx
@@ -1,0 +1,147 @@
+import { SectionBox, SimpleTable } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import {
+  Box,
+  Button,
+  CircularProgress,
+  Dialog,
+  DialogContent,
+  DialogTitle,
+  Typography,
+} from '@mui/material';
+import { useState } from 'react';
+import { useKmeshDaemonPods } from '../../hooks/useKmeshDaemonPods';
+import type { DaemonAuthorizationPolicy } from '../../types/daemonApi';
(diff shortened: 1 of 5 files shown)


illume (changes requested):
Thanks for this.


The commit should probably have a better idea of where the changes are in the subject line. eg. kmesh/src/components/daemon/AuthzPolicies.tsx

Also, rather than having a central "hooks" folder with all sorts of different things in there... it makes sense to instead co-locate hooks for things closer to where they are used. Especially for things which are only used within one or two localised components.

itvi-1234:
Hi @illume , Thanks for the suggestions , I have applied them and pushed , Please take a look whenever have some time

itvi-1234:
Hi @illume , I have rebased the branch

illume (approved):
🎉 thanks!

PR #909, opened 2026-07-16
prometheus: Settings: Fix connection test crash on plain-text response

### LINKED ISSUE #6541: bug(plugins/prometheus): Prometheus connection fails with "SyntaxError: Unexpected token 'P' ... is not valid JSON" due to plain-text health check
**Describe the bug**
When configuring an external Prometheus service address (e.g., `monitoring/prometheus-service:9090`) in Headlamp, the connection fails with the console error:

```text
Connection failed: Unexpected token 'P', "Prometheus"... is not valid JSON
```

This is because Headlamp queries the standard Prometheus health check endpoint `/-/healthy` to verify the connection, but attempts to parse the response as JSON. By default, Prometheus returns a plain-text payload (`text/plain; charset=utf-8`):

```text
Prometheus Server is Healthy.
```

Because the payload starts with `P`, the frontend JSON parser crashes immediately.

**To Reproduce**

1. Run Headlamp app image.
2. Configure a service/endpoint that targets an external, standard Prometheus server over port `9090` (HTTP).
3. Set the Prometheus service address in Headlamp's plugin settings.
4. Try to connect. Observe the JSON parsing crash in the browser's developer console.

**Expected behavior**
Headlamp should handle the `text/plain` MIME-type returned by standard Prometheus' `/-/healthy` endpoint, or verify health via a JSON-producing endpoint (like `GET /api/v1/targets` or `GET /api/v1/query?query=up`).

**Environment:**

* **Headlamp version:** v0.43.0 git commit 2ca733d234a5eca4159ccd28279c0f8787fbb6e3
* **Installation type:** app-image
* **Prometheus backend:** Standard Prometheus server (external)

**Screenshots/Logs:**

```bash
$ curl -i http://<prometheus-host>:9090/-/healthy
HTTP/1.1 200 OK
Date: Thu, 16 Jul 2026 10:28:37 GMT
Content-Length: 30
Content-Type: text/plain; charset=utf-8

Prometheus Server is Healthy.
```

I have an external prometheus server anyway and just created a monitoring/prometheus-service in the cluster and use haproxy to proxy it to my actual prometheus server. While that part is working, it would be much better if I could configure the external prometheus server in headlamp directly.

Athang69:
## Summary
This PR fixes a crash in the Prometheus plugin connection test by replacing the plain text /-/healthy health check with a JSON producing query endpoint.

## Related Issue
Fixes kubernetes-sigs/headlamp#6541

## Changes
- Updated the connection test in Settings.tsx to query /api/v1/query?query=up instead of /-/healthy
- Added a status check on the response body to confirm the query API actually succeeded, not just that something answered on the port

## Steps to Test
1. Deploy a standard external Prometheus server and expose it as a service in your cluster, for example on port 9090
2. In Headlamp, open the Prometheus plugin settings and configure the namespace, service, and port pointing to that Prometheus instance
3. Click Test Connection and confirm it succeeds instead of throwing a JSON parse error in the console

## Notes for the Reviewer
- Standard Prometheus servers return text/plain from /-/healthy, for example "Prometheus Server is Healthy.", which crashed the JSON parser with "Unexpected token 'P'... is not valid JSON"
- Swapped to /api/v1/query?query=up, which always returns JSON and validates the query engine end to end rather than just liveness
- Verified locally with a script that mocks both the old and new proxy responses, reproducing the original crash and confirming the fix avoids it, without needing a live cluster
- Also confirmed clean tsc, eslint, and npm run build on this branch

### DIFF
--- a/prometheus/src/components/Settings/Settings.tsx
+++ b/prometheus/src/components/Settings/Settings.tsx
@@ -115,11 +115,26 @@ export function Settings(props: SettingsProps) {
         subPath = '/' + subPath;
       }
 
-      const proxyUrl = `/clusters/${selectedCluster}/api/v1/namespaces/${namespace}/services/${service}:${port}/proxy${subPath}/-/healthy`;
-      await request(proxyUrl, {
+      // Standard Prometheus servers return a plain-text body from /-/healthy
+      // (e.g. "Prometheus Server is Healthy."), which isn't valid JSON. Some
+      // setups also return non-JSON error pages for other statuses. Testing
+      // against the query API instead gets us a real JSON response either
+      // way, and it verifies the query engine actually works, not just that
+      // something is listening on the port.
+      const normalizedSubPath = subPath.replace(/\/+$/, '');
+      const queryParams = new URLSearchParams({ query: '1' });
+      const proxyUrl = `/clusters/${selectedCluster}/api/v1/namespaces/${namespace}/services/${service}:${port}/proxy${normalizedSubPath}/api/v1/query?${queryParams.toString()}`;
+      const response = await request(proxyUrl, {
         method: 'GET',
-        isJSON: false,
       });
+
+      if (response?.status !== 'success') {
+        const detail =
+          typeof (response as any)?.error === 'string' && (response as any).error.trim()
+            ? (response as any).error.trim()
+            : response?.status ?? 'unknown';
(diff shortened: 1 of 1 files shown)


Athang69:
@illume I've addressed the copilot's comment. PTAL. 
Thanks!!

PR #910, opened 2026-07-17
cert-manager: Add i18n

skoeva:
These changes add i18n support for the cert manager plugin in Headlamp.

Locales supported:
en, de, es, fr, it, ja, id, ko, pt-BR, ru, zh-Hans, zh-Hant, cs, nl, hu, pt-PT, pl, sv, tr

### Remaining work
Add translations for different locales


illume (approved):
🎉 thanks!

PR #902, opened 2026-07-11
metal3: Link BareMetalHost and Metal3Machine details to each other

mastermaxx03:
## Summary

Makes the link between a BareMetalHost and its Metal3Machine clickable in the detail views. A host is consumed by a machine, and the machine records which host it claimed, so this lets you click straight from one to the other instead of copying a name and searching for it. First piece of the resource chain.

## Included

- On the BareMetalHost detail, the Consumer field links to its Metal3Machine when the consumer is one; other consumer kinds stay plain text.
- On the Metal3Machine detail, the BareMetalHost field links to the host it claimed, read from the metal3.io/BareMetalHost annotation.
- A small parseHostAnnotation helper with unit tests for the "namespace/name" parsing.

## How to test

- A cluster with the Metal3 and CAPM3 CRDs installed, a host with a consumerRef pointing at a Metal3Machine, and that machine carrying the metal3.io/BareMetalHost annotation.
- cd metal3 && npm install && npm start, then run Headlamp.
- Open a BareMetalHost that has a consumer, click the Consumer link to reach its Metal3Machine, then click that machine's BareMetalHost link back.

PR #913, opened 2026-07-18
metal3: Add a Map view for the Machine to BareMetalHost chain

mastermaxx03:
This adds the Map view for the Metal3 plugin. It registers a custom GraphSource
that draws the links Headlamp cannot work out on its own.

There are two such links. The first is Metal3Machine to BareMetalHost. A
Metal3Machine records the host it claimed in the `metal3.io/BareMetalHost`
annotation, not as an ownerReference, so Headlamp does not draw it. The second
is the Cluster API Machine to Metal3Machine. The Metal3Machine has an
ownerReference back to its Machine, but the cluster-api plugin only links
Cluster API resources to each other and this plugin only links Metal3 resources,
so that reference falls between the two and neither draws it. This source emits
both edges.

The full chain is Cluster → Machine → Metal3Machine → BareMetalHost. The top
half, Cluster to Machine, is drawn by the cluster-api plugin, so seeing the whole
chain needs that plugin installed. Without it, the Map still shows the
Metal3Machines and BareMetalHosts and the link between them.

On a cluster that runs the bare metal operator without Cluster API, the
Metal3Machine CRD is absent. The source checks for the CRD and contributes
nothing when it is missing, so the Map does not hang and still shows the hosts.

The edge builders are pure functions in `mapEdges.ts` with unit tests in
`mapEdges.test.ts`.

Stacked on #902, which adds the annotation helper this reuses.

## Screenshots

_Full chain, Cluster → Machine → Metal3Machine → BareMetalHost:_
image

_The Metal3Machine → BareMetalHost link on its own:_
image

### DIFF
--- a/metal3/src/index.tsx
+++ b/metal3/src/index.tsx
@@ -16,6 +16,7 @@
 
 import {
   registerDetailsViewHeaderAction,
+  registerMapSource,
   registerRoute,
   registerSidebarEntry,
 } from '@kinvolk/headlamp-plugin/lib';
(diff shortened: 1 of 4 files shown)


mastermaxx03:
@Rozzii @illume  Ready for review!

Rozzii (commented):
Exactly what I wanted to see, thank you very much!

mastermaxx03:
@illume Ready to be merged!

illume (approved):
🎉 thanks!

PR #917, opened 2026-07-20
strimzi: Fix crash when the Kafka cluster list is still loading

cesaroangelo:
## Problem

Opening the **Kafka Topics** or **Kafka Users** page throws
`Cannot read properties of null (reading 'map')` and blanks the page.
It is reliably reproducible on a cluster that has no topics or users (the
lists render before the Kafka cluster list has resolved).

`KubeObject.useList().items` is typed `Array<T> | null` and is `null` while
the query is loading, but both list components mapped over it directly to
build the create-dialog namespace/cluster dropdowns.

## Fix

- Normalise the cluster list to an array at the `useList` call site.
- Extract the dropdown derivations into null-tolerant helpers
  (`clusterNamespaces`, `clusterNamesInNamespace`) with unit tests covering
  the null/loading case, so the crash cannot silently regress.
- De-duplicate the shared `KafkaClusterRef` type into `utils`.

It's a small, self-contained change, and the null handling is covered by
unit tests. `tsc`, `eslint`, `prettier` and the full test suite (71 tests)
pass.

## Note

This is a focused, standalone fix for the crash only. #850 also touches this
null crash, but bundles it with a v1-resources rework that has since landed
on `main` separately, and appears stale.

### DIFF
--- a/strimzi/CHANGELOG.md
+++ b/strimzi/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Fixed
+- Crash on the Kafka Topics and Users pages when the Kafka cluster list was still loading
+
 ## [0.3.9] - 2026-03-07
 
 ### Added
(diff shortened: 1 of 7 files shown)


illume (approved):
🎉 thanks!

PR #921, opened 2026-07-21
strimzi: Declare both API versions on the resource classes

cesaroangelo:
## Problem

On a Strimzi 1.0+ cluster, which only serves the `v1` API, the **Kafka Clusters**
list stays empty ("Resource not found") and the detail pages fail to load, while
the Topics and Users lists work. The failing requests go to `v1beta2`:

```
GET /apis/kafka.strimzi.io/v1beta2/kafkas                     404
GET /apis/kafka.strimzi.io/v1beta2/namespaces/…/kafkas/…      404
```

The plugin picked a `v1` or `v1beta2` KubeObject class at render time, with the
`v1` variants subclassing the `v1beta2` ones:

```ts
export class Kafka   extends KubeObject { static apiVersion = 'kafka.strimzi.io/v1beta2'; }
export class KafkaV1 extends Kafka      { static apiVersion = 'kafka.strimzi.io/v1'; }
```

Headlamp caches the computed endpoint on the class (`_internalApiEndpoint`), and
static members are inherited in JavaScript, so whichever variant is used first
wins. The version probe starts out reporting `v1beta2` until it resolves, so the
landing page renders with the base class, that caches its `v1beta2` endpoint, and
`KafkaV1` then reads the same cached endpoint through the prototype chain and
keeps querying `v1beta2` forever. Topics happen to work only because you reach
them after the probe has resolved, so the `v1` variant is touched first — which is
why the failure looked erratic.

## Fix

Declare every supported version on a single class and let Headlamp negotiate the
one the cluster serves (its endpoint dispatcher moves to the next entry on a 404):

```ts
static apiVersion = ['kafka.strimzi.io/v1', 'kafka.strimzi.io/v1beta2'];
```

This is the pattern the `flux` and `kubeflow` plugins already use. It removes the
five `*V1` subclasses and the per-component class selection, so there is no longer
any inheritance to poison — 60 lines lighter overall.

`useStrimziApiVersions` stays, but only for what it is actually needed for: the
direct `ApiProxy.request` paths (create/edit/delete, node pools, pod sets) and
telling a cluster without Strimzi apart from one that has no resources yet.

## Testing

Verified on a Strimzi **1.1.0** cluster serving only `v1`
(`kafka.strimzi.io/v1`, CRD `v1 served=true storage=true`):

| Build | Lists | Kafka Clusters | Detail pages |
| --- | --- | --- | --- |
| published `0.4.0-alpha` | broken | broken | broken |
| current `main` | topics/users OK | empty | fail to load |
| this branch | OK | OK | OK |

`tsc`, `eslint` and the full test suite (71 tests) pass.

### DIFF
--- a/strimzi/src/components/KafkaConnectList.tsx
+++ b/strimzi/src/components/KafkaConnectList.tsx
@@ -6,7 +6,7 @@ import {
   type ColumnType,
   type ResourceTableColumn,
 } from '@kinvolk/headlamp-plugin/lib/components/common';
-import { KafkaConnect, KafkaConnectV1 } from '../resources/kafkaConnect';
+import { KafkaConnect } from '../resources/kafkaConnect';
 import { readyChipProps } from '../utils/readyChip';
 import { useStrimziApiVersions } from '../hooks/useStrimziApiVersions';
 import { StrimziNotInstalledMessage } from './StrimziNotInstalledMessage';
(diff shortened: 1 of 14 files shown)


illume (approved):
🎉 thanks!

PR #925, opened 2026-07-23
argocd: README: Clarify RBAC behavior for actions

Joshna907:
Clarifies that Sync/Refresh actions are hidden when Kubernetes RBAC denies patch permission, and that the 403 snackbar is only shown if a patch request is attempted and denied.


illume (approved):
🎉

PR #927, opened 2026-07-24
strimzi: Refresh the connector actions cell when the state changes

cesaroangelo:
## Problem

Pausing a connector from the Kafka Connectors list leaves the button on "Pause".
The Desired state column updates, but the Actions cell keeps the button it
rendered first, so the correct "Resume" only shows up after reopening Headlamp.

## Fix

The Actions column returned a constant value from `getValue`, so the cell value
never changed. Return the desired state instead, the same way the Status and
Desired state columns already follow the resource.

Tested on a Strimzi 1.1 cluster: pausing and resuming now switch the button
straight away. `tsc`, `eslint` and the test suite (71 tests) pass.

### DIFF
--- a/strimzi/src/components/KafkaConnectorList.tsx
+++ b/strimzi/src/components/KafkaConnectorList.tsx
@@ -139,7 +139,9 @@ export function KafkaConnectorList() {
     {
       id: 'actions',
       label: 'Actions',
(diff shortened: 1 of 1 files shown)


illume (approved):
🎉 thanks!

PR #929, opened 2026-07-24
argocd: namespaces: Add GitOps insights section

Joshna907:
## Summary
This PR adds a Namespace GitOps Insights section to the built-in Headlamp Namespace detail page. It gives users Argo CD context directly where they are already inspecting namespace resources.

### Why Namespace GitOps Insights?
In Argo CD, an Application CR can be stored in one namespace, usually the Argo CD control-plane namespace, while deploying workloads into another namespace. Headlamp's default Namespace page shows Kubernetes-native objects like ResourceQuotas, LimitRanges, and Pods, but it does not explain which GitOps Applications are connected to that namespace.

This section closes that gap by showing Applications that are either stored in the current namespace or targeting it as their destination namespace. This keeps the namespace page useful without replacing the normal Argo CD Application views.

### Added
src/components/namespaces/ArgoNamespaceInsights.tsx ? Adds a Namespace detail section with summary cards for Applications stored here, Applications deploying here, and Applications needing attention. Also includes a table with Application, scope, project, target namespace, sync status, and health status.

### Changed
src/index.tsx ? Registers a details view sections processor that inserts Namespace GitOps Insights after the Namespace metadata section.
src/index.test.tsx ? Adds test coverage to verify the Namespace GitOps Insights section is registered in the expected position.

### Steps to Test
Have a running Kubernetes cluster with Argo CD installed and Headlamp accessible.
Navigate to a Namespace detail page, for example the argocd namespace.
Verify the Namespace GitOps Insights section appears after the metadata section.
Verify Applications stored in or targeting that namespace appear in the table.
Verify the summary cards show stored, deploying, and attention counts.
Navigate to a namespace with no related Argo CD Applications and verify the empty state appears.

### DIFF
--- a/argocd/src/components/namespaces/ArgoNamespaceInsights.tsx
+++ b/argocd/src/components/namespaces/ArgoNamespaceInsights.tsx
@@ -1,0 +1,266 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


ashu8912:
Hii @Joshna907  can we name it just GitOps Insights inside Namespace detail as we are already in namespace detail view

Joshna907:
> Hii @Joshna907 can we name it just GitOps Insights inside Namespace detail as we are already in namespace detail view

Okayy done

ashu8912 (approved):
LGTM!!

PR #945, opened 2026-07-28
Flux: Fix map load if flux is not installed

ashu8912:


illume (approved):

illume:
The CI was stalled... tried to close/open to see if it works now.

PR #790, opened 2026-06-05
fix(knative): prefer 'ingress-class' key over legacy 'ingress.class'

Psykii22:
### Description
This PR updates the Knative plugin to recognize the modern `ingress-class` key in the `config-network` ConfigMap. 

As stated in the [Knative documentation](https://knative.dev/docs/serving/config-network-adapters/#ingress-configurations), the legacy `ingress.class` key with a dot has been superseded by the `ingress-class` key with a dash. Previously, Headlamp only checked for the legacy key, which caused it to incorrectly report a non-configured ingress class for newer deployments (like Contour).

### Changes Made
- Updated `Overview.tsx`, `List.tsx`, and `NetworkingSection.tsx` to check for `ingress-class` first.
- Added a fallback to the legacy `ingress.class` key to ensure backwards compatibility with older Knative deployments.
- Updated the documentation comment in `ingress.ts`.

Fixes #787

PR #949, opened 2026-07-28
knative: recognize newer ingress-class ConfigMap key

### LINKED ISSUE #787: knative: ingress.class ConfigMap key is outdated
According to the [Knative documentation](https://knative.dev/docs/serving/config-network-adapters/#ingress-configurations) the `ingress.class` key in the _config-network_ ConfigMap is outdated and has been superseeded by the `ingress-class` key.

> Look for the ingress-class key. It could also be the ingress.class key with a dot. The dash usage is more current and supersedes any key with the dot. In the following example, the ingress.class key was initially set for the Kourier controller, but is now set to Contour because the ingress key with a dash takes precedence.

Currently, the knative plugin for headlamp does not recognize this newer key and thus incorrectly reports a non-configured ingress class which leads to confusion.

The solution would be to add recognition of the newer key to the code as well and then prefer it over the other, in the same way knative does.

vipulpandey21:
## What broke

Per Knative's docs, the `ingress.class` key in the `config-network` ConfigMap was superseded by `ingress-class` (dash), which takes precedence when both are present. The knative plugin only ever checked `ingress.class`, so clusters configured with just the newer key show up as "not configured" in the plugin, even though Knative itself resolves it correctly. Reported in #787.

## Fix

Added a shared `getIngressClass()` helper in `config/ingress.ts` that checks `ingress-class` first and falls back to `ingress.class`. Replaced the three separate inline lookups (`Overview.tsx`, `List.tsx`, `NetworkingSection.tsx` all had their own copy of essentially the same trim/fallback logic) with calls to the shared helper, so the precedence is defined in one place.

## Testing

Added test cases in `ingress.test.ts` covering both keys individually, precedence when both are set, and blank-value handling. Ran locally in the `knative` plugin directory:
- `tsc --noEmit`: clean
- `vitest run`: all 27 tests pass (10 in `ingress.test.ts`)

Fixes #787

### DIFF
--- a/knative/src/components/kservices/List.tsx
+++ b/knative/src/components/kservices/List.tsx
@@ -25,7 +25,7 @@ import ConfigMap from '@kinvolk/headlamp-plugin/lib/k8s/configMap';
 import Pod from '@kinvolk/headlamp-plugin/lib/k8s/pod';
 import { Chip, Stack, Typography } from '@mui/material';
 import React, { useMemo } from 'react';
-import { formatIngressClass, INGRESS_CLASS_GATEWAY_API } from '../../config/ingress';
+import { formatIngressClass, getIngressClass, INGRESS_CLASS_GATEWAY_API } from '../../config/ingress';
 import { useAuthorization } from '../../hooks/useAuthorization';
 import { useClusters } from '../../hooks/useClusters';
 import { useKnativeInstalled } from '../../hooks/useKnativeInstalled';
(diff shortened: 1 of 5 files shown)


vipulpandey21:
Closing this — I hadn't checked for linked PRs on the issue and just found #790 already covers this exact fix and has been under review since June. Apologies for the noise, will look for something else to work on.

PR #951, opened 2026-07-28
kyverno: Probe kyverno.io/v2 independently of legacy v1

### LINKED ISSUE #950: kyverno: cleanup/exceptions/v2-reports pages show "not detected" on CEL-only installs
## Description

`useKyvernoCRDs` only probes `kyverno.io/v2` (which hosts CleanupPolicy, PolicyException, and admission/background-scan reports) when `kyverno.io/v1` (legacy ClusterPolicy/Policy) is also present:

```ts
if (legacy) {
  const v2 = await checkAPIGroup('/apis/kyverno.io/v2');
  if (v2) {
    cleanup = true;
    exceptions = true;
    kyvernoV2Reports = true;
  }
}
```

But Kyverno's own Helm chart (`charts/kyverno/charts/crds/values.yaml` in kyverno/kyverno) toggles these as independent flags — `clusterpolicies`/`policies` (v1) and `cleanuppolicies`/`policyexceptions` (v2) can each be enabled or disabled separately. A cluster running Kyverno in CEL-only mode (`clusterpolicies: false`, `policies: false`) can still have cleanup policies and exceptions enabled.

On that kind of cluster, `CRDGuard` (`requires="cleanup"` / `"exceptions"` / `"kyvernoV2Reports"`) always shows "not detected on this cluster" for the Cleanup Policies, Policy Exceptions, and admission/background-scan report pages — even though those CRDs are genuinely installed — because the `kyverno.io/v2` probe never runs when `legacy` is false.

## Expected

`kyverno.io/v2` should be probed independently (in parallel with the other API groups), not gated behind `legacy`.

## How I found this

Was reading through `useKyvernoCRDs.ts` and noticed the `if (legacy)` gate looked like it assumed v1 and v2 always ship together. Checked Kyverno's Helm chart CRD toggles to confirm cleanup/exceptions are genuinely independent of the legacy engine, which is what the code doesn't currently account for.

I have a fix and a regression test ready, opening a PR shortly.

vipulpandey21:
## What broke

`useKyvernoCRDs` only checks whether `kyverno.io/v2` (CleanupPolicy, PolicyException, admission/background-scan reports) is available when `kyverno.io/v1` (legacy ClusterPolicy/Policy) is also present:

```ts
if (legacy) {
  const v2 = await checkAPIGroup('/apis/kyverno.io/v2');
  ...
}
```

Kyverno's own Helm chart (`charts/kyverno/charts/crds/values.yaml`) toggles `clusterpolicies`/`policies` (v1) and `cleanuppolicies`/`policyexceptions` (v2) as independent flags, so a CEL-only install (`clusterpolicies: false`, `policies: false`) can still have cleanup policies and exceptions enabled. On that kind of cluster, `CRDGuard` (`requires="cleanup"` / `"exceptions"` / `"kyvernoV2Reports"`) always shows "not detected on this cluster" for those pages, even though the CRDs are genuinely installed, because the v2 probe never runs.

Filed as #950.

## Fix

Probe `kyverno.io/v2` in the same `Promise.all` as the other API groups instead of gating it behind `legacy`.

## Testing

Added `useKyvernoCRDs.test.ts` (none existed before) covering:
- CEL-only + v2 present → cleanup/exceptions/kyvernoV2Reports all `true` (the regression case — confirmed this fails against the old code with `cleanup: false`)
- nothing installed → everything `false`
- stock install (v1 + v2 both present) → everything `true`

Ran locally in the `kyverno` plugin directory:
- `tsc --noEmit`: clean
- `vitest run`: all 8 tests pass

Fixes #950

### DIFF
--- a/kyverno/src/hooks/useKyvernoCRDs.test.ts
+++ b/kyverno/src/hooks/useKyvernoCRDs.test.ts
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { ApiProxy } from '@kinvolk/headlamp-plugin/lib';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { probeCluster } from './useKyvernoCRDs';
+
+vi.mock('@kinvolk/headlamp-plugin/lib', () => ({
+  ApiProxy: {
+    request: vi.fn(),
+  },
+}));
+
+function mockApiGroups(available: string[]) {
+  vi.mocked(ApiProxy.request).mockImplementation((path: string) => {
+    if (available.includes(path)) {
+      return Promise.resolve({});
+    }
+    return Promise.reject(new Error('not found'));
+  });
+}
+
+describe('probeCluster', () => {
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


vipulpandey21:
@joaquimrocha @itvi-1234 could one of you take a look when you have a moment?

PR #952, opened 2026-07-28
plugin-catalog: add Install/Update buttons on plugin cards

### LINKED ISSUE #142: plugin-catalog: add install/upgrade button to card
I think it would be nice to be able to install/upgrade a plugin from the catalog with one click, instead of having to open the plugin detail page and perform the actions there.

So for example, having the Install button on the right here:

image

Clickable from the plugin card, probably where the "Installed" label is currently sitting:

image

This would also look more consistent

Pragati5-DEBUG:
Fixes #142
## Summary
- Add **Install** / **Update** buttons on Plugin Catalog cards so users can install or upgrade without opening the detail page
- Reuses the existing `PluginManager` install/update + confirm/reload flow from the detail view
- Fixes #142

## Test plan
- [ ] Run `npm start` in `plugin-catalog` and load it in desktop Headlamp
- [ ] Uninstalled plugin card shows **Install** → confirm → install succeeds → reload → card shows **Installed**
- [ ] Installed plugin with a newer version shows **Update** and updates successfully
- [ ] Up-to-date installed plugin shows **Installed** (no button)
- [ ] Storybook: PluginCard NotInstalled / Installed / UpdateAvailable

image
image

### DIFF
--- a/plugin-catalog/src/components/plugins/PluginCard.tsx
+++ b/plugin-catalog/src/components/plugins/PluginCard.tsx
@@ -18,6 +18,7 @@ import {
 import React, { useEffect, useState } from 'react';
 import { PluginPackage } from './List';
 import PluginIcon from './plugin-icon.svg';
+import { pollPluginManagerStatus } from './pluginManagerAction';
 
 export interface PluginCardProps {
   plugin: PluginPackage;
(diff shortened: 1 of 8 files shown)


plugin-catalog/src/components/plugins/PluginCard.tsx
```diff
@@ -1,30 +1,179 @@
...
+              {alertMessage ? alertMessage.substring(0, Math.min(50, alertMessage.length)) : null}
+            </Typography>
+          </Tooltip>
+        }
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
+        action={pluginSnackbarAction(() => setAlertMessage(null), t)}
+      />
       <Card
         sx={{
           height: '100%',
```
skoeva, line 189:
```suggestion
          display: 'flex',
          flexDirection: 'column',
          height: '100%',
```

plugin-catalog/src/components/plugins/PluginCard.tsx
```diff
@@ -182,16 +330,11 @@ export function PluginCard(props: PluginCardProps) {
         </CardContent>
         <CardActions
           sx={{
-            justifyContent: 'space-between',
+            justifyContent: 'flex-end',
```
skoeva, line 344:
```suggestion
            justifyContent: 'flex-end',
            marginTop: 'auto',
```

plugin-catalog/src/components/plugins/PluginCard.tsx
```diff
@@ -1,30 +1,179 @@
...
+  }, [currentAction, identifier, t]);
+
+  const handleInstall = (event: React.MouseEvent) => {
+    event.preventDefault();
+    event.stopPropagation();
+    setCurrentAction('Install');
+    PluginManager.install(identifier, plugin.display_name || plugin.name, artifactHubURL);
+  };
+
+  const handleUpdate = async (event: React.MouseEvent) => {
```
skoeva: could we pass the installed plugin name into the card instead of calling `PluginManager.list()` again here?

plugin-catalog/src/components/plugins/PluginCard.tsx
```diff
@@ -1,30 +1,179 @@
...
+    event.preventDefault();
+    event.stopPropagation();
+    PluginManager.cancel(identifier);
+    setCurrentAction(null);
+  };
+
+  let actionContent: React.ReactNode;
+  if (currentAction) {
+    actionContent = (
+      <Button variant="contained" onClick={handleCancel} sx={actionButtonSx} size="small">
```
skoeva: could this button have an accessible label, like `aria-label={t('Cancel')}`?

skoeva (changes requested):
thanks, this looks useful. I left some suggestions to keep the buttons aligned at the bottom right and improving the loading button's accessibility.

the card also duplicates some action and polling logic from `Detail.tsx`. Could that be shared, or tracked as follow-up work? a small interaction test for Install and Update would be helpful too.

PR #953, opened 2026-07-28
prometheus: Refresh catalogs for 18 languages

skoeva:
These changes complete i18n coverage for the Prometheus plugin in Headlamp by making the remaining chart labels, metric names, accessibility text, and user-visible fallback values translatable.

The changes also make dynamically configured chart labels extraction-safe, update the English source catalog, and refresh the locale catalogs while preserving locale-specific plural forms.

Locales supported:
en, de, es, fr, it, ja, id, ko, pt-BR, ru, zh-Hans, zh-Hant, cs, nl, hu, pt-PT, pl, sv, tr

### Remaining work
Add translations for the different locales.


illume (approved):
🎉 thanks!

PR #955, opened 2026-07-29
Fix cross-namespace HelmChart source links

pratik340:
## Summary

This fixes source links for HelmChart resources that reference a source in a different namespace.

The link was always using the HelmChart namespace, even when spec.sourceRef.namespace was set. As a result, clicking the source link could navigate to the wrong resource.

This change uses spec.sourceRef.namespace when available and falls back to the HelmChart namespace otherwise.

## Testing

- Built the plugin successfully with npx headlamp-plugin build.
- Verified that the project builds without errors.


skoeva (commented):
this is half of the bug from #954, could you squash this change into that PR? we can close this one

PR #964, opened 2026-07-30
keda: Fix a11y violations

skoeva:
These changes fix accessibility (a11y) violations in the KEDA plugin surfaced by an axe-core pass over the components. The plugin looks and behaves exactly as before for sighted mouse users, while becoming more usable for keyboard and screen-reader users.

### Changes

- Add a translated accessible name to the KEDA installation status indicator
- Hide decorative job status icons from screen readers
- Fix heading order for the scale-up and scale-down configuration sections


yolossn (approved):
LGTM

PR #968, opened 2026-07-30
metal3: Show the BMC protocol and flag a missing scheme

mastermaxx03:
## Summary

Adds a Protocol row to the BareMetalHost detail view, derived from the BMC address scheme, plus the client-side scheme check from the design.

The parser reads the scheme of the BMC address, maps it to a protocol like Redfish, iDRAC, iLO, or IPMI, and classifies it as ok, missing, or unknown against the operator's registered handlers. It handles the redfish +http and +https suffix, and it is pure and unit-tested so a create form can reuse it later.

The detail view shows the protocol for a recognised scheme, and warns when the scheme is missing or unknown, because the operator silently treats a schemeless address as IPMI and the host then fails to register with no hint.

Confirmed against the metal3-docs design, bmc-address.md, which documents both the protocol-in-scheme model and the missing-scheme-becomes-IPMI behaviour.

### DIFF
--- a/metal3/src/BareMetalHost/Details.tsx
+++ b/metal3/src/BareMetalHost/Details.tsx
@@ -16,8 +16,10 @@
 
 import { DetailsGrid } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
 import { Link } from '@kinvolk/headlamp-plugin/lib/components/common';
+import { Typography } from '@mui/material';
 import { useParams } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


mastermaxx03:
Closing this. The protocol is already visible in the BMC address, so the badge is largely redundant, and the scheme validation it adds only pays off inside a create form, which isn't on the plan. Dropping it to keep the plugin's surface focused for the alpha.

Issue #969, opened 2026-07-31
Allow configuring the default Prometheus URL declaratively for Helm Headlamp deployments


nixon89:
## Problem

PR #508 added support for configuring a custom Prometheus URL, which is a great improvement.

However, the URL is currently stored only in the user's local settings (ConfigStore). This means every user has to manually open the plugin settings and configure the Prometheus URL in their browser.

For GitOps-managed Headlamp deployments this is problematic because there is currently no way to provide a deployment-wide default configuration.

## Use case

We deploy Headlamp via the official Helm chart and manage everything declaratively (GitOps).

We would like all users to have a working Prometheus integration immediately after opening Headlamp, without any manual configuration.

Ideally, administrators should be able to specify a default Prometheus URL during deployment, while still allowing users to override it in their personal settings if necessary.

## Expected behavior

The plugin should support a deployment-wide default Prometheus URL.

The precedence could be:

```
User settings (ConfigStore)
    >
Deployment configuration
    >
Built-in defaults
```

This keeps the current behavior while enabling fully declarative deployments.

## Possible implementation

I'm intentionally not suggesting a specific implementation, since I'm not familiar with the preferred configuration mechanism within Headlamp.

Some possible approaches could include:

- plugin configuration passed by Headlamp
- deployment configuration exposed by Headlamp
- ConfigMap-backed configuration
- another mechanism already used by Headlamp plugins

The important part is having a way for cluster administrators to configure a default Prometheus endpoint without requiring every user to configure it manually.

Would such functionality be considered for the Prometheus plugin?

singhshresth26:
Hi @nixon89 , I would like to work on this issue.

Here is my proposed approach to support declarative, deployment-wide Prometheus URL configuration:

1. **ConfigMap-Backed Defaults**:
   The plugin will fetch default settings from a ConfigMap in the `kube-public` namespace labeled with `headlamp.kubernetes.io/prometheus-config=true`. This allows all users to read the configuration regardless of their namespace-specific permissions.

   A sample ConfigMap:
   ```yaml
   apiVersion: v1
   kind: ConfigMap
   metadata:
     name: headlamp-prometheus-plugin-config
     namespace: kube-public
     labels:
       headlamp.kubernetes.io/prometheus-config: "true"
   data:
     address: "monitoring/prometheus-k8s:9090"
     subPath: "/prometheus"
     defaultTimespan: "24h"
     defaultResolution: "medium"
     autoDetect: "false"
[...]

PR #973, opened 2026-07-31
feat(prometheus): add deployment-wide default configuration via Confi…

mishradwaterlaw:
## Description

This PR introduces support for a deployment-wide default configuration in the Prometheus plugin, leveraging a Kubernetes `ConfigMap`.. it fixes/builds on #969 

### Motivation

Currently, the Prometheus plugin stores the configured URL and other settings exclusively in the user's local `ConfigStore`. While this works well for individual preferences, it creates friction for administrators managing Headlamp deployments declaratively via GitOps they cannot easily provision a working out-of-the-box Prometheus integration for all users without manual configuration per browser.

### Solution

This change implements a mechanism to fetch cluster-wide default settings by reading from a `ConfigMap` labeled `headlamp.kubernetes.io/prometheus-config=true`. 

A new configuration resolution precedence is established:
1. **User Settings** (Local `ConfigStore` - highest priority)
2. **Deployment Configuration** (`ConfigMap`)
3. **Built-in Defaults** (e.g., `autoDetect: true`)

### Changes Made

- **`util.ts`**:
  - Added `fetchDeploymentConfig` to asynchronously fetch a Prometheus config `ConfigMap` across the active cluster.
  - Added `getEffectiveClusterConfig` which merges user settings, deployment settings, and default settings according to the precedence rules.
  - Refactored `getPrometheusInterval`, `getPrometheusResolution`, and `getPrometheusSubPath` to be asynchronous so they can accurately return effective configurations.
- **Chart Components**: Refactored all 8 Prometheus chart components (e.g., `GenericMetricsChart`, `KarpenterChart`, `StrimziChart`) to properly `await` the configuration getters inside their `useEffect` lifecycle hooks.
- **`Settings.tsx` UI**: Updated the settings page to properly reflect deployment defaults as placeholders/hints when no local overrides are present. It now gracefully displays the effective configuration without eagerly persisting the defaults back to the local `ConfigStore`, ensuring true user overrides are sparse and intentional.

## How to Test

1. Create a ConfigMap in your cluster with the required label (e.g., in `kube-public` namespace):
    ```yaml
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: headlamp-prometheus-default
      namespace: kube-public
      labels:
        headlamp.kubernetes.io/prometheus-config: "true"
    data:
      address: "monitoring/prometheus:9090"
      defaultTimespan: "1h"
      defaultResolution: "medium"
      autoDetect: "false"
      isMetricsEnabled: "true"
    ```
2. Open Headlamp without any prior Prometheus configuration.
3. Navigate to a metrics view (e.g., a node or pod with charts). The plugin should automatically use the `monitoring/prometheus:9090` endpoint.
4. Navigate to the Prometheus settings page. Ensure the fields dynamically reflect the `ConfigMap` defaults.
5. Override a setting (e.g., set a new address). Ensure your local settings take precedence over the `ConfigMap` defaults.

### DIFF
--- a/prometheus/src/components/Chart/CapiChart/CapiChart.tsx
+++ b/prometheus/src/components/Chart/CapiChart/CapiChart.tsx
@@ -66,6 +66,9 @@ export function CapiChart(props: CapiChartProps) {
   const [refresh, setRefresh] = useState<boolean>(true);
   const [prometheusPrefix, setPrometheusPrefix] = useState<string | null>(null);
   const [state, setState] = useState<prometheusState>(prometheusState.LOADING);
+  const [timespan, setTimespan] = useState('24h');
+  const [resolution, setResolution] = useState('medium');
+  const [subPath, setSubPath] = useState<string | null>(null);
   const [isVisible, setIsVisible] = useState<boolean>(false);
   useEffect(() => {
     const isEnabled = clusterConfig?.[cluster]?.isMetricsEnabled || false;
@@ -84,6 +87,13 @@ export function CapiChart(props: CapiChartProps) {
         if (prefix) {
           setPrometheusPrefix(prefix);
           setState(prometheusState.INSTALLED);
+          
+          const loadedInterval = await getPrometheusInterval(cluster);
+          const loadedRes = await getPrometheusResolution(cluster);
+          const loadedSubPath = await getPrometheusSubPath(cluster);
+          setTimespan(loadedInterval ?? '1h');
+          setResolution(loadedRes ?? 'medium');
+          setSubPath(loadedSubPath);
         } else {
           setState(prometheusState.UNKNOWN);
         }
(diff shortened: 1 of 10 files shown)


mishradwaterlaw:
hi @illume I've updated the PR to address the copilot review...

- **`util.ts`**: Caching and promises are now properly cluster-scoped. Transient fetch errors are no longer permanently cached, and the ConfigMap lookup has been restricted to the `kube-public` namespace to minimize RBAC requirements.
- **`Settings.tsx`**: Added proper type safety (`Partial<ClusterData>`) and updated the UI so deployment defaults act as field placeholders instead of overriding user input values.
- **Charts**: Initialized `timespan` to `'24h'` across all chart components to prevent flickering while the async configuration loads.

PR #982, opened 2026-08-01
fix: prevent setState-after-unmount in Chart.tsx auto-refresh setInte…

### LINKED ISSUE #979: [Bug] setState-after-unmount & stale closure in Chart.tsx auto-refresh setInterval
The shared Chart component (src/components/Chart/Chart/Chart.tsx) sets up a 10-second setInterval to auto-refresh metrics. Two defects exist in this refresh loop:

setState after unmount  when the user navigates away while a fetch is in progress (triggered by the interval tick), the awaited fetchMetrics call resolves on a dead component and the subsequent setMetrics, setState, and setError calls run against an unmounted tree.

Stale closure over fetchMetricsData fetchMetricsData is an async arrow function defined in the render scope. It closes over props.interval, props.resolution, props.prometheusPrefix, and props.subPath at the time of the last render. The setInterval callback captures the fetchMetricsData reference at effect-setup time. If any of those props change after the interval is set up but before the next tick fires, the interval callback calls the stale closure, which then fetches with the old interval/resolution/prefix values.

The effect's dependency array (lines 210–215) does include all those props, so the interval is cleared and recreated on prop change  this mitigates the stale-closure problem as long as the component stays mounted, but does not guard against the unmount case.

Affected file is Chart.tsx

```tsx

// fetchMetricsData — defined inside the render scope, closes over props
const fetchMetricsData = async (plots, firstLoad = false) => {
  // ...
  for (const plot of plots) {
    try {
      response = await fetchMetrics({
        prefix: props.prometheusPrefix,   // ← stale if props changed after effect ran
        query:  plot.query,
        from:   currentTimeRange.from,
        // ...
      });
    } catch (e) {
      setError(e.message);               // ← no unmount guard
      setState(ChartState.ERROR);
      break;
    }
  }
  setMetrics(mergedData);               // ← no unmount guard
  setState(ChartState.SUCCESS);
};

// Auto-refresh effect
useEffect(() => {
  let refreshInterval: NodeJS.Timeout;

  if (props.autoRefresh) {
    refreshInterval = setInterval(() => {
      fetchMetricsData(props.plots, false); // ← stale fetchMetricsData closure
    }, 10000);
  }

  return () => {
    if (refreshInterval) {
      clearInterval(refreshInterval);       // ← clears interval but in-flight fetch
    }                                       //   continues running unguarded
  };
}, [props.autoRefresh, props.interval, props.resolution,
    props.prometheusPrefix, props.subPath, JSON.stringify(props.plots)]);


```

Steps to Reproduce
setState after unmount:

Navigate to a page with a Chart component and autoRefresh={true}.
Wait for the 10-second interval to tick and start a fetch.
Navigate away before the fetch resolves.
Observe React warning about state update on unmounted component in the console.
Stale props (edge case):

Render a Chart with autoRefresh={true} and interval="1h".
Change the interval prop to "24h" just after an interval tick starts but before fetchMetrics resolves.
Observe that the in-flight request still uses the old "1h" time range parameters.

I can work on this issue myself,

mishradwaterlaw:
Fixes #979 

What was happening:

fetchMetricsData had no cancellation mechanism. Any setState / setMetrics / setError call inside it could fire against an unmounted component in two scenarios:

User navigates away while a setInterval-triggered fetch is still in flight
A prop change re-runs the initial-load effect while the previous async chain hasn't resolved yet

How i fixed it:

Added a cancelledRef: { current: boolean } parameter to fetchMetricsData. Every state-setter is now guarded by a cancelledRef.current check.

Both useEffects create their own cancelledRef and flip it to true in their cleanup function — cancelling any in-flight async work before the next run starts.

Changes
src/components/Chart/Chart/Chart.tsx  1 file, no behaviour change when the component stays mounted

### DIFF
--- a/prometheus/src/components/Chart/Chart/Chart.tsx
+++ b/prometheus/src/components/Chart/Chart/Chart.tsx
@@ -86,7 +86,8 @@ export default function Chart(props: ChartProps) {
 
   const fetchMetricsData = async (
     plots: Array<{ query: string; name: string; dataProcessor: (data: any) => any }>,
-    firstLoad: boolean = false
+    firstLoad: boolean = false,
+    cancelledRef: { current: boolean } = { current: false }
   ) => {
     const fetchedMetrics: {
       [key: string]: {
@@ -95,16 +96,17 @@ export default function Chart(props: ChartProps) {
       };
     } = {};
 
-    if (firstLoad) {
+    if (firstLoad && !cancelledRef.current) {
       setState(ChartState.LOADING);
     }
 
     // clear any previous errors on refresh
-    if (!firstLoad) {
+    if (!firstLoad && !cancelledRef.current) {
       setError(null);
     }
 
     for (const plot of plots) {
+      if (cancelledRef.current) return;
       var response;
       try {
         // recalculate time range for each fetch to ensure we get current data
@@ -119,6 +121,7 @@ export default function Chart(props: ChartProps) {
           subPath: props.subPath,
         });
       } catch (e) {
+        if (cancelledRef.current) return;
         fetchedMetrics[plot.name] = { data: [], state: ChartState.ERROR };
         setError(e.message);
         setState(ChartState.ERROR);
@@ -137,6 +140,8 @@ export default function Chart(props: ChartProps) {
       const data = plot.dataProcessor(response);
       fetchedMetrics[plot.name] = { data: data, state: ChartState.SUCCESS };
     }
+
+    if (cancelledRef.current) return;
 
     // if all the plots are in no data state, set the state to no data
     if (Object.values(fetchedMetrics).every(plot => plot.state === ChartState.NO_DATA)) {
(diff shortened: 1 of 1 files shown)


mishradwaterlaw:
hi @skoeva can you please review this once..,

Bhargava-Ram-Thunga:
Hi @mishradwaterlaw,

I'm reviewing this PR! I checked the code diff — using `cancelledRef` inside the `useEffect` cleanup functions properly guards `setState` and `setError` calls when the component unmounts during in-flight metrics fetches. 

I'll also test and review this further to help get it ready for maintainer review. Great work!

Bhargava-Ram-Thunga (commented):
Hi @mishradwaterlaw,

I'm reviewing this PR! I checked out the branch and ran `npm run tsc`, `npm run lint`, and the test suite locally — all pass (36 tests). The fix itself is the canonical stale-async guard done right: each `useEffect` owns its own `cancelledRef` and flips it in cleanup, and the early `return` inside the plot loop also stops issuing further fetches after unmount rather than just muting the setters. Consolidating the interval cleanup to unconditionally run `clearInterval(refreshInterval)` is safe (`clearInterval(undefined)` is a no-op).

Two non-blocking suggestions:

1. The default parameter `cancelledRef: { current: boolean } = { current: false }` means any future call site that forgets to pass a ref silently gets no cancellation. (Both current call sites pass it correctly.) Making it a required argument would let TypeScript catch unguarded call sites added later.
2. Longer-term, an `AbortController` threaded into the fetch layer would also cancel the in-flight network requests, not just the state updates — fine as a follow-up.

A small regression test (unmount mid-fetch, assert no state update) would lock this in, but I don't consider it blocking.

PR #983, opened 2026-08-01
fix: restore DiskMetricsChart toolbar visibility

### LINKED ISSUE #980: [Bug] DiskMetricsChart toolbar controls are permanently invisible due to width: '0%'
The DiskMetricsChart component renders its toolbar (the "Disk" label and the pause/play IconButton) inside a Box container that has width: '0%' set inline. This collapses the container to zero width, making all controls invisible to the user. The pause/play button cannot be used to stop auto-refresh.

Affected File : 

DiskMetricsChart.tsx

```tsx

<Box
  display="flex"
  justifyContent="space-around"
  alignItems="center"
  style={{ marginBottom: '0.5rem', margin: '0 auto', width: '0%' }}
  {/*                                                  ^^^^^^^^^^  */}
  {/*                               zero width collapses the container */}
>
  {state === prometheusState.INSTALLED && (
    <>
      <Box>{t('Disk')}</Box>
      <Box pl={2}>
        <IconButton
          onClick={() => { setRefresh(prev => !prev); }}
          size="large"
        >
          {refresh
            ? <Icon icon="mdi:pause" width="20px" height="20px" />
            : <Icon icon="mdi:play"  width="20px" height="20px" />
          }
        </IconButton>
      </Box>
    </>
  )}
</Box>


```

width: '0%' is almost certainly a copy-paste error from an earlier iteration of the layout. Every sibling chart component (KarpenterChart, StrimziChart, KedaChart) uses a proper full-width flex row for its toolbar. The width: '0%' overrides the default width: auto / 100% block behaviour, and the display: flex children are clipped entirely inside a zero-width viewport.

Note also that margin: '0 auto' overrides marginBottom: '0.5rem' (shorthand margin takes precedence), so the bottom margin is lost too  a secondary consequence of the same line.

Steps to Reproduce
Navigate to any page that renders DiskMetricsChart with a PersistentVolumeClaim that has disk metrics available.
Observe that no "Disk" label or pause/play button appears above the chart area.
Inspect the DOM — the Box and IconButton elements are present but have width: 0.

mishradwaterlaw:
fixes #980 

The toolbar Box had width:'0%' set inline, which collapsed the container to zero pixels. The 'Disk' label and pause/play IconButton were rendered in the DOM but completely invisible and inaccessible to users, making it impossible to pause auto-refresh on the disk chart.

The margin shorthand (margin:'0 auto') also overrode the intended marginBottom:'0.5rem', losing the bottom spacing as a side effect.

Replace the inline style with the same MUI Box props used by sibling charts (KedaChart, KarpenterChart): display=flex, justifyContent= flex-end, alignItems=center, mb={2}.

### DIFF
--- a/prometheus/src/components/Chart/DiskMetricsChart/DiskMetricsChart.tsx
+++ b/prometheus/src/components/Chart/DiskMetricsChart/DiskMetricsChart.tsx
@@ -83,12 +83,7 @@ export function DiskMetricsChart(props: DiskMetricsChartProps) {
   const subPath = getPrometheusSubPath(cluster);
   return (
     <SectionBox>
-      <Box
-        display="flex"
-        justifyContent="space-around"
-        alignItems="center"
-        style={{ marginBottom: '0.5rem', margin: '0 auto', width: '0%' }}
-      >
+      <Box display="flex" justifyContent="flex-end" alignItems="center" mb={2}>
         {state === prometheusState.INSTALLED && (
           <>
             <Box>{t('Disk')}</Box>


Bhargava-Ram-Thunga:
Hi @mishradwaterlaw,

I'm reviewing this PR! Removing the inline `width: '0%'` from the `Box` container in `DiskMetricsChart.tsx` restores the toolbar controls and pause/play button visibility.

I'll also test and review this further to help get it ready for maintainers. Clean fix!

Bhargava-Ram-Thunga (commented):
Hi @mishradwaterlaw,

I'm reviewing this PR! LGTM from my side — I checked out the branch and ran `tsc`, `lint`, and the test suite locally, all green. The diagnosis is correct: `width: '0%'` collapses the toolbar container to zero pixels, and the `margin: '0 auto'` shorthand coming after `marginBottom` in the same style object overrides the intended bottom spacing. Replacing the inline `style` with standard Box props (`display="flex" justifyContent="flex-end" alignItems="center" mb={2}`) restores the Disk label and pause/play button.

One tiny accuracy note on the description: the sibling charts' toolbars are similar but not identical (KedaChart uses `gap={1} justifyContent="flex-end" mb={2}` without `alignItems`, KarpenterChart uses `space-between`), so this is "consistent in style" rather than a literal copy — which is fine, and the `space-around` → `flex-end` / `0.5rem` → `mb={2}` changes read as deliberate improvements, not regressions. Pure style fix, no logic touched, nothing else needed.

PR #986, opened 2026-08-02
feat(knative): add Serving resource views and navigation

kahirokunn:
views for the Serving resources the plugin did not cover yet, the ones Serving derives from a Service (Configurations, Routes, Pod Autoscalers, Metrics, KIngresses, Serverless Services, Certificates, Images).

With that many entries a flat menu stops working, so the sidebar is now grouped into three sections using the section headers from Headlamp 0.44: Serving (Services, Revisions, Domain Mappings), Configuration (the networking page, renamed from "Overview"), and Serving Internals (the new views plus the existing Cluster Domain Claims one).

Notes on the implementation:

- The new views are read-only on purpose: no create button or row actions, and the default Edit/Delete/Scale/Restart actions are hidden on the Custom Resource detail view for exactly those eight kinds. Cluster Domain Claims sit in the same section but keep their create action, for clusters where autocreate-cluster-domain-claims is disabled.
- DomainMapping and ClusterDomainClaim each carried their own copy of the CRD routing boilerplate and the new kinds would have multiplied it, so the first commit moves that into a shared KnativeCustomResource base class.
- The README documents each view, who it is for, and which ones are intentionally read-only.

CleanShot 2026-08-02 at 14 22 01@2x
CleanShot 2026-08-02 at 14 22 52@2x
CleanShot 2026-08-02 at 14 22 58@2x
CleanShot 2026-08-02 at 14 23 06@2x
CleanShot 2026-08-02 at 14 23 14@2x
CleanShot 2026-08-02 at 14 23 24@2x
CleanShot 2026-08-02 at 14 23 33@2x
CleanShot 2026-08-02 at 14 23 41@2x
CleanShot 2026-08-02 at 14 23 49@2x
CleanShot 2026-08-02 at 14 23 57@2x
CleanShot 2026-08-02 at 14 24 10@2x
CleanShot 2026-08-02 at 14 24 19@2x
CleanShot 2026-08-02 at 16 18 41@2x
CleanShot 2026-08-02 at 20 44 29@2x


kahirokunn:
@illume Hi ✋ . I would greatly appreciate it if you could leave a review when you have time.
Thank you.

knative/src/config/ingress.ts
```diff
@@ -17,17 +17,30 @@
...
 export const INGRESS_CLASS_GATEWAY_API = `gateway-api${INGRESS_CLASS_SUFFIX}`;
 
 // Future ingress classes (commented out for reference)
 // export const INGRESS_CLASS_CONTOUR = `contour${INGRESS_CLASS_SUFFIX}`;
 // export const INGRESS_CLASS_ISTIO = `istio${INGRESS_CLASS_SUFFIX}`;
 
+export function readIngressClass(
+  configData: Record<string, string | undefined> | null | undefined
+): { raw: string | null; value: string | null } {
+  const raw = configData?.[INGRESS_CLASS_CONFIG_KEY] ?? null;
```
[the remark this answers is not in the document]
kahirokunn: Good catch. The plugin used to read `ingress.class`, so switching to `ingress-class` only was a regression. I’ve updated it to prefer `ingress-class` and fall back to `ingress.class`. The precedence and blank-value cases are covered by tests too.

knative/src/mapView.tsx
```diff
@@ -15,310 +15,610 @@
...
+
+function isResource(resource: any, resourceClass: ResourceClass): boolean {
+  const apiVersions = Array.isArray(resourceClass.apiVersion)
+    ? resourceClass.apiVersion
+    : [resourceClass.apiVersion];
+  return (
+    resource?.kind === resourceClass.kind &&
+    apiVersions.some(apiVersion => resource?.apiVersion === apiVersion)
+  );
+}
```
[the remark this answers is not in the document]
kahirokunn: You’re right. Headlamp’s `KubeObject` exposes `kind` on the instance, but `apiVersion` lives under `jsonData`, so the previous check would never match. I’ve updated the matcher to use `jsonData` when present and fall back to raw resource data, with tests for both shapes.

illume:
Nice one.

Can you please update the commit messages with the "knative: area: Verb ..." format?

kahirokunn:
@illume Thank you. I did it 🙏

knative/src/config/gatewayReadiness.ts
```diff
@@ -0,0 +1,202 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type ReadinessState = 'ready' | 'not-ready' | 'unknown';
+
+interface GatewayCondition {
```
illume, line 31: These interfaces and fields could do with documentation. Are these defined anywhere in the docs?

knative/src/readOnlyResources.ts
```diff
@@ -0,0 +1,62 @@
...
+    apiVersion && kind && READ_ONLY_KNATIVE_GVKS.has(`${apiVersion}/${kind}`)
+  );
+  const hasControllerOwner = resource.jsonData.metadata?.ownerReferences?.some(
+    owner => owner.controller
+  );
+
+  return isReadOnlyKind && Boolean(hasControllerOwner);
+}
+
+export function filterReadOnlyKnativeHeaderActions<T extends { id: string }>(
```
illume, line 68: Please add docs for these functions?

knative/src/mapTopology.ts
```diff
@@ -0,0 +1,444 @@
...
+
+import type {
+  GraphEdge,
+  GraphNodeStatus,
+} from '@kinvolk/headlamp-plugin/lib/components/resourceMap/graph/graphModel';
+import type { KubeOwnerReference } from '@kinvolk/headlamp-plugin/lib/k8s/cluster';
+import type { Traffic } from './resources/knative';
+import { findReadyCondition, getFirstOwner } from './resources/knative/resourceData';
+
+export type TopologyResource = {
```
illume, line 33: Docs for these please?

illume (changes requested):
Thanks for this.

I left some notes.

For the new things, I think some more test-files/ would be handy for testing.

knative/src/config/gatewayReadiness.ts
kahirokunn: Thanks, I added comments for the interfaces and fields, with links to the relevant Gateway API and Kubernetes Service definitions.

knative/src/readOnlyResources.ts
kahirokunn: Done. I added comments explaining what each helper checks and when a resource is treated as read-only.

knative/src/mapTopology.ts
kahirokunn: Done. I added docs for the resource shape and topology helpers, including how resources are matched and connected.

knative/src/index.tsx
```diff
@@ -98,105 +117,86 @@ registerSidebarEntry({
...
+
+knativeNavigationItems.forEach(item => {
+  const routeName = item.routeName as KnativeListRouteName;
+  registerRoute({
+    path: item.path,
+    sidebar: item.name,
+    name: routeName,
+    component: withQueryClient(listRouteComponents[routeName]),
+  });
 });
```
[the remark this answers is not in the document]
kahirokunn: Fixed. Route names now keep their literal types, and missing component registrations are checked at runtime. I added tests for this too.

kahirokunn:
I added fixtures for the generated Serving resources and image resolution failures, along with setup and cleanup scripts. I also added a guide showing which fixtures to use for each view and Map relationship.

Suyog241005:
Great work @kahirokunn! Structuring the sidebar into Serving, Configuration, and Serving Internals makes navigation much cleaner. 

I haven't fully gone through all the code yet since it's a huge PR, but looking through the PR overview and screenshots, a few curious questions popped up:

1.  In the Map screenshot, the graph connects `Service` ➔ `Configuration` ➔ `Revision`. If a Service splits traffic across multiple active revisions (e.g. 80%/20%), does the edge label display the traffic split percentage between those revision nodes?
2.  The diagnostic alert on the Networking Configuration card (`AddressMissing` for Gateway IPs) looks super helpful! Does it surface similar status diagnostics when `config-network` uses Kourier or Istio instead of Gateway API?
3. The `Ready` status badges across all 8 internal tables look very consistent. Are these conditions parsed automatically via the new `KnativeCustomResource` base class?

kahirokunn:
Thanks for taking a look and for the thoughtful questions!

1. Yes. For an 80/20 traffic split, the Map creates two `Route → Revision` edges labeled `80%` and `20%`. The `Service → Configuration → Revision` edges represent ownership, so traffic percentages intentionally appear on the Route edges instead. The Map prefers the resolved traffic from `Route.status.traffic` and falls back to `spec.traffic`. Tags are also included, for example `0% (candidate)`.

2. Not at the moment. The detailed diagnostics shown here are currently specific to Gateway API. Kourier, Istio, and Contour expose different resources and status models, so providing equivalent diagnostics would require a separate implementation—and ongoing maintenance—for each controller. Covering all of them is outside the scope of this PR.

   If a concrete need arises and someone is willing to contribute and maintain support for a particular controller, we can consider adding it as a follow-up.

   Gateway API, on the other hand, is designed to provide a common interface across different networking implementations and reduce these controller-specific differences. That is why this PR focuses on Gateway API. Rather than maintaining separate integrations for every controller, I think encouraging the use of Gateway API where possible is the better direction.

   The common Knative Ingress `Ready` and `Reason` conditions are still available in the KIngresses view, regardless of the networking implementation.

3. Yes, the condition handling is shared, with one small implementation detail. `KnativeCustomResource` provides common access to CRD data and routing behavior, while `ConditionedKnativeCustomResource` reads the `Ready` condition from `status.conditions`. All eight internal resource classes inherit from that conditioned base, and their tables use the same shared `Ready` and `Reason` columns.

Suyog241005:
Thank you @kahirokunn for taking the time to explain! 

That makes total sense regarding Route vs Configuration edge modeling, and I 100% agree with prioritizing Gateway API as the standard networking interface to prevent controller fragmentation.

illume (approved):
🎉 thanks!

PR #987, opened 2026-08-02
fix(plugin-catalog): handle ArtifactHub fetch errors gracefully in PluginDetail

### LINKED ISSUE #866: "An error occurred while fetching plugin info from https://artifacthub.io/packages/headlamp/headlamp-plugins/headlamp_cluster-api"
Tried installing the `cluster-api` plugin as per:
- https://github.com/headlamp-k8s/plugins/blob/main/cluster-api/README.md#2-install-plugin

Getting error:
>**Failed to fetch plugin info**
>An error occurred while fetching plugin info from https://artifacthub.io/packages/headlamp/headlamp-plugins/headlamp_cluster-api.

Like this:
>Image

...over here:
>Image

Tried both with my old `headlamp` version `0.39.0`, and then again after upgrading (`brew upgrade headleamp`) to `0.43.0`.

A-d-i-t-y:
Fixes #866

## Problem
When fetching plugin info from ArtifactHub fails (network issue, proxy
error, ArtifactHub downtime, etc.), the error thrown inside
`fetchHeadlampPluginDetail()` was never caught in the `PluginDetail`
component's `useEffect`. This left the Plugin Detail page stuck on an
infinite loading spinner with no feedback to the user, instead of a
clear error message.

## Fix
Wrapped the fetch + install-status check in `initialize()` with a
try/catch block. On failure, the existing `alertMessage`/Snackbar UI
(already used elsewhere in this component for install/update/uninstall
actions) now surfaces a clear error message instead of failing silently.
Also added `isMounted` guards so state isn't updated after the component
unmounts.

## Testing
- `npm run build` passes cleanly with no errors
- Verified via direct `eslint` invocation that the code has zero
  lint errors/warnings (the `npm run lint` CLI wrapper has an unrelated
  Windows-specific issue running the command, but direct `eslint`
  confirms clean code)
- Manually verified the error path by simulating a failed fetch —
  the Snackbar now correctly displays the error message instead of an
  infinite spinner

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/plugin-catalog/src/components/plugins/Detail.tsx
+++ b/plugin-catalog/src/components/plugins/Detail.tsx
@@ -460,7 +460,7 @@ export function PluginDetail() {
         } catch (error) {
           if (isMounted) {
             const message = error instanceof Error ? error.message : String(error);
-            setAlertMessage(`Failed to fetch plugin info from ArtifactHub: ${message}`);
+            setAlertMessage(t('Failed to fetch plugin info from {{url}}: {{message}}', { url, message }));
           }
           console.error('Error fetching plugin detail:', error);
           return;
--- a/plugin-catalog/src/components/plugins/Detail.tsx
+++ b/plugin-catalog/src/components/plugins/Detail.tsx
@@ -446,22 +446,29 @@ export function PluginDetail() {
 
   useEffect(() => {
     let isMounted = true;
-
     const initialize = async () => {
       if (isMounted) {
-        const data = await fetchHeadlampPluginDetail(repoName, pluginName);
-        const enrichedPluginData = await checkIfPluginIsInstalled(data);
-        setPluginDetail(enrichedPluginData);
-        // Default to the latest version, but don't override a version the user
-        // has already picked (this runs again whenever currentAction changes).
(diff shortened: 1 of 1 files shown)


A-d-i-t-y:
Hi! I'd like to work on this. I found a related, older ask in the core repo 
(kubernetes-sigs/headlamp#2088, currently stale) requesting a similar 
Helm-chart-level Prometheus source configuration — so this seems like a 
long-standing, real gap rather than a one-off request.

My proposed approach for the plugins-repo side:
- Add a deployment-level default Prometheus URL, exposed via a mechanism 
  Headlamp already surfaces to plugins from Helm values (need to confirm 
  the right injection point — could be a ConfigMap or an env var read at 
  plugin load time)
- Precedence: existing user ConfigStore setting > deployment config > 
  current built-in default
- Keep it backward-compatible so #508's per-user override keeps working

Given this might also need a small addition on the core repo side to expose 
the Helm value to plugins, should I scope this PR to just the plugin-side 
change (assuming the config value is already reachable), or does someone 
have context on whether core-repo changes are needed too?

PR #993, opened 2026-08-03
kyverno: Fix Results column stuck on loading placeholder

### LINKED ISSUE #992: kyverno: Results column stays on a loading placeholder when PolicyReport CRDs are absent
## Description

`usePolicyResultCounts` derives its loading flag from whether the report lists are still `null`:

```ts
const { items: policyReports } = PolicyReport.useList();
const { items: clusterPolicyReports } = ClusterPolicyReport.useList();
...
loading: policyReports === null || clusterPolicyReports === null,
```

`KubeObject.useList` returns `items: null` both while a query is in flight *and* when it settles with an error — the error is reported separately, via the `errors` / `isError` fields on the same response. So when a list can never succeed, `items` stays `null` permanently and this flag never flips back to `false`.

That happens whenever the `wgpolicyk8s.io/v1alpha2` CRDs aren't readable, for example:

- Kyverno installed with the reports controller disabled, so no PolicyReport CRDs exist
- Kyverno running with `--openreportsEnabled`, where reporting moves to the `openreports.io/v1alpha1` group instead
- the user's RBAC doesn't grant list on policyreports/clusterpolicyreports

## Impact

Both policy list views read the flag to decide what to draw in the Results column:

```ts
if (counts.loading && !c) return <Typography variant="body2">…</Typography>;
if (!c || c.total === 0) return <Typography variant="body2">—</Typography>;
```

`ClusterPolicyList.tsx:106` and `PolicyList.tsx:119`

Because the flag is stuck at `true`, every row renders the `…` placeholder indefinitely and never falls through to the `—` empty state that the next line already handles. The pages otherwise work — only this column is wrong — so it reads as a permanently loading table rather than "no report data".

## Expected

Loading should reflect whether the queries have settled, not whether they produced items. A list that settled with an error should count as resolved and simply contribute no results, letting the existing `—` fallback render.

## How I found this

I was tracing how report data reaches the policy lists while working on #951, and noticed the hook destructures only `items` from `useList` and drops the error channel. Checking the SDK's `QueryListResponse` type confirmed `items` and `errors` are siblings and that `items` is `null` in the error case too, which makes the current derivation unable to distinguish "still fetching" from "failed".

I have a fix and a regression test ready, opening a PR shortly.

vipulpandey21:
## What broke

`usePolicyResultCounts` decided it was still loading by checking whether the report lists were `null`:

```ts
loading: policyReports === null || clusterPolicyReports === null,
```

`KubeObject.useList` leaves `items` at `null` when a query settles with an error, not just while it's in flight — the failure is reported alongside it via `errors` / `isError` on the same `QueryListResponse`. So if the `wgpolicyk8s.io/v1alpha2` CRDs aren't readable, the lists never become non-null and the flag stays `true` for good.

`ClusterPolicyList.tsx:106` and `PolicyList.tsx:119` both branch on it:

```ts
if (counts.loading && !c) return <Typography variant="body2">…</Typography>;
if (!c || c.total === 0) return <Typography variant="body2">—</Typography>;
```

so every row keeps rendering the `…` placeholder and never reaches the `—` empty state on the following line. Reported in #992, with the situations that trigger it (reports controller disabled, `--openreportsEnabled` moving reporting to `openreports.io/v1alpha1`, or RBAC not granting list on policyreports).

## Fix

Take loading from the queries' own `isLoading` rather than inferring it from `items`. A list that settled with an error is then treated as resolved and contributes no results, so the existing `—` fallback renders. The aggregation itself is unchanged — it already tolerates a missing list via `?? []`.

## Testing

Added `usePolicyResultCounts.test.ts` (there were no tests for this hook) covering:
- both lists settled with an error → `loading` is `false` (the regression; fails on the previous logic with `expected true to be false`)
- one list still in flight → `loading` is `true`
- one list resolved, the other errored → still aggregates the good one and reports `{ fail: 1, total: 2 }`

I confirmed the first and third cases fail against the old derivation before applying the fix.

Ran locally in the `kyverno` plugin directory:
- `tsc --noEmit`: clean
- `vitest run`: 11 tests pass across 3 files

Fixes #992

### DIFF
--- a/kyverno/src/hooks/usePolicyResultCounts.test.ts
+++ b/kyverno/src/hooks/usePolicyResultCounts.test.ts
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { renderHook } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { ClusterPolicyReport, PolicyReport } from '../resources/policyReport';
+import { usePolicyResultCounts } from './usePolicyResultCounts';
+
+vi.mock('../resources/policyReport', () => ({
+  PolicyReport: { useList: vi.fn() },
+  ClusterPolicyReport: { useList: vi.fn() },
+}));
+
+/** Mimics the shape `KubeObject.useList` returns for a still-in-flight query. */
+function pendingList() {
+  return { items: null, errors: null, isLoading: true, isError: false };
+}
+
+/** A list that settled with an error, e.g. the CRD is absent or RBAC denied it. */
+function erroredList() {
+  return { items: null, errors: [new Error('not found')], isLoading: false, isError: true };
+}
+
+function resolvedList(items: unknown[]) {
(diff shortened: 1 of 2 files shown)


vipulpandey21:
@joaquimrocha @itvi-1234 Sir could you take a look when you have a moment?

PR #997, opened 2026-08-03
kueue: extract resourceFlavor formatters and add tests

### LINKED ISSUE #988: kueue: resourceFlavor formatters are untested and not extracted like clusterQueue/localQueue
While reviewing the Kueue plugin , I noticed `resourceFlavor.ts` doesn't follow the same structure as `clusterQueue.ts` and `localQueue.ts`.

**Current state:**
- `clusterQueue.ts` has its display logic extracted into `clusterQueueFormatters.ts` 
  (exported, tested in `clusterQueue.test.ts`)
- `localQueue.ts` follows the same pattern with `localQueueFormatters.ts` / `localQueue.test.ts`
- `resourceFlavor.ts` keeps `renderTaints`, `renderTaint`, `renderTolerations`, and 
  `renderToleration` as unexported private functions in the same file — so there's no way 
  to unit test them, and no test file exists for this resource.

**Also found while reading the code:** `renderToleration()` doesn't have a defined/tested behavior for the case where `operator: 'Exists'` is combined with a `value` on the same toleration (technically an invalid manifest, but the API doesn't reject it, so it's worth locking down the display behavior with a test).

jayaprajapatii:
Fixes #988

## PR Summary

This PR brings `resourceFlavor.ts` in line with the pattern already used by `clusterQueue.ts` and `localQueue.ts`, where display-formatting logic lives in a separate, exported, and independently-tested file.

## Problem

`clusterQueue.ts` and `localQueue.ts` both extract their rendering logic into dedicated formatter files (`clusterQueueFormatters.ts`, `localQueueFormatters.ts`), which are exported and covered by their own test files (`clusterQueue.test.ts`, `localQueue.test.ts`).

`resourceFlavor.ts` didn't follow this pattern `renderTaints`,`renderTaint`,`renderTolerations`, and `renderToleration` were defined as unexported functions directly inside the resource file. Because they weren't exported, they couldn't be unit tested, and no test file existed for this resource.

While reading through the logic, I also found that `renderToleration()` had no defined or tested behavior for the case where a toleration sets `operator: 'Exists'` *and* also sets a `value`. This isn't a valid Kubernetes manifest (the API ignores `value` when `operator` is `Exists`), but since the Kubernetes API doesn't reject it outright, the display behavior is worth locking down with a test rather than leaving it implicit.

## Changes

* **Added** `resourceFlavorFormatters.ts` — exported `renderNodeLabels`, `renderTaints`, `renderTaint`, `renderTolerations`, and `renderToleration`, along with `ResourceFlavorTaintLike` / `ResourceFlavorTolerationLike` types, following the same structure as the existing `clusterQueueFormatters.ts`.
* **Added** `resourceFlavor.test.ts` — covers node label formatting, taint formatting (single and multiple), toleration formatting (including the key-less/wildcard case), `tolerationSeconds` handling (including an explicit `0`), and the `Exists` + `value` edge case described above.
* **Updated** `resourceFlavor.ts` — now imports and uses the extracted functions instead of defining them privately; no behavior change for existing use cases.

## Testing

* `npx tsc --noEmit` — no errors
* `npx headlamp-plugin lint` — no warnings
* `npx vitest run` — 17/17 tests passing 
* `npx headlamp-plugin build` — builds cleanly

### DIFF
--- a/kueue/src/resources/resourceFlavor.test.ts
+++ b/kueue/src/resources/resourceFlavor.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from 'vitest';
+import { kueueRoutePaths } from '../utils/kueueRoutes';
+import { renderNodeLabels, renderTaints, renderTolerations } from './resourceFlavorFormatters';
+
+describe('ResourceFlavor formatters', () => {
+  it('formats node labels', () => {
+    expect(renderNodeLabels({})).toBe('-');
+    expect(renderNodeLabels({ 'kubernetes.io/arch': 'amd64' })).toBe('kubernetes.io/arch=amd64');
+    expect(
+      renderNodeLabels({ 'kubernetes.io/arch': 'amd64', 'cloud.provider/spot': 'true' })
+    ).toBe('kubernetes.io/arch=amd64, cloud.provider/spot=true');
+  });
+
+  it('formats node taints', () => {
+    expect(renderTaints([])).toBe('-');
+    expect(renderTaints([{ key: 'spot', effect: 'NoSchedule' }])).toBe('spot:NoSchedule');
+    expect(renderTaints([{ key: 'spot', value: 'true', effect: 'NoSchedule' }])).toBe(
+      'spot=true:NoSchedule'
+    );
+    expect(
+      renderTaints([
+        { key: 'spot', effect: 'NoSchedule' },
+        { key: 'gpu', value: 'nvidia', effect: 'NoExecute' },
+      ])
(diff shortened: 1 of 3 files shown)


skoeva (changes requested):
hi, could you sign off your commit with `git commit -s`?

jayaprajapatii:
Hello @skoeva  Thanks for the review.
 I've signed off the commit and pushed the updated commit.

jayaprajapatii:
Hi @skoeva 
I've addressed the feedback, please have a look at this PR if anything else needed from my side for this to move forward?

PR #999, opened 2026-08-03
prometheus: Fix a11y violations

skoeva:
These changes fix accessibility (a11y) violations in the Prometheus plugin surfaced by an axe-core pass over the components. The plugin looks and behaves exactly as before for sighted mouse users, while becoming more usable for keyboard and screen-reader users.

### Changes

- Add translated accessible names to settings controls and chart selectors
- Add an accessible name to the disk chart pause/resume button
- Make the Prometheus configuration control keyboard-accessible
- Secure external links opened in a new tab with `rel="noopener noreferrer"`

### DIFF
--- a/prometheus/locales/cs/translation.json
+++ b/prometheus/locales/cs/translation.json
@@ -16,6 +16,8 @@
   "write": "",
   "Pause": "",
   "Resume": "",
+  "Timespan": "",
+  "Resolution": "",
   "CPU": "",
   "Memory": "",
   "Network": "",
--- a/prometheus/locales/de/translation.json
+++ b/prometheus/locales/de/translation.json
@@ -16,6 +16,8 @@
   "write": "",
   "Pause": "",
   "Resume": "",
+  "Timespan": "",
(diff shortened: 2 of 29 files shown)


KhyFee (commented):
﻿Nice focused a11y pass — `t()` on previously hard-coded “metric chooser”, timespan/resolution `inputProps`, pause/resume on the disk chart, and `rel="noopener noreferrer"` on the learn-more link all look right.

Two small follow-ups if you want a later pass:
1. Most locale files only add empty strings for `Timespan` / `Resolution` while `en` has real copy — fine for English, but other languages will fall back or show empty depending on i18n config; worth confirming the project’s empty-string fallback is intentional.
2. Settings “Select Cluster” / Prometheus fields are a good pattern; if any other chart still uses a native Select without `inputProps`, same treatment would finish the axe clean-up.

Approving from a review angle: shippable as-is for EN + keyboard/SR users.

yolossn (approved):

PR #1000, opened 2026-08-03
ai-assistant: Pin headlamp e2e image by digest

skoeva:
This change pins the Headlamp e2e container image by digest to ensure reproducible builds.


illume (approved):
🎉 thanks!

PR #1001, opened 2026-08-03
cert-manager: Fix a11y violations

skoeva:
These changes fix accessibility (a11y) violations in the cert-manager plugin surfaced by an axe-core pass over the components. The plugin looks and behaves exactly as before for sighted mouse users, while becoming more usable for keyboard and screen-reader users.

### Changes

- Add a translated accessible name to certificate copy buttons
- Add a translated accessible name to the cert-manager installation status indicator


yolossn (approved):
LGTM

PR #1002, opened 2026-08-03
fix(ai-assistant): redact Kubernetes Secret data before GET

### LINKED ISSUE #995: ai-assistant: Kubernetes GET responses skip redactSecrets, so Secret data reaches the model provider
## Summary

`redactSecrets()` is never applied to the responses of Kubernetes `GET` requests made by the assistant's `kubernetes_api_request` tool. The raw apiserver JSON is pushed straight into the conversation history, and that history is sent verbatim to the configured model provider on every subsequent turn.

The result is that base64 Secret `data` (passwords, TLS private keys, service-account tokens, whatever is in the Secret) leaves the cluster and goes to OpenAI/Anthropic/Azure/whatever the user configured.

This is exactly what 3ca7a9ff ("ai-assistant: Redact Tool Results and Gate Secret Access") set out to prevent. Its commit message says "Prevent Kubernetes Secret data and common credential formats from being sent to model providers", and the comment in `toolDefinitions.ts:65` claims "secret values are also redacted before reaching the LLM". That commit touched `LangChainAssistantSession.ts` and `formatToolResults.ts` but not `ai-assistant/src/api/clusterActions.tsx`, which is where the GET path actually writes to history.

There is a second, independent gap in `redactSecrets()` itself (details below) that means list responses would not be redacted even if the function were called.

## Steps to reproduce

1. Configure the AI assistant with any model provider.
2. Create a Secret:
   ```
   kubectl create secret generic db-credentials -n default \
     --from-literal=DATABASE_PASSWORD=hunter2
   ```
3. Ask the assistant: `show me the secret db-credentials in default`.
4. The model issues `kubernetes_api_request(url="/api/v1/namespaces/default/secrets/db-credentials", method="GET")`. The sensitive-path gate fires and you get an approval prompt. Approve it (this is the expected flow, the user asked for the Secret).
5. Inspect the outbound request to the provider, or `session.history`.

**Expected:** the tool message contains `"DATABASE_PASSWORD": "[REDACTED]"`. `redactSecrets()` handles this exact input correctly, see below.

**Actual:** the tool message contains `"DATABASE_PASSWORD": "aHVudGVyMg=="`.

### Why redaction never runs on GET

`KubernetesTool.handler` executes GETs immediately and returns `shouldAddToHistory: false`, because `handleActualApiRequest` already pushed the result:

`ai-assistant/packages/ai-common/src/tools/kubernetes/langchain/KubernetesTool.ts:180-183`
```ts
// The handleActualApiRequest already adds to history (now tagged with toolCallId)
return {
  content: ...,
  shouldAddToHistory: false, // Already added by handleActualApiRequest
```

The only `redactSecrets()` call on the tool-result path is guarded by that flag:

`ai-assistant/packages/ai-common/src/assistant/LangChainAssistantSession.ts:1841-1847`
```ts
if (toolResponse.shouldAddToHistory) {
  this.history.push({
    role: 'tool',
    content: redactSecrets(toolResponse.content),
    ...
```

So for GETs the branch is skipped. The entry that *does* land in history comes from `clusterActions.tsx`, which has no redaction at all:

`ai-assistant/src/api/clusterActions.tsx:712-718`
```ts
} else if (typeof response === 'object') {
  formattedResponse = JSON.stringify(response, null, 2);
  aiManager.history.push({
    success: true,
    role: 'tool',
    content: formattedResponse,
  });
```

Same for the `Table` branch (702-706) and the string branch (724-730).

Nothing downstream fixes it. `prepareChatHistory()` only filters system/display-only entries and repairs tool alignment before `convertPromptsToMessages()`:

`ai-assistant/packages/ai-common/src/assistant/LangChainAssistantSession.ts:742-758`

and `convertPromptsToMessages` passes `message.content` through untouched (`conversation/langchain/messages.ts:67-71`). `prepareChatHistory()` is used on every send path (lines 398, 894, 1004, 1396), so the unredacted entry is re-sent on every following turn of the conversation too.

`formatToolResultsForLLM` does redact, but it only runs in the orchestration path on the value returned by `executeTool`, not on the history copy, so the leak persists regardless.

### Second gap: redactSecrets misses real list responses

Even where `redactSecrets()` is called, it does not handle what the apiserver actually returns for a list. `GET /api/v1/namespaces/default/secrets` returns `kind: "SecretList"` with items that carry **no** `kind` field. Both detection points miss that:

`ai-assistant/packages/ai-common/src/security/redactSecrets.ts:145-148`
```ts
if (!/\bkind\b["\s]*:\s*["']?Secret\b/i.test(input)) {
  return input;
}
```
`Secret\b` does not match `SecretList` (`t` and `L` are both word characters, so there is no boundary), so the function returns early with the input unchanged.

`redactSecrets.ts:198`
```ts
if (obj.kind === 'Secret') {
```
would also miss the items, since list items have no `kind`.

The existing test uses `{ kind: 'List', items: [{ kind: 'Secret', ... }] }` (`redactSecrets.test.ts:269-274`), which is the `kubectl get -o json` shape for a mixed list, not the `SecretList` shape the plugin actually receives from the apiserver.

## Evidence

**Repro 1: GET path writes unredacted history.** Ran the real `handleActualApiRequest` from `ai-assistant/src/api/clusterActions.tsx` with `clusterRequest`/`getCluster` stubbed to return what the apiserver returns for `GET /api/v1/namespaces/default/secrets/db-credentials`:

```
history entry that will be sent to the LLM provider:

{
  "kind": "Secret",
  "apiVersion": "v1",
  "metadata": {
    "name": "db-credentials",
    "namespace": "default"
  },
  "data": {
    "DATABASE_PASSWORD": "aHVudGVyMg==",
    "tls.key": "LS0tLXByaXZhdGUta2V5LS0tLQ=="
  },
  "type": "Opaque"
}

leaks DATABASE_PASSWORD value: true
leaks tls.key value:           true
```

Note this same payload is redacted correctly if you pass it through `redactSecrets()`. The function works, it is just never called here.

**Repro 2: redactSecrets misses SecretList.** Calling `redactSecrets()` directly on a real `SecretList` response:

```
{
  "kind": "SecretList",
  "apiVersion": "v1",
  "metadata": { "resourceVersion": "12345" },
  "items": [
    {
      "metadata": { "name": "db-credentials", "namespace": "default" },
      "data": {
        "DATABASE_PASSWORD": "aHVudGVyMg==",
        "session": "czNjcjN0LXNlc3Npb24ta2V5"
      },
      "type": "Opaque"
    }
  ]
}

leaks DATABASE_PASSWORD value: true
leaks session value:           true
```

Output is byte-for-byte the input. `DATABASE_PASSWORD` is not caught by the generic key-name patterns either, since those require the quoted key to be exactly `password`/`secret`/`token`/etc.

## Impact

Anyone using the AI assistant against a cluster where they can read Secrets. Credentials leave the cluster boundary and land in a third-party provider's logs and possibly its training/retention pipeline. That is a disclosure that cannot be walked back, and for a lot of users it is a compliance problem on its own.

The approval gate helps but does not cover this. The user is approving "read this Secret", not "ship its contents to OpenAI", and the code explicitly promises redaction on top of the gate. It also gets worse over a conversation: once the unredacted entry is in history it is re-sent on every subsequent turn, including turns that have nothing to do with Secrets.

Non-Secret paths leak too. A GET on a Pod spec with credentials in env vars, or a ConfigMap holding a kubeconfig, goes through the same unredacted branch.

## Suggested fix

Two separate changes:

1. Redact in `clusterActions.tsx` before pushing to history. All three GET branches (`Table`, object, string) should wrap `formattedResponse` in `redactSecrets()`. The log branch at 627-631 is worth covering too, since logs routinely contain tokens. Doing it there rather than at the call site keeps the guarantee in one place for anything else that reaches `handleActualApiRequest`.

   Alternatively, stop having `handleActualApiRequest` write to history at all and let `KubernetesTool` return `shouldAddToHistory: true` so the existing redaction in `LangChainAssistantSession` applies. That is cleaner but a bigger change, since the toolCallId back-tagging loops in `KubernetesTool` depend on the current push-then-tag behaviour.

2. Make `redactSecrets` recognise list responses: match `Secret(List)?\b` in the early-out, and treat items as Secrets when the enclosing object's `kind` is `SecretList` (or when the parent list is a `SecretList` and items omit `kind`, which is the normal apiserver behaviour). A regression test using a genuine `SecretList` payload rather than the hand-built `kind: List` one would keep this from coming back.

Worth checking `isSensitiveBuiltInToolCall` coverage at the same time, though it looks correct for both the singular and list Secret paths.

Sachin-0001:
## Summary
- Kubernetes `GET` responses (including Secrets) were pushed into AI chat history without going through `redactSecrets()`, so Secret `data`/`stringData` values (passwords, tokens, TLS keys) could be sent to the configured LLM provider.
- `redactSecrets()` also failed to recognize `SecretList` responses (the shape the apiserver returns for a list of secrets), so even where it was called, list responses were not redacted.

## Fix

- `clusterActions.tsx`: redact the response content (Table, object, string, and log branches) before pushing it to `aiManager.history`.
- `redactSecrets.ts`: recognize `kind: "SecretList"` and redact `data`/`stringData` on each item, since list items omit their own `kind`.

## Test plan

- Added regression tests that fetch a Secret and a SecretList and assert the resulting chat history contains `[REDACTED]` instead of the real base64 values.
- Verified these tests fail against the old code and pass with the fix.
- Full `ai-assistant` and `ai-common` test suites pass; `tsc` and `eslint` clean.

Fixes #995

### DIFF
--- a/ai-assistant/packages/ai-common/src/security/redactSecrets.test.ts
+++ b/ai-assistant/packages/ai-common/src/security/redactSecrets.test.ts
@@ -299,6 +299,32 @@ namespace: default`;
     expect(out).toContain('"name": "db-credentials"');
   });
 
+  it('redacts a real apiserver SecretList response in YAML (`-o yaml` shape)', () => {
+    // Same SecretList shape as above but as the YAML `kubectl get secrets -o
+    // yaml` would render it: a single document, top-level `kind: SecretList`,
+    // with secrets nested under `items` (no `---` document separators).
+    const secretList = [
+      'apiVersion: v1',
+      'kind: SecretList',
+      'metadata:',
+      '  resourceVersion: "12345"',
+      'items:',
+      '- metadata:',
+      '    name: db-credentials',
+      '    namespace: default',
+      '  data:',
+      '    DATABASE_PASSWORD: aHVudGVyMg==',
+      '    session: czNjcjN0LXNlc3Npb24ta2V5',
+      '  type: Opaque',
+    ].join('\n');
+    const out = redactSecrets(secretList);
+    expect(out).not.toContain('aHVudGVyMg==');
+    expect(out).not.toContain('czNjcjN0LXNlc3Npb24ta2V5');
+    expect(out).toContain('DATABASE_PASSWORD: [REDACTED]');
+    expect(out).toContain('session: [REDACTED]');
+    expect(out).toContain('name: db-credentials');
+  });
+
   it('does not over-redact ConfigMap data', () => {
     const cm = ['kind: ConfigMap', 'data:', '  app.conf: hello-world'].join('\n');
     expect(redactSecrets(cm)).toContain('app.conf: hello-world');
--- a/ai-assistant/packages/ai-common/src/security/redactSecrets.test.ts
+++ b/ai-assistant/packages/ai-common/src/security/redactSecrets.test.ts
@@ -15,7 +15,7 @@
  */
 
 import { describe, expect, it } from 'vitest';
-import { redactSecrets } from './redactSecrets';
+import { redactSecrets, redactSecretsInValue } from './redactSecrets';
 
 describe('redactSecrets', () => {
   it('returns the input unchanged when there are no secrets', () => {
@@ -274,6 +274,31 @@ namespace: default`;
     expect(redactSecrets(list)).not.toContain('YWJj');
   });
 
+  it('redacts a real apiserver SecretList response (items omit their own kind)', () => {
+    // This is the shape returned by `GET /api/v1/namespaces/{ns}/secrets`,
+    // as opposed to `kubectl -o json`'s generic `kind: List` wrapper above.
+    const secretList = JSON.stringify({
+      kind: 'SecretList',
+      apiVersion: 'v1',
+      metadata: { resourceVersion: '12345' },
+      items: [
+        {
+          metadata: { name: 'db-credentials', namespace: 'default' },
+          data: {
+            DATABASE_PASSWORD: 'aHVudGVyMg==',
+            session: 'czNjcjN0LXNlc3Npb24ta2V5',
+          },
+          type: 'Opaque',
+        },
+      ],
+    });
+    const out = redactSecrets(secretList);
+    expect(out).not.toContain('aHVudGVyMg==');
+    expect(out).not.toContain('czNjcjN0LXNlc3Npb24ta2V5');
+    expect(out).toContain('[REDACTED]');
+    expect(out).toContain('"name": "db-credentials"');
+  });
+
   it('does not over-redact ConfigMap data', () => {
     const cm = ['kind: ConfigMap', 'data:', '  app.conf: hello-world'].join('\n');
     expect(redactSecrets(cm)).toContain('app.conf: hello-world');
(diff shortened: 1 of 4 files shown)


illume (changes requested):
Thanks for this!

- Would you mind having a look at the open review comment? Please mark it resolved when it's done?
- Please check the commit messages? See git log for examples.
- I'm marking this as draft, please mark it as ready for review when the above are addressed?

Sachin-0001:
@illume all requests are addressed

Issue #994, opened 2026-08-03
kyverno: Add support for the openreports.io reporting group


vipulpandey21:
## Description

Kyverno can now emit reports under `openreports.io/v1alpha1` instead of `wgpolicyk8s.io/v1alpha2`. It's controlled by a flag in Kyverno itself (`cmd/internal/flag.go`):

```go
flagset.BoolVar(&openreportsEnabled, "openreportsEnabled", false,
    "Use openreports.io/v1alpha1 for the reporting group")
```

with matching `openreports.enabled` / `openreports.installCrds` values in the Kyverno Helm chart, and a `pkg/openreports/` package upstream that adapts between the two shapes.

This plugin only knows the wgpolicyk8s group. `useKyvernoCRDs` probes `/apis/wgpolicyk8s.io/v1alpha2` for the `reports` flag, and `resources/policyReport.ts` declares only:

```ts
export class PolicyReport extends PolicyReportBase {
  static kind = 'PolicyReport';
  static apiName = 'policyreports';
  static apiVersion = 'wgpolicyk8s.io/v1alpha2';
  static isNamespaced = true;
}
```

So on a cluster where Kyverno is running with openreports enabled, the plugin behaves as if there is no report data at all:

- Policy Reports, Cluster Policy Reports and Violations render CRDGuard's "not detected on this cluster" banner (`index.tsx` gates all three on `requires: 'reports'`)
- `ComplianceBadge` returns `null`, so the app-bar badge disappears entirely
- the Dashboard's report-derived panels and `PolicyViewer`'s results have nothing to read
- the Results column on the policy lists has no counts

Kyverno itself is healthy in that scenario — the reports just live under a different API group.

## Why it's a small change on the wire

The two groups aren't a big translation problem. `Report` keeps `scope`, `summary` and `results` at the top level, and the result entries use the same JSON keys the plugin already reads. In `openreports.io/v1alpha1`'s `report_types.go` the Go field names differ but the serialized keys don't:

```go
Subjects    []corev1.ObjectReference `json:"resources,omitempty"`
Description string                   `json:"message,omitempty"`
```

[...]

PR #1009, opened 2026-08-04
kyverno: add openreports.io/v1alpha1 support (resources, CRD probe)

Mansi2007275:
## What

Adds support for Kyverno's `openreports.io/v1alpha1` reporting group, per #994.

This PR covers:
- New `Report` / `ClusterReport` resource classes reusing the existing
  `PolicyReportBase` (the wire format matches wgpolicyk8s.io, so no
  changes were needed to the base class or interfaces)
- `useKyvernoCRDs` now probes both `wgpolicyk8s.io/v1alpha2` and
  `openreports.io/v1alpha1`. The `reports` capability is now `true` if
  either group is present, with two new granular flags
  (`wgPolicyReports`, `openReports`) exposed for consumers that need to
  know which specific group(s) are installed
- Updated `CRDGuard`'s default messages to match the new fields, and
  reworded the `reports` message since it now covers both groups

## Deferred

Updating the consumption points (Dashboard, PolicyViewer, ViolationsView,
PolicyReportList, ClusterPolicyReportList, usePolicyResultCounts,
policyResultBucket) to actually read and merge reports from both API
groups is deferred to a follow-up PR, since:
- That work depends on #992 / #993 landing first (the list hooks
  currently can't distinguish "still loading" from "this group isn't
  installed", which matters more once there are two possible sources)
- #856 currently has open PRs touching ComplianceBadge, ViolationsView,
  PolicyReportList, and ClusterPolicyReportList — avoiding those files
  here to prevent conflicts

## Testing

- `npx tsc --noEmit` passes with no errors
- `npm run build` completes successfully
- ESLint passes with no errors/warnings

Related to #994 (partial — see Deferred section above)

### DIFF
--- a/kyverno/src/components/CRDGuard.tsx
+++ b/kyverno/src/components/CRDGuard.tsx
@@ -13,31 +13,34 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 import { useTranslation } from '@kinvolk/headlamp-plugin/lib';
 import { ReactNode } from 'react';
 import { KyvernoCRDStatus, useKyvernoCRDs } from '../hooks/useKyvernoCRDs';
 import { NotInstalledBanner } from './common';
-
 export type CRDGroup = keyof Omit<KyvernoCRDStatus, 'loading'>;
-
 interface CRDGuardProps {
   requires: CRDGroup;
   children: ReactNode;
   message?: string;
 }
-
 export function CRDGuard({ requires, children, message }: CRDGuardProps) {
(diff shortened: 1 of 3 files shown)


skoeva (changes requested):
hi, could you sign off your commit with `git commit -s`?

Mansi2007275:
Hi @skoeva, thanks for pointing that out! I've rebased all 3 commits with git commit -s and force-pushed.

Mansi2007275:
please review ..  sir

PR #1018, opened 2026-08-04
flux: Fix base a11y issues

vyncent-t:
### Summary

Baseline accessibility fixes for the Flux plugin: safer external links, non-visual icons hidden from assistive tech, accessible names for icon-only status cells, and a labeled settings toggle.

### Changes

**Decorative icons — `aria-hidden="true"`**
- `flux/src/flagger/canaries.tsx` — canary bird icon.
- `flux/src/overview/index.tsx` — health-banner status icons, empty-state Flux logo, accordion chevrons, and all `mdi:check` / `mdi:close` status icons in the Flux Checks section.

Screen readers no longer announce these purely-decorative glyphs alongside the adjacent text.

**A/B Testing column — accessible name on icon-only cells**
- `flux/src/flagger/canaries.tsx` — check / close icons now carry `aria-label="Enabled"` / `aria-label="Disabled"` so the column value is conveyed non-visually without altering the visual layout.

**Settings switch — programmatic label**
- `flux/src/settings/index.tsx` — `AutoSaveSwitch` now takes a required `label` prop and forwards it as `inputProps={{ 'aria-label': label }}` on the MUI `Switch`, giving the underlying checkbox input an accessible name (the `NameValueTable` name cell does not label the input).


KhyFee (commented):
﻿Helpful pass on the Flux plugin:
- `rel="noopener noreferrer"` on external `target="_blank"` links is the right secure default.
- Decorative bird icons with `aria-hidden` and status icons with action/state labels match icon-only a11y habits.

One nit worth a glance: if those table cells already have adjacent text that communicates Enabled/Disabled, prefer `aria-hidden="true"` on the check/close icons to avoid double announcements; if the icon is the only cue, the labels are correct.

Thanks for keeping the change focused.

KhyFee (commented):
Helpful pass: noopener on external blanks + aria on icon-only status cues. Nit: if table cells already textually announce Enabled/Disabled, hide icons instead of double-labeling. Focused change overall — thanks.

flux/src/settings/index.tsx
```diff
@@ -86,7 +86,7 @@ export function FluxSettings() {
       value: (
         <AutoSaveSwitch
           settingKey="linkHRelToKs"
-          label={linkHRelLabel}
+          label="Link HelmReleases to Kustomizations instead of HelmRepositories"
```
skoeva: this label is already translated

flux/src/settings/index.tsx
```diff
@@ -46,7 +46,7 @@ function AutoSaveSwitch({
 
   return (
     <>
-      <Switch checked={value} inputProps={{ 'aria-label': label }} onChange={handleChange} />
+      <Switch checked={value} onChange={handleChange} inputProps={{ 'aria-label': label }} />
```
skoeva: I think this diff can be removed, it doesn't look like anything changed here

flux/src/settings/index.tsx
```diff
@@ -8,16 +8,16 @@ import { type ChangeEvent, type ReactNode, useEffect, useState } from 'react';
...
+  label: string;
   defaultValue?: boolean;
   helperText?: ReactNode;
 }
 
 function AutoSaveSwitch({
   settingKey,
-  label,
   onSave,
+  label,
```
skoeva, line 19: same with these label props

flux/src/flagger/canaries.tsx
```diff
@@ -163,11 +163,11 @@ function CanaryList({ canaryResourceClass }) {
             Cell: ({ row: { original: item } }) =>
               item?.hasAbTesting ? (
                 <Box display="flex" alignItems="center">
-                  <Icon icon="mdi:check" color="#4caf50" width="20" />
+                  <Icon icon="mdi:check" color="#4caf50" width="20" aria-label="Enabled" />
```
skoeva: can this be translated?

flux/src/flagger/canaries.tsx
```diff
@@ -163,11 +163,11 @@ function CanaryList({ canaryResourceClass }) {
...
                 </Box>
               ) : (
                 <Box display="flex" alignItems="center">
-                  <Icon icon="mdi:close" color="#f44336" width="20" />
+                  <Icon icon="mdi:close" color="#f44336" width="20" aria-label="Disabled" />
```
skoeva: can this also be translated?

skoeva (commented):
left a few comments

flux/src/flagger/canaries.tsx
vyncent-t: fixed

flux/src/flagger/canaries.tsx
vyncent-t: fixed

flux/src/settings/index.tsx
vyncent-t: fixed

flux/src/settings/index.tsx
vyncent-t: fixed

flux/src/settings/index.tsx
```diff
@@ -8,8 +8,8 @@ import { type ChangeEvent, type ReactNode, useEffect, useState } from 'react';
 
 interface AutoSaveSwitchProps {
   settingKey: keyof PluginConfig;
-  label: string;
   onSave: (key: keyof PluginConfig, value: boolean) => void;
+  label: string;
```
skoeva, line 11: this can be removed too

skoeva (changes requested):
I think this is the last change

Issue #978, opened 2026-08-01
[Bug] setState-after-unmount & async race condition in chart shell components (KarpenterChart, StrimziChart, KedaChart, DiskMetricsChart)


mishradwaterlaw:
All four chart shell components share an identical useEffect that fires an unguarded async IIFE. Because the IIFE is never cancelled when the component unmounts or when its dependencies change, two defects occur at runtime:

setState after unmount — if the user navigates away while any await is in flight, every subsequent state-setter (setState, setPrometheusPrefix, setTimespan, setResolution, setSubPath) runs against an already-unmounted component. In React 18 Strict Mode this surfaces as a console warning; in production SPA navigation it causes memory leaks and can corrupt state on re-mount.

Async race on cluster switch — when clusterConfig or cluster changes rapidly (e.g. the user switches clusters), a new IIFE is launched but the previous in-flight chain is never cancelled. Whichever chain finishes last wins, so a slower response from an old cluster can silently overwrite the state that was correctly set by the new cluster.

Affected files are :

KarpenterChart.tsx

StrimziChart.tsx

KedaChart.tsx

DiskMetricsChart.tsx

Root cause:  
The pattern below is copy-pasted verbatim across all four files. There is no cleanup function returned from useEffect, so React has no way to signal the IIFE to stop.

```tsx
useEffect(() => {
  const isEnabled = clusterConfig?.[cluster]?.isMetricsEnabled || false;
  setIsVisible(isEnabled);

  if (!isEnabled) {
    setState(prometheusState.UNKNOWN);
    setPrometheusPrefix(null);
    return; // ← no cancellation token returned
  }

  setState(prometheusState.LOADING);
  (async () => { // fire-and-forget
    try {
      const prefix = await getPrometheusPrefix(cluster); // await #1
      if (prefix) {
        setPrometheusPrefix(prefix); // ← unsafe after unmount
        setState(prometheusState.INSTALLED);
        const loadedInterval = await getPrometheusInterval(cluster); // await #2
        const loadedRes      = await getPrometheusResolution(cluster);
        const loadedSubPath  = await getPrometheusSubPath(cluster);  // await #3
        setTimespan(loadedInterval ?? '1h'); // ← unsafe after unmount
        setResolution(loadedRes ?? 'medium');
        setSubPath(loadedSubPath);
      } else {
        setState(prometheusState.UNKNOWN);
      }
    } catch (e) {
      setState(prometheusState.ERROR); // ← unsafe after unmount
[...]

PR #1019, opened 2026-08-04
Fix: clear close timer on unmount to prevent stale timeout

### LINKED ISSUE #1016: strimzi: Toast's fade-out setTimeout in handleClose() isn't cleared on unmount, risking setState-after-unmount
### What happens

In `strimzi/src/components/Toast.tsx`, the auto-dismiss timer (inside the `useEffect`) is correctly captured and cleared:

const timer = setTimeout(() => {
  handleClose();
}, duration);

return () => clearTimeout(timer);

But `handleClose()` itself starts a second, uncleared timer for the fade-out animation:

const handleClose = () => {
  setIsExiting(true);

  setTimeout(() => {
    setIsVisible(false);
    setIsExiting(false);
    onClose();
  }, 300); // Match animation duration
};

This inner setTimeout id is never stored or cleared. If the component unmounts while this 300ms window is in flight (e.g. the user navigates away right as a toast starts closing), the callback still fires and calls setIsVisible, setIsExiting, and onClose on an unmounted component.

### Why it matters

- Produces a React warning ("Can't perform a React state update on an unmounted component...") in dev.
- onClose() firing after unmount can trigger unexpected side effects in the parent.
- Same category of bug already fixed elsewhere in this repo for Chart.tsx (#978, #979 — setState-after-unmount / stale closures in auto-refresh timers).

### Suggested fix

Store the timeout id in a ref and clear it on unmount, same pattern already used for the auto-dismiss timer.

### Where

strimzi/src/components/Toast.tsx, handleClose() function.

Mansi2007275:
Closes #1016

## Problem
The `Toast` component's auto-dismiss timer (inside the main `useEffect`) is 
correctly captured and cleared on unmount. However, `handleClose()` starts a 
second, uncleared `setTimeout` for the fade-out animation. If the component 
unmounts while this 300ms window is in flight (e.g. the user navigates away 
right as a toast starts closing), the callback still fires and calls 
`setIsVisible`, `setIsExiting`, and `onClose` on an unmounted component — 
producing a React "state update on an unmounted component" warning and 
potentially triggering unexpected side effects in the parent via `onClose()`.

This is the same category of bug already fixed elsewhere in this repo for 
`Chart.tsx` (#978, #979).

## Fix
Stored the fade-out timeout id in `closeTimerRef` and added a cleanup 
`useEffect` that clears it on unmount — the same pattern already used for 
the auto-dismiss timer.

## Changes
- Added `closeTimerRef` (via `useRef`)
- `handleClose` now stores its timeout in `closeTimerRef.current` instead of 
  a local variable
- Added a new `useEffect(() => { return () => clearTimeout(closeTimerRef.current) }, [])` 
  to clear it on unmount

## Testing
- Manually verified toast still shows, fades out, and auto-dismisses correctly
- Verified no console warnings when triggering a toast and unmounting the 
  parent component before the fade-out timer fires

### DIFF
--- a/strimzi/src/components/Toast.tsx
+++ b/strimzi/src/components/Toast.tsx
@@ -26,6 +26,7 @@ export function Toast({ toast, onClose }: ToastProps) {
   const isDark = muiTheme.palette.mode === 'dark';
   const [isVisible, setIsVisible] = React.useState(false);
   const [isExiting, setIsExiting] = React.useState(false);
+  const closeTimerRef = React.useRef<ReturnType<typeof setTimeout>>();
 
   React.useEffect(() => {
     if (toast) {
@@ -45,12 +46,17 @@ export function Toast({ toast, onClose }: ToastProps) {
     }
   }, [toast]);
 
+  // Clear the fade-out timer if the component unmounts while it's in flight
+  React.useEffect(() => {
+    return () => clearTimeout(closeTimerRef.current);
+  }, []);
+
   const handleClose = () => {
     // Trigger fade-out animation
     setIsExiting(true);
 
     // Wait for animation to complete before removing
-    setTimeout(() => {
+    closeTimerRef.current = setTimeout(() => {
       setIsVisible(false);
       setIsExiting(false);
       onClose();


KhyFee (commented):
﻿Root cause + fix match the Chart.tsx pattern: stash the fade-out `setTimeout` id and clear on unmount so `onClose` / setState can’t run after teardown. LGTM on the behavior.

Nits (non-blocking):
- `handleClose` lost two levels of indentation (function body at class level indent) — please restore prettier/project indent so this doesn’t look accidental in blame.
- File lost the trailing newline (`\ No newline at end of file`).

Otherwise good merge candidate for #1016.

skoeva (changes requested):
hi, could you adjust your commit title based on our [commit guidelines](https://headlamp.dev/docs/latest/contributing#2-follow-commit-guidelines)?

Mansi2007275:
Hi @skoeva, I've updated the commit title to "strimzi: fix fade-out timer not cleared on unmount" — please let me know if this matches your commit guidelines or if you'd like a different format.

@KhyFee, I've also addressed your indentation and trailing newline nits in the latest push.

Bhargava-Ram-Thunga:
Hi @Mansi2007275,

I'm reviewing this PR! Storing the fade-out animation timer in `closeTimerRef` and clearing it in `useEffect` cleanup prevents potential `setState` calls on unmounted Toast components.

I'll also review and test this further to support your PR. Great job!

Bhargava-Ram-Thunga (commented):
Hi @Mansi2007275,

I'm reviewing this PR! I verified this one hands-on: I checked out the branch, ran `tsc`/`lint`/tests (71 passing), and wrote a fake-timer React Testing Library test that unmounts the Toast during the 300ms fade window. On the base code, the fade timeout survives unmount and `onClose` fires against the unmounted component — the exact bug in #1016. On this branch, the pending timer count drops to zero on unmount and `onClose` never fires, while a toast that stays mounted still closes exactly once. The fix works, and `React.useRef<ReturnType<typeof setTimeout>>()` is the right portable typing.

Two non-blocking suggestions:

1. Consider adding a test like the above to the PR — the repo already has vitest + RTL set up in `strimzi`, and this bug class keeps recurring (#978, #979, #1016), so a regression test would pay for itself.
2. Related pre-existing edge (not blocking): if a new toast arrives while a close animation is in flight, the pending `closeTimerRef` timeout still fires ~300ms later and hides the new toast / calls `onClose`. Clearing `closeTimerRef.current` at the top of the `[toast]` effect when a new toast shows would close that race too.

Also note the outstanding "changes requested" on this PR is about the commit title format per the contributing guidelines — worth fixing that so this can move.

PR #1024, opened 2026-08-05
cluster-api: add accessible names to filter selects and copy button

omlahore:
Four Select controls in the cluster-api overview render with no accessible name. They sit directly inside a `Box` with no `FormControl`, `InputLabel`, `label` or `aria-label`, so assistive technology announces them as unlabeled comboboxes and cannot tell the cluster, resource and severity filters apart.

- `Overview.tsx`: cluster, resource and severity filters under "Conditions"
- `ClusterDetailsErrorOverview.tsx`: cluster filter under "Clusters"

The copy button in the inline solution panel has the same problem. Its only child is an `Icon`, so it exposes no name. Its label now also reflects the copied state, which was previously conveyed only by the icon swapping to a checkmark.

Uses `aria-label` via `inputProps` on the selects, following the pattern in prometheus (#999). `lint`, `tsc --noEmit` and prettier are clean.

Found by reading the source, not by an automated audit. I do not have a Cluster API cluster to run the plugin against, so these have not been confirmed against axe in a browser.

### DIFF
--- a/cluster-api/src/components/overview/ClusterDetailsErrorOverview.tsx
+++ b/cluster-api/src/components/overview/ClusterDetailsErrorOverview.tsx
@@ -186,7 +186,15 @@ function InlineSolutionPanel({
                   <Typography variant="caption" color="text.secondary">
                     {cmd.description}
                   </Typography>
-                  <IconButton size="small" onClick={() => handleCopy(cmd.command)}>
+                  <IconButton
+                    size="small"
+                    aria-label={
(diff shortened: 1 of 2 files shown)


omlahore:
@skoeva could you take a look when you get a chance? Checks are green and the Copilot review you kicked off came back clean. It is a small one, accessible names on four unlabeled `Select` filters and the icon-only copy button in cluster-api.

PR #1032, opened 2026-08-06
feat(kyverno): support openreports.io/v1alpha1 reports

### LINKED ISSUE #994: kyverno: Add support for the openreports.io reporting group

Pyush-10:
### Description 
Fixes #994 
This PR adds support for OpenReports (`openreports.io/v1alpha1`) alongside the legacy `wgpolicyk8s.io/v1alpha2` reports in the Kyverno Headlamp plugin. 

- Added `Report` and `ClusterReport` resource classes.
- Updated capability checks to allow the reports page, compliance badges, and dashboard to function if either API group is active.
- Prevented stuck-loading states by replacing uninstalled API groups' responses with empty arrays rather than blocking on `null` list responses.
- Verified and merged report/violation counts in all widgets, dashboards, details, and list components.

### Verification
- Verified compilation: `npm run build` bundles successfully.
- Verified TypeScript checks: `npx tsc --noEmit` runs with 0 errors.
- Verified lint checks: `npx eslint` passes with 0 warnings/errors.

### DIFF
--- a/knative/src/components/common/ReadyStatusLabel.stories.tsx
+++ b/knative/src/components/common/ReadyStatusLabel.stories.tsx
@@ -56,3 +56,10 @@ WithMessage.args = {
   reason: 'RevisionFailed',
   message: 'Initial scale was never achieved',
 };
+
+export const ScaledToZero = Template.bind({});
+ScaledToZero.args = {
+  status: 'True',
+  actualReplicas: 0,
+};
+
--- a/knative/src/components/common/ReadyStatusLabel.tsx
+++ b/knative/src/components/common/ReadyStatusLabel.tsx
@@ -22,6 +22,7 @@ interface ReadyStatusLabelProps {
   reason?: string;
   message?: string;
   isReadyType?: boolean;
+  actualReplicas?: number;
 }
 
 export function ReadyStatusLabel({
@@ -29,13 +30,19 @@ export function ReadyStatusLabel({
   reason,
   message,
   isReadyType = true,
+  actualReplicas,
 }: ReadyStatusLabelProps) {
   let headlampStatus: 'success' | 'error' | 'warning' | '' = '';
   let labelText = 'Unknown';
 
   if (status === 'True') {
-    headlampStatus = 'success';
-    labelText = isReadyType ? 'Ready' : 'True';
+    if (actualReplicas === 0) {
+      headlampStatus = '';
+      labelText = isReadyType ? 'Scaled to Zero' : 'True';
+    } else {
+      headlampStatus = 'success';
+      labelText = isReadyType ? 'Ready' : 'True';
+    }
   } else if (status === 'False') {
     headlampStatus = 'error';
     labelText = isReadyType ? 'Not Ready' : 'False';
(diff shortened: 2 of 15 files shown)


skoeva (commented):
this is a kyverno PR but it touches the knative plugin code - we can close this in favor of #1009

PR #1038, opened 2026-08-06
app-catalog: prevent stale chart data from race condition in ChartDet…

### LINKED ISSUE #1037: app-catalog: ChartDetails useEffect has no cleanup/cancellation — race condition sets stale chart data
### What happens
In `app-catalog/src/components/charts/Details.tsx`, the data-fetching
`useEffect` (lines 49–58) calls `fetchChartDetailFromArtifact(chartName, repoName).then(setChart)`
with no cancellation guard.

If `chartName`/`repoName` change again before the in-flight request resolves
(e.g. user quickly clicks through multiple charts in the catalog list), an
older request can resolve *after* a newer one and overwrite `chart` state
with data for the wrong chart. There's also no unmount check, so a
setState-after-unmount warning is possible if the user navigates away
before the fetch completes.

### Fix
Add a `let ignore = false` flag (or AbortController) and cleanup:

```tsx
useEffect(() => {
  let ignore = false;
  fetchChartDetailFromArtifact(chartName, repoName).then(response => {
    if (!ignore) setChart(response);
  });
  return () => { ignore = true; };
}, [chartName, repoName]);
```

### To reproduce
1. Open App Catalog
2. Rapidly click between two different charts before either detail page
   finishes loading
3. Observe the displayed chart details can briefly (or persistently)
   mismatch the URL/chartName

Mansi2007275:
Fixes #<issue-number>

## What was broken
`ChartDetails`'s data-fetching useEffect had no cancellation guard.
If a user navigated quickly between charts, an older network response
could resolve after a newer one and overwrite state with the wrong
chart's data.

## Fix
Added an `ignore` flag that's set to `true` in the effect's cleanup,
so a stale response is ignored instead of overwriting state.

## Testing
- `npx tsc --noEmit` — clean, no errors
- `npx eslint` on the changed file — clean, no errors
- Manually verified by rapidly switching between two charts in the
  catalog list — details now stay consistent with the selected chart

### DIFF
--- a/app-catalog/src/components/charts/Details.tsx
+++ b/app-catalog/src/components/charts/Details.tsx
@@ -46,6 +46,18 @@ export default function ChartDetails({ vanillaHelmRepo }: ChartDetailsProps) {
   const [openEditor, setOpenEditor] = useState(false);
   const chartCfg = getCatalogConfig();
 
+  // useEffect(() => {
+  //   // Note: This path is not enabled for vanilla helm repo. Please check the following comment in charts/List.tsx
+  //   // TODO: The app-catalog using artifacthub.io loads the details about the chart with an option to install the chart
+  //   //
+  //   // An API to get details about a particular chart is required to achieve this. For example, take a look at the response
+  //   // from https://artifacthub.io/api/v1/packages/helm/grafana/grafana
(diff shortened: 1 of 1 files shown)


skoeva (commented):
#1054 is the fuller PR, closing this in favor of that one

PR #1040, opened 2026-08-06
flux: Add controller logs and terminal actions to overview

DhruvShetty22:
Adds an **Actions** column to the Controllers table on the Flux Overview, with a **Logs** and a **Terminal** button per controller.

The table rows are Deployments, but logs and exec need a Pod, so each row resolves its pods through `spec.selector.matchLabels`. Pods are listedonce for the whole table rather than per row.

I manually tested on the following:
- **Multiple pods** : a Pod selector appears in the log viewer
- **Multiple containers** — a Container selector appears

Also verified: logs stream and follow; terminal opens a working shell.

Added 9 unit tests for the pod-matching logic. `build`, `tsc`, `lint`,
`test` and `format --check` all pass locally.

## Screencast

https://github.com/user-attachments/assets/fbadaebb-4c1f-44b4-af4d-82597518e17d


DhruvShetty22:
This is my first time contributing to Headlamp... please let me know if i need to follow some rules incase i missed them.
( I have been contributing to zulip for past 6 months and i am currently part of the GSOC mentee for same)

@illume Tagging you for a review.

PR #1052, opened 2026-08-07
fix(kyverno): make NotInstalledBanner accessible

Goyamjain06:
### Summary

`NotInstalledBanner`'s loading and not-installed states gave assistive
technology nothing to work with — an unlabeled spinner and a message
that wasn't announced when it appeared.

### Why

Screen reader users got silence through both states: no accessible name
on the loading spinner, and no announcement when the "not detected"
message rendered.

### What changed

- Loading state now uses `Loader` with a `title`, giving the spinner an
  accessible name.
- Not-installed state is now a `role="status" aria-live="polite"
  aria-atomic="true"` region, so the message is announced when it
  appears.
- No visual or layout changes.
- Added focused unit tests covering both states.

### Testing

- `prettier --check` — clean
- `eslint --max-warnings 0` — clean
- `tsc --noEmit` — clean
- `vitest run` — 7/7 passing (5 pre-existing, 2 new)
- Manually verified both states render the expected ARIA attributes.

### DIFF
--- a/kyverno/src/components/common.test.tsx
+++ b/kyverno/src/components/common.test.tsx
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


Goyamjain06:
Hi @illume , just following up on this PR when you get a chance. The change is intentionally scoped to NotInstalledBanner accessibility, with focused tests and all local checks passing. Would appreciate a review whenever convenient. Thanks!

Goyamjain06:
Hi @illume, just a gentle bump on this one — no rush at all. All checks are still passing and it's ready for review whenever you have some time. Thanks!

PR #1056, opened 2026-08-07
flux: fix crashes when spec.force or spec.chart are missing

needhieshsingxh:
### Summary

Fixes runtime crashes in Flux plugin detail views when `spec` or `spec.force` are omitted:

1. **KustomizationSingle.tsx**: Accessing `force.toString()` crashes with a `TypeError` when `spec.force` is omitted in the resource manifest. Added optional chaining (`force?.toString()`) with a fallback to `'false'`.
2. **HelmReleaseSingle.tsx**: When `jsonData.spec` is absent (`jsonData: {}`), accessing `cr?.jsonData?.spec.chart` throws a `TypeError` on `.chart` because the dot after `spec` is unguarded. Added optional chaining for `spec?.chart` across `reconcileStrategy` and `version` fields.

### Steps to test

1. Open a Kustomization detail view for a resource without `spec.force` set.
2. Verify the page renders cleanly with `Force: false` instead of throwing a `TypeError` crash.
3. Open a HelmRelease detail view for a resource without a `spec` block (or empty `jsonData`).
4. Verify the page renders without crashing on `reconcileStrategy` or `version`.

### DIFF
--- a/flux/src/helm-releases/HelmReleaseSingle.tsx
+++ b/flux/src/helm-releases/HelmReleaseSingle.tsx
@@ -126,7 +126,7 @@ function CustomResourceDetails(props) {
       },
       {
         name: t('Reconcile Strategy'),
-        value: cr?.jsonData?.spec.chart?.spec?.reconcileStrategy,
+        value: cr?.jsonData?.spec?.chart?.spec?.reconcileStrategy,
       },
     ];
 
--- a/flux/src/helm-releases/HelmReleaseSingle.tsx
+++ b/flux/src/helm-releases/HelmReleaseSingle.tsx
@@ -167,11 +167,11 @@ function CustomResourceDetails(props) {
     }
     extraInfo.push({
       name: t('Version'),
-      value: cr?.jsonData?.spec.chart?.spec?.version,
+      value: cr?.jsonData?.spec?.chart?.spec?.version,
     });
     extraInfo.push({
       name: t('Suspend'),
-      value: cr?.jsonData.spec?.suspend ? t('True') : t('False'),
+      value: cr?.jsonData?.spec?.suspend ? t('True') : t('False'),
     });
 
     const interval = cr?.jsonData.spec?.interval;
--- a/flux/src/helm-releases/HelmReleaseSingle.tsx
+++ b/flux/src/helm-releases/HelmReleaseSingle.tsx
@@ -174,13 +174,13 @@ function CustomResourceDetails(props) {
       value: cr?.jsonData?.spec?.suspend ? t('True') : t('False'),
     });
 
-    const interval = cr?.jsonData.spec?.interval;
+    const interval = cr?.jsonData?.spec?.interval;
     extraInfo.push({
       name: t('Interval'),
       value: interval,
     });
 
-    if (!cr?.jsonData.spec?.suspend) {
+    if (!cr?.jsonData?.spec?.suspend) {
       extraInfo.push({
         name: t('Next Reconciliation'),
         value: <RemainingTimeDisplay item={cr} />,
--- a/flux/src/kustomizations/KustomizationSingle.tsx
+++ b/flux/src/kustomizations/KustomizationSingle.tsx
@@ -60,11 +60,11 @@ function KustomizationDetails(props) {
       },
       {
         name: t('Path'),
-        value: cr?.jsonData.spec?.path,
+        value: cr?.jsonData?.spec?.path,
       },
       {
         name: t('Prune'),
-        value: cr?.jsonData.spec?.prune,
+        value: cr?.jsonData?.spec?.prune,
       },
       {
         name: t('SourceRef'),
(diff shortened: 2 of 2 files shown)


Ralthos (commented):
Both guards look right to me, and the `?? 'false'` reads correctly since Flux treats an omitted `spec.force` as false.

Two things came out of checking the chains.

The same unguarded access appears a second time in `HelmReleaseSingle.tsx`, about forty lines below the one this fixes:

```ts
// line 129, fixed here
value: cr?.jsonData?.spec.chart?.spec?.reconcileStrategy,

// line 170, unchanged
value: cr?.jsonData?.spec.chart?.spec?.version,
```

Both are `?.spec.chart`, with the dot after `spec` unguarded, so whatever crashes the first one crashes the second. Worth folding into the same PR.

On the cause: I think the description has it slightly off, which matters because it changes what to test. `spec.chartRef` in place of `spec.chart` does not crash. The throw needs `spec` itself to be absent:

```js
const cr = { jsonData: { spec: { chartRef: {} } } };
cr?.jsonData?.spec.chart?.spec?.version   // undefined, no throw

const cr2 = { jsonData: {} };
cr2?.jsonData?.spec.chart?.spec?.version  // TypeError
```

`spec.chart` evaluates to undefined when `chart` is missing, and the `?.` after it absorbs that. The optional chain only breaks where a plain dot follows something undefined, which here is `spec`.

So the fix is right. The reasoning behind it is the part I would change. If you have a manifest that reproduced this, it would be useful in the PR body, since a HelmRelease with no `spec` at all is the interesting case and is easy to miss when testing.

needhieshsingxh:
> Both guards look right to me, and the `?? 'false'` reads correctly since Flux treats an omitted `spec.force` as false.
> 
> Two things came out of checking the chains.
> 
> The same unguarded access appears a second time in `HelmReleaseSingle.tsx`, about forty lines below the one this fixes:
> 
> ```ts
> // line 129, fixed here
> value: cr?.jsonData?.spec.chart?.spec?.reconcileStrategy,
> 
> // line 170, unchanged
> value: cr?.jsonData?.spec.chart?.spec?.version,
> ```
> 
> Both are `?.spec.chart`, with the dot after `spec` unguarded, so whatever crashes the first one crashes the second. Worth folding into the same PR.
> 
> On the cause: I think the description has it slightly off, which matters because it changes what to test. `spec.chartRef` in place of `spec.chart` does not crash. The throw needs `spec` itself to be absent:
> 
> ```js
> const cr = { jsonData: { spec: { chartRef: {} } } };
> cr?.jsonData?.spec.chart?.spec?.version   // undefined, no throw
> 
> const cr2 = { jsonData: {} };
> cr2?.jsonData?.spec.chart?.spec?.version  // TypeError
> ```
> 
> `spec.chart` evaluates to undefined when `chart` is missing, and the `?.` after it absorbs that. The optional chain only breaks where a plain dot follows something undefined, which here is `spec`.
> 
> So the fix is right. The reasoning behind it is the part I would change. If you have a manifest that reproduced this, it would be useful in the PR body, since a HelmRelease with no `spec` at all is the interesting case and is easy to miss when testing.

Thanks for the review @Ralthos! 🙌

1. **Second unguarded access fixed:** Added optional chaining to `cr?.jsonData?.spec?.chart?.spec?.version` on line 170 of `HelmReleaseSingle.tsx`.
2. **Audit of remaining spec accesses:** Swept both `HelmReleaseSingle.tsx` and `KustomizationSingle.tsx` to ensure all `spec` accesses use optional chaining (`jsonData?.spec?.`).
3. **PR Description updated:** Updated the summary to clarify that an omitted `spec` (`jsonData: {}`) causes the plain dot after `spec` to throw a `TypeError`.

Pushed the update to this PR!

Ralthos (commented):
Both files are clean at this head. Every chain in `KustomizationSingle.tsx` and `HelmReleaseSingle.tsx` is guarded now, including the `spec.chart` pair on Reconcile Strategy and Version that I pointed at before. Nothing left in either file.

One for a follow-up, since it sits outside what this PR set out to do.

`ImageAutomationSingle.tsx` has the same shape in the `ImagePolicy` branch, and it is inconsistent inside a single element:

```tsx
name: resource.jsonData.spec?.imageRepositoryRef.name,
namespace:
  resource.jsonData.spec.imageRepositoryRef.namespace ??
  resource.jsonData.metadata.namespace,
```

`spec` is optional-chained on the first line and bare three lines later, so the guard on one does nothing for the other.

I have not checked whether an ImagePolicy can reach that view without `imageRepositoryRef`, so I do not know if it is reachable in practice or only untidy. Your case here is stronger either way, since `spec.force` is omitted in ordinary manifests and that crash lands on valid resources.

Different file and a different view, so I would keep it out of this PR.

needhieshsingxh:
> Both files are clean at this head. Every chain in `KustomizationSingle.tsx` and `HelmReleaseSingle.tsx` is guarded now, including the `spec.chart` pair on Reconcile Strategy and Version that I pointed at before. Nothing left in either file.
> 
> One for a follow-up, since it sits outside what this PR set out to do.
> 
> `ImageAutomationSingle.tsx` has the same shape in the `ImagePolicy` branch, and it is inconsistent inside a single element:
> 
> ```tsx
> name: resource.jsonData.spec?.imageRepositoryRef.name,
> namespace:
>   resource.jsonData.spec.imageRepositoryRef.namespace ??
>   resource.jsonData.metadata.namespace,
> ```
> 
> `spec` is optional-chained on the first line and bare three lines later, so the guard on one does nothing for the other.
> 
> I have not checked whether an ImagePolicy can reach that view without `imageRepositoryRef`, so I do not know if it is reachable in practice or only untidy. Your case here is stronger either way, since `spec.force` is omitted in ordinary manifests and that crash lands on valid resources.
> 
> Different file and a different view, so I would keep it out of this PR.

Thanks for the review and approval, @Ralthos! This PR is ready for merge.

I've opened the follow-up PR #1132 to add optional chaining to `ImageAutomationSingle.tsx` as suggested.  Please review it whenever you are free!

Issue #922, opened 2026-07-22
WIP: knative eventing and other improvements


illume:
Draft. 
Please do not start this issue.

## Description
Knative brings serverless and event‑driven capabilities to Kubernetes, but operators often lack a clear, unified UI for understanding how Knative resources behave at runtime. Headlamp’s existing Knative plugin provides partial support for Knative Serving, but Eventing resources, metrics, accessibility, and internationalization remain incomplete. 

This mentorship focuses on expanding and polishing Knative support within Headlamp. The mentee will work with React/TypeScript, Kubernetes APIs, and Knative CRDs to build intuitive visualizations, improve resource detail pages, integrate Prometheus metrics, and complete foundational UX work such as accessibility and i18n coverage. 

## Expected outcomes
By the end of the term, the mentee will deliver a set of concrete improvements to the Knative plugin for Headlamp, including:

- Some improvements to existing Knative Service
- Knative Eventing UI support: List and detail views for Brokers, Triggers, Channels, Subscriptions, and Sources, with linked relationships and status indicators. Forms for creating common Eventing resources.
- Maps visualisation of Knative Eventing resources.
- Prometheus metrics integration: Embed Knative‑Events specific charts (e.g., admission request rate, revision performance) following Headlamp’s existing Prometheus plugin conventions.
- Complete i18n coverage: Audit all components, wrap hardcoded strings with `t()`, finalize the English locale file, and scaffold empty locale files for future translations.
- Accessibility (a11y) improvements: Add ARIA labels, ensure keyboard navigation, provide screen‑reader‑friendly chart descriptions, adopt color‑blind‑friendly palettes, and add visible focus indicators.
- Documentation: Update the plugin README, add developer notes, and provide user‑facing documentation for new features.
- Blog post: on Kubernetes blog describing outcome of project.

illume:
# Add a title
[CNCF LFX Proposal] <project> <program name>

# CNCF Project
*
The CNCF project this program belongs to. If your project isn't listed, check that it appears in the CNCF Landscape and ask in #cncf-mentoring on CNCF Slack.

Headlamp

# Term
*
Which term is this proposal for? All programs are full-time (~350 hours across the term).

2026 Term 3 (Sept - Nov)


# Program Name

The descriptive part of the program name shown on LFX. The full name is composed automatically as CNCF - <Project>: <your program name here> (<term>). Keep it concise — available characters depend on your project name (typically 53-73 chars), the full program name must be < 100 characters.







[...]

PR #1066, opened 2026-08-08
knative: Add i18n support

Eyasdm:
## Summary

Adds complete English i18n coverage to the Knative plugin. The plugin declared i18n support in package.json ("i18n" script and headlamp.i18n config) but had zero translation coverage — every user-facing string was hardcoded.

Follows the pattern established by the flux plugin and the recent i18n work across ai-assistant (#941), keda (#940), prometheus (#963), and cert-manager (#964) by @skoeva.

## Scope

- Files changed: 36
- Strings wrapped: 260
- Locale file: locales/en/translation.json (260 keys, English only)

### Directories covered

- common/: ConditionsSection, NotInstalledBanner, ReadyStatusLabel, Activity
- kservices/: List, Detail, Glance, header actions, log viewer, pods, traffic splitting, networking, autoscaling, domain mapping
- revisions/: List, Detail, Glance, delete/view-yaml headers, useRevisionActions hook, ContainerSection
- domainmappings/: List, Glance, DomainMappingRowAction
- clusterdomainclaims/: List, Glance
- networking/: Overview
- mapView.tsx: all detail components

## What was NOT wrapped (and why)

- Sidebar registration labels in index.tsx — module scope, outside React components. Matches how flux handles registration-time strings.
- formatIngressClass() in config/ingress.ts — utility function, not a hook context. Left for a follow-up that threads t through the call site.
- Kubernetes API comparison values (`'True'`, 'False', resource kinds, apiVersions, annotation keys) — machine-consumed, not user-facing.
- Console output, test files, stories — not rendered.
- getAge() in utils/time.ts returns 5m / 2h / 3d — deliberately left untranslated. This is the Kubernetes duration convention that kubectl prints untranslated in every locale; translating it here would make Headlamp inconsistent with the CLI users read alongside it.

## Gap found after opening

utils/domain.ts throws 6 user-facing error strings — form validation feedback and API failure messages — that this PR did not wrap, even though both call sites (`components/domainmappings/List.tsx` and components/kservices/detail/sections/networking/DomainMappingSection.tsx`) *are* translated here. It's a non-React module, so `t needs threading from the call sites in the same way this PR already does for getContainerDisplayStatus and makePodStatusLabel. Flagging it rather than leaving it to be found in review — I'll fix it as part of the split offered above.

## Implementation notes

- Standalone helpers that need translated strings receive t as a parameter from the calling component (e.g. getContainerDisplayStatus and makePodStatusLabel in PodsSection.tsx) rather than calling useTranslation() outside a React context.
- Loop variables named t were renamed to avoid shadowing the translation function (`item`, tag, etc.).
- 3 keys were added manually to the locale file where the i18next parser could not extract them from interpolated expressions.

## Note on PR #1057

This branch is based on main and does not include my open PR #1057 (a11y fix in PodsSection.tsx). There is a small expected merge conflict in that file. Either merge order works.

## Checks

- [x] npm run tsc
- [x] npm run lint
- [x] npm run build
- [x] npm run format -- --check
- [x] tests: 22/22 (4 files)

Part of the i18n work tracked in #922.


Eyasdm:
This is large for a first contribution to this plugin — 36 files in one diff is more than I'd want to ask anyone to review in one sitting. Happy to split it. My default would be by directory, opening one at a time and waiting for each to merge before the next. Would that work, or is there a grouping that fits how you review better?

soumadipcodeEnthusiast:
Nice work on this @Eyasdm, that's a lot of ground covered cleanly for a first PR here.

curious how you arrived at this approach — did you go through flux and the other recent i18n PRs (ai-assistant, keda, prometheus, cert-manager) first to figure out the conventions, or did you work it out fresh and then check it lined up after?

also, how'd you catch the utils/domain.ts gap after opening the PR — was that from re-reading your own diff, or did something in testing surface it?

Eyasdm:
@soumadipcodeEnthusiast: I read `flux` first, then checked `@skoeva`'s PRs (`#941`, `#940`, `#963`, `#964`) to confirm the convention held across all four before committing to it.

The `domain.ts` gap surfaced from tracing call sites after opening, not from re-reading the diff — `domain.ts` is a standalone helper outside the component tree, so it takes `t` as a parameter instead of `useTranslation()`, and the initial sweep only caught strings inside components. `List.tsx` and `DomainMappingSection.tsx` are both translated here, so their submit-time errors shouldn't be the lone untranslated thing in that path.

Issue #972, opened 2026-07-31
Flux sidebar install detection treats CRD list failures as operator absence


pratik340:
## Summary

While reviewing the Flux sidebar install detection introduced in #580, I noticed that any failure during the CRD discovery probe currently results in:

```ts
fluxInstalledByCluster[cluster] = false;
```

The sidebar filter then hides Flux child entries whenever `fluxInstalledByCluster[cluster] === false`.

As a result, the same behavior is applied to both:

- Flux CRDs are genuinely absent
- CRD listing is denied (e.g. RBAC / 403)
- Other probe failures (transient API errors, network issues, etc.)

## Context

My understanding from #580 is that the goal was to hide Flux sidebar entries when Flux is not installed on the cluster.

While tracing the history of this behavior, I found discussion around hiding entries when Flux is absent, but I could not find any discussion around whether probe failures should also be treated as operator absence.

## Question

Was it intentional for all CRD probe failures to be interpreted as "Flux is not installed"?

Or should sidebar entries only be hidden when a successful CRD probe determines that Flux CRDs are absent?

I wanted to clarify the intended behavior before proposing any changes.

illume:
@pratik340 thanks for the report.

I think it's this one:
> sidebar entries only be hidden when a successful CRD probe determines that Flux CRDs are absent?

If flux is there, it should show in the sidebar.

Nahid-NHB:
@illume can u assign me this

Ralthos:
Following the thread from the Flux side into the other plugins, I measured what the probes return, because the answer depends on which endpoint a plugin hits. Cluster was kind v1.36.1, using a ServiceAccount with read on pods/deployments/namespaces and no rules for the operator's API group or for `apiextensions.k8s.io`.

**CRD list/get is RBAC-gated.** This is the case the issue describes.

```
$ kubectl get --raw '/apis/apiextensions.k8s.io/v1/customresourcedefinitions/backups.velero.io' \
    --as=system:serviceaccount:default:limited
Error from server (Forbidden): customresourcedefinitions.apiextensions.k8s.io "backups.velero.io"
is forbidden: User "system:serviceaccount:default:limited" cannot get resource ...
```

Flux, Argo CD, kro and Kyverno all probe this way, so the three outcomes in the issue body are all reachable for them.

**API-group discovery is not RBAC-gated.**

```
$ kubectl get --raw '/apis/velero.io/v1' --as=system:serviceaccount:default:limited
[...]

PR #1077, opened 2026-08-08
knative: distinguish CRD probe errors from absence

Elvand-Lie:
## Summary

Only classify a confirmed missing Knative CRD as evidence that Knative is not
installed.

## Problem

The installation probe previously treated every CRD lookup failure as absence.

That included:

- RBAC denials
- transient server failures
- callback errors without an HTTP status

As a result, Headlamp could display "Knative was not detected" or hide Knative
navigation when the probe had not actually established that the CRD was absent.

## Fix

The probe now treats only a confirmed HTTP 404 as absence.

Other API error callbacks are treated as non-confirmed absence. Normal Knative
resource requests still enforce the user's Kubernetes permissions and can surface
their own API errors.

The existing request-promise fallback, boolean API, sidebar TTL, routing,
multi-cluster aggregation, and RBAC behavior are unchanged.

## Regression tests

Added coverage for:

- successful CRD lookup
- confirmed 404 absence
- forbidden response
- server error
- error callback without HTTP status
- null error callback
- empty cluster selection
- cluster option forwarding
- multi-cluster success
- partial multi-cluster absence
- settle-once behavior

## Verification

- `npm run test`
- `npm run tsc`
- `npm run lint`
- `npm run format -- --check`
- `npm run build`
- `git diff --check`

## Related work

PR #1075 adds tests for the helper's current behavior and touches the same new
test-file path. This PR additionally changes the production behavior so that
probe errors are not represented as confirmed absence.

The distinction follows the same error-versus-absence principle discussed for
the Flux installation probe in #972/#984.

## Scope

No sidebar architecture, cache TTL, routes, permissions, or installation UI was
changed.

## AI assistance

AI-assisted tooling was used during investigation and implementation. I reviewed
the final diff and ran the listed validation locally.

### DIFF
--- a/knative/src/isKnativeInstalled.test.ts
+++ b/knative/src/isKnativeInstalled.test.ts
@@ -103,14 +103,6 @@ describe('isKnativeInstalled', () => {
     expect(result).toBe(true);
   });
 
-  // This exercises the .catch() branch: the request function itself rejects
-  // rather than invoking the error callback.
-  it('does not classify a rejected request as Knative absence', async () => {
-    mockApiGet.mockImplementation(() => () => Promise.reject(new Error('Request setup failed')));
-
-    await expect(isKnativeInstalled(['cluster-a'])).resolves.toBe(true);
-  });
-
   it('returns false for empty clusters', async () => {
     const result = await isKnativeInstalled([]);
     expect(result).toBe(false);
--- a/knative/src/isKnativeInstalled.test.ts
+++ b/knative/src/isKnativeInstalled.test.ts
@@ -1,0 +1,176 @@
+/*
+ * Copyright 2025 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Capture the callbacks passed to apiGet so tests can trigger them.
+let onSuccess: (() => void) | undefined;
+let onError: ((err: any) => void) | undefined;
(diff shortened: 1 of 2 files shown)


Ralthos (commented):
**The `.catch` branch cannot fire here.**

`CustomResourceDefinition` declares one `apiVersion`, so `apiFactory` takes the `singleApiFactory` path and `get` is `streamResult`. `streamResult` kicks off an unawaited `run()` and returns `Promise.resolve(cancel)`, and every failure inside `run()` is caught there and handed to `errCb`. The promise that `request()` returns never rejects. Traced through `KubeObject.js:307`, `factories.js:115` and `:176`, and `streamingApi.js:41`, in `@kinvolk/headlamp-plugin` 0.14.0.

So flipping `settle(false)` to `settle(true)` in the `.catch` is a no-op for this resource class, and the whole behaviour change rests on `err => settle(err?.status !== 404)`. The change surface is one line.

The test `does not classify a rejected request as Knative absence` mocks `apiGet` into returning a rejecting function, a shape the real `apiGet` does not produce. It passes, and it pins a branch production cannot reach. Harmless, but it reads as coverage of the `.catch` decision when nothing is being covered.

**For RBAC-restricted users the probe now returns `true` unconditionally.**

I measured this while writing up #1083. A ServiceAccount without `get` on `customresourcedefinitions` gets 403 from `/apis/apiextensions.k8s.io/v1/customresourcedefinitions/<name>`, and authorization is checked before existence, so 403 comes back whether the CRD is there or not.

That user can no longer reach the absence branch, which is the correct failure direction and does fix the bug in the title. It also means the probe has stopped answering the question for them: `true` regardless of what the cluster serves. `/apis/serving.knative.dev/v1` would answer it, since group discovery is not RBAC gated and `system:discovery` is bound to `system:authenticated`.

I would leave that alone here. It is the case #1083 argues, and the boolean return is why the choice had to be made at all. A third state would let the sidebar show Knative and let the pages surface their own errors, without encoding "could not tell" as "installed".

One more, and it predates this PR: `hasCrdInCluster` leaks a watch connection whenever the probe succeeds.

`settle(true)` runs synchronously inside `cb(item)` and calls `cancelFn()`, which sets `isCancelled` and cancels `socket`. `socket` is undefined at that moment. `run()` then resumes on the next statement and opens the watch:

```js
cb(item);
const watchUrl = ...;
socket = stream(watchUrl, (x) => cb(x.object), { isJson: true, cluster: clusterName });
```

The `isCancelled` guard stops `cancel()` from running a second time, so that socket stays open. One per cluster on every successful probe, which is the common path. Probably belongs in its own issue.

PR #1087, opened 2026-08-09
opencost: Guard isOpenCostInstalled against a missing http-ui port

### LINKED ISSUE #1086: [Bug]opencost: isOpenCostInstalled throws when the OpenCost service has no port named "http-ui", silently breaking the Cost section
## What happens

`isOpenCostInstalled` in `opencost/src/request.tsx` finds the OpenCost Service by label
(`app.kubernetes.io/name=opencost`), then assumes it always exposes a port literally named
`http-ui`:

```ts
const httpUIPort = response.items[0].spec.ports.filter(port => port.name === 'http-ui');
return [
  true,
  `${response.items[0].metadata.name}:${httpUIPort[0].name}`,
  response.items[0].metadata.namespace,
];
```

`.filter()` returns `[]` when nothing matches, so `httpUIPort[0]` is `undefined`, and
`.name` throws `TypeError: Cannot read properties of undefined (reading 'name')`. This
happens whenever the matching Service doesn't expose a port named exactly `http-ui` —
different OpenCost Helm chart versions and custom installs commonly name it differently.

## Why it matters

`isOpenCostInstalled` is only ever awaited inside `useEffect`s with no try/catch
(`detail.tsx` and `index.tsx`), so the throw becomes an unhandled promise rejection.
Downstream state never gets corrected:

- `detail.tsx`: `installed` state starts as `true` and is never updated on failure, so the
  component believes OpenCost is installed and gets stuck permanently rendering
  `<Loader title="Fetching Data" />` instead of the "Install Opencost" fallback.
- `index.tsx`'s list-view cost column: `isLoading` stays `true` forever for the same
  reason, so every row shows `...` indefinitely instead of `N/A` or a real cost.

## Where

- `opencost/src/request.tsx` (`isOpenCostInstalled`, ~line 16-19)

## Steps to Reproduce

1. Install OpenCost with a Service that matches `app.kubernetes.io/name=opencost` but
   exposes its port under a name other than `http-ui` (reproduced directly by unit test:
   mock the API response with `spec.ports: [{ name: 'http', port: 9003 }]` and call
   `isOpenCostInstalled()` — it rejects with the TypeError above instead of resolving).
2. In the running app, open any Pod/Node/Namespace/Service/Deployment detail page — the
   Cost section gets stuck on "Fetching Data" forever.
3. Check the browser console for the uncaught `TypeError`.

## Expected Behavior

If no port named `http-ui` is found, `isOpenCostInstalled` should resolve to
`[false, null, null]` (same as when no matching Service exists at all), so the UI falls
back to "Install Opencost for accessing cost info" instead of getting stuck.

KariraLakshya:
## Summary

Fixes `isOpenCostInstalled` crashing when the matched OpenCost Service doesn't expose a port literally named `http-ui`.

## Related Issue

Fixes #1086

## Changes

- `opencost/src/request.tsx`: `isOpenCostInstalled` used `.filter(...)[0].name`, which throws `TypeError: Cannot read properties of undefined (reading 'name')` when no port is named `http-ui` (different OpenCost Helm chart versions and custom installs commonly name it differently). Switched to `.find()` with an explicit `undefined` check, returning the same "not installed" tuple (`[false, null, null]`) used when no matching Service exists at all.
- No behavior change for the common case where the port is named `http-ui`.

## Why it matters

`isOpenCostInstalled` is only ever awaited inside `useEffect`s with no try/catch (`detail.tsx`, `index.tsx`), so the throw became an unhandled rejection. Downstream `installed`/`isLoading` state (both default to a "loading/installed" value) never got corrected, so the Cost section got stuck permanently on "Fetching Data" instead of falling back to the "Install Opencost" message.

## Steps to Test

1. Mock the Kubernetes API response for `isOpenCostInstalled` with a Service matching `app.kubernetes.io/name=opencost` but with `spec.ports: [{ name: 'http', port: 9003 }]` (no `http-ui` port).
2. Before the fix: `isOpenCostInstalled()` rejects with the TypeError above.
3. After the fix: `isOpenCostInstalled()` resolves to `[false, null, null]`, and the UI correctly falls back to "Install Opencost for accessing cost info" instead of an infinite loader — verified locally via a throwaway test against the mocked response.

### DIFF
--- a/opencost/src/request.tsx
+++ b/opencost/src/request.tsx
@@ -12,10 +12,13 @@ export async function isOpenCostInstalled() {
 
   if (response.items && response.items.length > 0) {
     // find the service with name http-ui
-    const httpUIPort = response.items[0].spec.ports.filter(port => port.name === 'http-ui');
+    const httpUIPort = response.items[0].spec.ports.find(port => port.name === 'http-ui');
+    if (!httpUIPort) {
+      return [false, null, null];
+    }
     return [
       true,
-      `${response.items[0].metadata.name}:${httpUIPort[0].name}`,
+      `${response.items[0].metadata.name}:${httpUIPort.name}`,
       response.items[0].metadata.namespace,
     ];
   }


Bhargava-Ram-Thunga:
Hi @KariraLakshya,

I'm reviewing this PR! Replacing `.filter()` with `.find()` and adding a guard check prevents the `TypeError` when the OpenCost service doesn't expose an `http-ui` port.

I'll also review and test this further to help get it reviewed by maintainers. Well done!

Bhargava-Ram-Thunga (commented):
Hi @KariraLakshya,

I'm reviewing this PR! I verified it hands-on: checked out the branch, ran `tsc`/`lint` (both pass), and exercised `isOpenCostInstalled` with a mocked `ApiProxy.request` across four scenarios. With an `http-ui` port it returns `[true, 'opencost:http-ui', 'opencost']`; with only a port named `http` it now resolves `[false, null, null]` where the old `.filter(...)[0].name` threw a TypeError; with no matching services it returns `[false, null, null]` as before. Since the function is awaited in `useEffect`s with no try/catch, killing that unhandled rejection genuinely unsticks the permanent "Fetching Data" state.

Two non-blocking suggestions:

1. My mock testing confirmed one adjacent gap remains: a matched Service whose `spec.ports` is `undefined` still throws at the `.find()` call. `response.items[0].spec?.ports?.find(...)` would harden that for free in the same line.
2. Semantics question for maintainers: when OpenCost *is* installed but the chart names the port differently (which #1086 says is common), this returns "not installed" and shows the "Install Opencost" message to someone who has it. A fallback (match by port 9090, or first HTTP port) might serve those installs better — though this change is already a strict improvement over crashing. The mocked scenarios above would also make a nice small `.test.ts` to include in the PR.

PR #1093, opened 2026-08-09
metal3: Add a fleet overview landing page

mastermaxx03:
## Summary

Adds a fleet overview as the Metal3 section's landing page, aimed at operators running many
bare-metal hosts at once.

Metal3 has no fleet resource, so hosts are grouped into fleets by a label you pick from a
dropdown, and namespace scoping uses Headlamp's shared namespace filter rather than a grouping of
its own. Each fleet is a row in a sortable table showing how many hosts it has and, of those, how
many are free, claimed, and specifically backing a Metal3Machine (the real "in use" number for
bare metal), how many are in error or currently in progress, and how many are powered on.

Below the table there are two lists across all fleets:

- In progress: hosts in a transitional state (registering, inspecting, preparing, provisioning, deprovisioning, or deleting), with how long each has been in that
  state, longest first, so a stalled host rises to the top. The duration comes from the operation
  start time in status.operationHistory, not status.lastUpdated, which re-stamps on every
  reconcile.
- Needs attention: hosts in error, delayed, or carrying an error type, error first. Detached and
  servicing are treated as informational, not problems.

Everything is pure aggregation over the BareMetalHost list, so no new data is fetched beyond what
the list view already loads, and the aggregation helpers (grouping, allocation counts,
time-in-state, attention) are unit-tested.

## Notes for reviewers

- I made the overview the section's landing page, so the Metal3 group header now lands here,
  matching the cluster-api and flux plugins. Happy to make it a plain entry instead if you prefer.
- The presentation, a sortable row per fleet plus two shared action lists, is one reading of
  "multiple fleets by label." Open to a different shape.
- The fleet view is only as useful as the labels operators put on their hosts. With no useful
  labels it shows a single "All hosts" fleet.
- The namespace selection is read with react-redux useSelector on filter.namespaces.

## How to test

1. A cluster with the Metal3 CRDs and a few BareMetalHosts.
2. cd metal3 && npm install && npm start, then run Headlamp.
3. Open Metal3 in the sidebar; it lands on the Overview.
4. Label some hosts (for example kubectl label bmh <name> rack=r1) and pick that label under
   "Group by label" to split into fleets. Click In error or In progress to sort, and use the
   namespace filter to scope.

## Media

Screenshot for the overview grouped into fleets by label (the fleets table, in progress, and
needs attention)
image

Screen recording for switching the group-by label, sorting a column, and scoping with the
namespace filter

https://github.com/user-attachments/assets/d22278a7-359a-4690-8ef0-778765df686c

### DIFF
--- a/metal3/src/BareMetalHost/Overview.tsx
+++ b/metal3/src/BareMetalHost/Overview.tsx
@@ -1,0 +1,241 @@
+/*
+ * Copyright 2026 The Kubernetes Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  EmptyContent,
+  Link,
+  Loader,
+  NamespacesAutocomplete,
+  PageGrid,
+  SectionBox,
+  SimpleTable,
+} from '@kinvolk/headlamp-plugin/lib/components/common';
+import type { KubeObject } from '@kinvolk/headlamp-plugin/lib/k8s/cluster';
+import { Box, MenuItem, TextField } from '@mui/material';
+import { useMemo, useState } from 'react';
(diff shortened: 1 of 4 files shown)


mastermaxx03:
@Rozzii @illume  Ready for Review!

Rozzii:
Looks good to me, IMO this is a useful view , we can expand on the landing page in the future but these fleet views will be needed.

mastermaxx03:
> Looks good to me, IMO this is a useful view , we can expand on the landing page in the future but these fleet views will be needed.

yeah I think this will be a good starting point for the alpha release.

Rozzii:
LGTM

illume (approved):
Thanks! 🎉

PR #371, opened 2025-08-17
karpenter: fix memory limit view

SinghaAnirban005:
## Description
The memory bar now does not fill up when limit is not specified . 

## Video

[Screencast from 2025-08-22 22-49-29.webm](https://github.com/user-attachments/assets/896f6211-a79f-4f56-93e6-3e9e4fd0e2b1)

Fixes #370 
Part of https://github.com/kubernetes-sigs/headlamp/issues/3231

PR #1095, opened 2026-08-09
karpenter: NodePool: Fix CPU bar filling when no CPU limit is set

### LINKED ISSUE #1094: karpenter: CPU bar renders 100% when a NodePool has no CPU limit
### Description

The Memory column of the NodePool list had this exact problem in #370, fixed by
#371. That fix only touched the Memory column, so the CPU column in the same
table still behaves the old way.

In `karpenter/src/NodePool/List.tsx`, the CPU column feeds `PercentageBar`:

    const used = parseInt(nodePool.jsonData.status?.resources?.cpu || 0);
    const limit = parseInt(nodePool.jsonData.spec?.limits?.cpu || 0);
    ...
    value: used,
    const effectiveLimit = limit > 0 ? limit : used || 1;
    total={effectiveLimit}

`PercentageBar` fills as `(value / total) * 100`. When `spec.limits.cpu` is
absent, `limit` is 0, so `effectiveLimit` falls back to `used` — the same value
passed as `value` — and the ratio is exactly 1. The bar paints solid.

A saturated bar reads as "at capacity", which is the opposite of what an
unlimited NodePool means. The tooltip for the same cell correctly reports
"No limit set", so the bar and its tooltip contradict each other.

### Steps to reproduce

1. Apply a NodePool with no `spec.limits.cpu`. The default NodePool in the
   Karpenter docs sets `limits.cpu: "1000"`, so remove it.
2. Let it provision at least one node, so `status.resources.cpu` is non-zero.
3. Open Karpenter > NodePools.

### Actual

The CPU bar is completely full. The Memory bar in the same row is empty for the
same "no limit" condition, so the two columns of one row disagree.

### Expected

The CPU bar behaves like the Memory bar after #371: empty when no limit is set.

### Proposed fix

Apply the same change #371 made to Memory:

    value: limit > 0 ? used : 0,
    total: limit > 0 ? limit : 1,

The `limit > 0` path is unchanged by this, so only the unlimited case is
affected. Happy to open a PR for this.

### Environment

- karpenter plugin 0.2.0
- @kinvolk/headlamp-plugin 0.14.0
- Reproduced against `main` @ cc27c20

Shashank200345:
## Description

The CPU column of the NodePool list rendered a completely full PercentageBar for
any NodePool without `spec.limits.cpu`. `PercentageBar` fills as
`(value / total) * 100`, and the CPU column passed `used` as `value` while
falling back to `used || 1` for `total`, so the ratio was exactly 1 whenever no
limit was set.

The Memory column already handles this correctly since #371, which only touched
Memory. This applies the same two expressions to CPU so both columns of a row
agree:

    value: limit > 0 ? used : 0,
    total: limit > 0 ? limit : 1,

The `limit > 0` path computes identically to before, so only the unlimited case
changes. The tooltip already reported "No limit set" and is unchanged.

Fixes #1094 

## Steps to Test

1. Apply a NodePool with no `spec.limits.cpu` and let it provision a node.
2. Open Karpenter > NodePools.
3. The CPU bar is empty rather than full, matching the Memory bar in the same row.
4. Re-add `limits.cpu` and confirm the bar fills proportionally again.

Verified locally in `karpenter/`: `npm run build`, `npm run tsc`, `npm run lint`
and `npm run test` all pass.

### DIFF
--- a/karpenter/src/NodePool/List.tsx
+++ b/karpenter/src/NodePool/List.tsx
@@ -84,16 +84,14 @@ function NodePoolsList() {
             const data: ChartDataPoint[] = [
               {
                 name: 'CPU',
-                value: used,
+                value: limit > 0 ? used : 0,
               },
             ];
-
-            const effectiveLimit = limit > 0 ? limit : used || 1;
 
             return (
               <PercentageBar
                 data={data}
-                total={effectiveLimit}
+                total={limit > 0 ? limit : 1}
                 tooltipFunc={() => CPUtooltip(used, limit, t)}
               />
             );


Bhargava-Ram-Thunga:
Hi @Shashank200345,

I'm reviewing this PR! Aligning the CPU `PercentageBar` calculation in `NodePool/List.tsx` with the Memory column fix from #371 (`limit > 0 ? used : 0`) properly handles unlimited CPU NodePools.

I'll also review and test this further to help move it forward. Nice work!

Bhargava-Ram-Thunga (commented):
Hi @Shashank200345,

I'm reviewing this PR! LGTM from my side — I checked out the branch and ran `tsc` and `lint` locally (both pass), and confirmed the new CPU expressions (`value: limit > 0 ? used : 0`, `total: limit > 0 ? limit : 1`) are byte-for-byte identical to what the Memory column has used since #371, so both bars in a row now agree. The old `total = used || 1` made the fill ratio exactly 1 for any usage whenever `spec.limits.cpu` was unset, so the bar was always full and always wrong; the limited path computes identically to before, and the `CPUtooltip` already renders "No limit set" to explain the empty bar. Minimal, correct, consistent.

One optional follow-up: the CPU and Memory renderers are now identical modulo the parser and could share a small helper — cleanup only, not a requirement (worth noting the karpenter plugin currently has no real unit tests, so that refactor plus a couple of tests would be a nice future PR).

PR #1103, opened 2026-08-09
knative: Check Serving and Eventing separately

RajdeepKushwaha5:
Follow up to https://github.com/cncf/mentoring/issues/2001#issuecomment-5224692202, where @kahirokunn confirmed the install check is Serving only and said a standalone PR for it was welcome.

Scope is the detection split and the gating that uses it. No Eventing views, routes or sidebar entries, so #922 is untouched.

## The problem

Serving and Eventing install independently. A cluster can run either one, both, or neither. `isKnativeInstalled` looked for `services.serving.knative.dev` and returned that one answer as "Knative is installed", and every caller acted on it:

- the sidebar filter showed or hid all Knative entries on it
- `useKnativeInstalled` gated every view on it

That answer is correct for the Serving views we have today, and it is not something a caller can act on once anything reads Eventing resources.

## What this changes

`isKnativeComponentInstalled(component, clusters)` replaces `isKnativeInstalled`. Serving is detected through `services.serving.knative.dev` and Eventing through `brokers.eventing.knative.dev`, both part of their core install rather than an optional add-on. `hasCrdInCluster` is unchanged, and so is the rule that a component counts as installed only when it is present in every selected cluster.

`SECTION_COMPONENT` in `sidebarGating.ts` maps each of the three navigation sections to the component it needs. Items inherit the component of the section they sit in, derived from `knativeNavigationSections`, so `navigation.ts` stays
the source of truth for what exists and `SECTION_COMPONENT` is the only place a component is decided. Adding an Eventing section is one line in each.

The top level entry is gated on every component its sections need, so it is hidden exactly when all of them would be hidden. That is the reason it is not simply gated on "any component installed": `filterSublist` in `useSidebarItems.tsx` removes children but keeps the parent, and `SidebarItem` falls back to `routeName = ''` when `subList` is empty. A top level entry that outlived its sections would leave a dead item in the sidebar, which is what an Eventing only cluster would have got.

The gating logic moved to `sidebarGating.ts` with the install check passed in, so it can be tested without loading the plugin runtime. The cache is keyed by cluster and component instead of cluster alone, and keeps the 30 second TTL and the single in flight check per key.

**This costs no extra API calls.** Only components that some section asks for are ever looked up. All three sections map to Serving today, so the sidebar makes exactly the same one check per cluster that it made before. A test asserts this.

Two smaller things are tightened on the way. Neither is reachable in a running Headlamp and I am not claiming them as user facing fixes:

- a key that had never been checked compared against a default of `0`, which reads as a fresh result rather than a missing one. With `Date.now()` this never fires, but it made the cache impossible to test with an injected clock.
- a rejected check left its in flight flag set, which would freeze that cluster's result for good. `hasCrdInCluster` settles rather than rejects, so this needs `apiGet` to throw synchronously to happen at all.

## On YAGNI

You raised YAGNI on #2001, and it is the fair objection here, so I would rather answer it than hope it does not come up.

`eventing` is not speculation about a future view. It is the thing that makes this a split rather than a rename: without a second component there is no way to express "this section needs Serving" as distinct from "Knative is present", and the bug you confirmed stays exactly where it is. It costs one line in a lookup table, and the tests cover Serving only, Eventing only, both and neither, so the behavior is pinned before anything depends on it.

What I deliberately did not do is guess at what the Eventing views will need. There are no Eventing sections, routes or resource classes here. If the shape turns out wrong when those land, changing a map entry is cheap.

## Behavior today

None of this changes what a user sees, and I would rather say that plainly than dress it up. All three sections map to Serving, so the sidebar and the views resolve to the same answer they did before. This is the seam, so the Eventing work does not have to carry a detection rewrite as well.

## Tests and verification

`isKnativeInstalled.test.ts` covers Serving only, Eventing only, both, neither, the multi cluster rule, a failed request, and that the watch is cancelled once a result is known.

`sidebarGating.test.ts` covers the per section mapping, item inheritance, the top level rule, the per cluster and per component cache, TTL expiry, single in flight check, recovery after a rejection, and that no component is looked up unless a section needs it. It also walks every entry `index.tsx` registers, the top level one, the three sections and all thirteen items, and asserts each resolves to a known component and is shown or hidden with the rest, so a section added to `navigation.ts` without a component mapping fails the suite.

140 tests pass. `tsc`, `lint`, `format --check` and `build` are clean.

Verified on a real cluster as well: Kubernetes 1.34.3, Knative Serving 1.19.0, Headlamp 0.44.0. All thirteen menus open and match `kubectl`, and removing `services.serving.knative.dev` hides the whole tree and shows the not detected banner rather than an error. Videos are in the comments below.

I did not add manifests under `test-files/`. There is no Eventing view to look at yet, so a Broker fixture would only be there to make this look bigger than it is.

## Overlap with other open PRs

- Rebased onto `main` at 5bac06b6, which includes #986. That PR restructured the sidebar into three sections with items underneath, so the gate now walks sections and items rather than a flat list.
- #1077 rewrites the error handling inside `hasCrdInCluster`, which this PR leaves alone. The two apply cleanly in either order.
- #1075 adds tests for `isKnativeInstalled`, which this PR renames. If that lands first I am happy to rebase and fold its cases into `isKnativeInstalled.test.ts`.

## One thing I noticed and left alone

In `hasCrdInCluster`, `cancelFn` is assigned in the `.then` of `request()`. If a callback fires before that microtask runs, the result settles while `cancelFn` is still null and the watch is never cancelled. It needs a genuinely synchronous callback, which I have not seen happen, and it sits in the function #1077 is already rewriting, so it seemed better raised than patched here.

### DIFF
--- a/knative/src/components/certificates/List.tsx
+++ b/knative/src/components/certificates/List.tsx
@@ -35,7 +35,7 @@ import {
 
 export function CertificatesList() {
   const clusters = useClusters();
-  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled(clusters);
+  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled('serving', clusters);
   const showClusterColumn = clusters.length > 1;
   const columns = React.useMemo<
     (ResourceTableColumn<KnativeCertificate> | 'name' | 'namespace' | 'cluster' | 'age')[]
--- a/knative/src/components/configurations/List.tsx
+++ b/knative/src/components/configurations/List.tsx
@@ -27,7 +27,7 @@ import { getReadinessColumns, NamespacedResourceLink } from '../common/ResourceL
 
 export function ConfigurationsList() {
   const clusters = useClusters();
-  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled(clusters);
+  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled('serving', clusters);
   const showClusterColumn = clusters.length > 1;
   const columns = React.useMemo<
     (ResourceTableColumn<KnativeConfiguration> | 'name' | 'namespace' | 'cluster' | 'age')[]
--- a/knative/src/components/images/List.tsx
+++ b/knative/src/components/images/List.tsx
@@ -31,7 +31,7 @@ import {
 
 export function ImagesList() {
   const clusters = useClusters();
-  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled(clusters);
+  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled('serving', clusters);
   const showClusterColumn = clusters.length > 1;
   const columns = React.useMemo<
     (ResourceTableColumn<KnativeImage> | 'name' | 'namespace' | 'cluster' | 'age')[]
--- a/knative/src/components/ingresses/List.tsx
+++ b/knative/src/components/ingresses/List.tsx
@@ -29,7 +29,7 @@ import { getReadinessColumns, TextValue } from '../common/ResourceListCells';
 
 export function IngressesList() {
   const clusters = useClusters();
-  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled(clusters);
+  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled('serving', clusters);
   const showClusterColumn = clusters.length > 1;
   const columns = React.useMemo<
     (ResourceTableColumn<KnativeIngress> | 'name' | 'namespace' | 'cluster' | 'age')[]
--- a/knative/src/components/kservices/Detail.tsx
+++ b/knative/src/components/kservices/Detail.tsx
@@ -43,7 +43,7 @@ import { TrafficSection } from './detail/sections/traffic/TrafficSection';
 export function KServiceDetail() {
   const { name, namespace } = useParams<{ namespace: string; name: string }>();
   const clusters = useClusters();
-  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled(clusters);
+  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled('serving', clusters);
   return (
     <KServiceEditContextProvider>
       {isKnativeInstalled ? (
--- a/knative/src/components/kservices/List.tsx
+++ b/knative/src/components/kservices/List.tsx
@@ -613,7 +613,7 @@ function KServicesListContents({ clusters }: KServicesListContentsProps) {
 
 export function KServicesList() {
   const clusters = useClusters();
-  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled(clusters);
+  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled('serving', clusters);
 
   return isKnativeInstalled ? (
     <KServicesListContents clusters={clusters} />
--- a/knative/src/components/metrics/List.tsx
+++ b/knative/src/components/metrics/List.tsx
@@ -32,7 +32,7 @@ import {
 
 export function MetricsList() {
   const clusters = useClusters();
-  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled(clusters);
+  const { isKnativeInstalled, isKnativeCheckLoading } = useKnativeInstalled('serving', clusters);
   const showClusterColumn = clusters.length > 1;
   const columns = React.useMemo<
     (ResourceTableColumn<KnativeMetric> | 'name' | 'namespace' | 'cluster' | 'age')[]
--- a/knative/src/components/networking/Configuration.tsx
+++ b/knative/src/components/networking/Configuration.tsx
@@ -298,7 +298,7 @@ export function ClusterNetworkingCard({
 export function NetworkingConfiguration() {
   const clusters = useClusters();
   const hasCluster = clusters.length > 0;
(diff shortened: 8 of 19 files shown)


kahirokunn:
The implementation looked good. Thank you!
The `knative/README.md` seemed somewhat inconsistent with other READMEs I've seen, so please review it and remove any discrepancies.

kahirokunn:
Also, I would appreciate it if you could rebase using the latest main branch and confirm that opening all Knative menus works properly.
If possible, I would appreciate it if you could provide a screenshot or a video showing the test results.
Thank you 🙏

RajdeepKushwaha5:
Thanks @kahirokunn, all three done.

**README**

You were right, it did not match the other plugins. I had added Install Detection as a top level section between Demo and Development, and no other plugin README documents internals at that level. It is now a `### Install detection` subsection
inside `## Development`, next to the existing Prometheus notes, and trimmed to what a contributor needs rather than a walkthrough of the functions. The diff is 13 added lines and nothing removed.

One thing I did not change, since it came from #986 and is yours to call: this README uses `## Requirements` where the other plugins that have one all use `## Prerequisites` (keda, kyverno, kueue, argocd, cert-manager, strimzi, volcano, metal3). Happy to rename it if you would like consistency there too.

**Rebase**

Rebased onto main at 5bac06b6, so this now sits on top of #986. Three things needed real changes rather than conflict resolution:

- `index.tsx` moved to the three level tree, so gating now maps each navigation section to a component and items inherit from the section they sit in.   `navigation.ts` is the source of truth for what exists and `SECTION_COMPONENT` is the only place a component is decided, so adding an Eventing section is one line in each. 
- `knativeSidebarParents` is gone, since the gate owns that logic now.
- `networking/Overview.tsx` became `Configuration.tsx`, and `useKnativeInstalled` call sites went from 5 to 13 with the new list views. All 13 name their component now.

I also folded the API change and its callers into one commit, so each of the three commits typechecks on its own.

**Verification**

Built the branch and ran it against a real cluster:

- Headlamp v0.44.0
- Knative Serving v1.19.0 with net-kourier v1.19.0
- Kubernetes v1.34.3

Opened all 13 menus. Every one renders, and the row counts match the cluster:
Services 4, Revisions 11, Domain Mappings 3, Cluster Domain Claims 1, Configurations 4, Routes 4, Images 8, Pod Autoscalers 7, Metrics 7, KIngresses 5, Serverless Services 7, Certificates 1. Networking reads the live ingress class and shows `kourier`. No 404 pages, no not-detected banners, and no uncaught JavaScript errors across the whole pass.

**All 13 menus** (knative-all-menus.webm)

[knative-all-menus.webm](https://github.com/user-attachments/assets/8d460bae-7fe7-4ca6-8a7f-37358da0b70c)

It clicks the real sidebar links rather than typing URLs, so it also shows the entries registered under the three section headings, and ends on a Service detail view.

**The gate hiding** (knative-gate-hides.webm)

[knative-gate-hides.webm](https://github.com/user-attachments/assets/3ee9fe6e-251c-45d3-a8ba-af026863305c)


The first video only shows that nothing regressed, since every section maps to Serving and this cluster has Serving. It would look the same without this PR. The second one takes Serving away on the same cluster:

1. Serving present, Knative and all 13 entries in the sidebar.
2. `kubectl delete crd services.serving.knative.dev`.
3. Reload. The Knative entry and all three sections are gone, and nothing else in the sidebar is affected.
4. Opening /knative/services directly still resolves and shows the not detected banner rather than an error, which is the `useKnativeInstalled` path the 13 view call sites now use.

I restored the CRD and the fixtures afterwards.

<details>
<summary>Same 13 views as one image, if that is easier to skim</summary>



</details>

There is also a test that walks every entry `index.tsx` registers, the top level one, the three sections and all thirteen items, and asserts each resolves to a known component and is shown or hidden with the rest. A section added to `navigation.ts` without a component mapping fails the suite instead of silently falling through. 140 tests pass, and `tsc`, `lint`, `format --check` and `build` are clean.

One small thing I noticed while running it, unrelated to this PR: the plugin declares `"i18n": ["en"]` in package.json but ships no `locales` directory, so Headlamp requests `/plugins/knative/locales/en/translation.json` on every page and gets a 404. Harmless, but it lines up with the i18n discussion on the program issue.

knative/README.md
```diff
@@ -132,6 +132,19 @@ Remove the test namespace, optional PodMonitor, and cluster-scoped test claim wi
...
+Knative Serving and Eventing install independently, so a cluster can run either one, both, or
+neither. Each piece of UI checks for the component it needs rather than for Knative as a whole.
+
+`SECTION_COMPONENT` in `src/sidebarGating.ts` maps each navigation section to its component, and
+items inherit the component of the section they sit in. Adding a section to `src/navigation.ts`
+means adding one line there. Views pass the same component to `useKnativeInstalled`.
+
+To check this by hand, install one component without the other. Serving alone is the default of the
+[official installation guide](https://knative.dev/docs/install/); for Eventing alone, apply
+`eventing-crds.yaml` and `eventing-core.yaml` on a cluster where Serving was never installed.
```
kahirokunn: Since this is documentation for users, it’s probably best to omit this part because it’s too detailed.

RajdeepKushwaha5: Agreed, dropped it. I removed the whole section rather than just those lines, since what was left would have been two sentences that do not tell a user anything they can act on, and the README is now identical to main.

The detail lives in the code instead: `COMPONENT_CRD_NAME` in `isKnativeInstalled.ts` and `SECTION_COMPONENT` in `sidebarGating.ts` both document what they are for and what adding an Eventing section would take.

RajdeepKushwaha5:
Hi @illume @skoeva when you have a moment, could one of you release the workflow run on this? The checks have not started yet, and I believe that is the first time contributor approval gate rather than anything in the branch. DCO passed.

It is rebased on main at 5bac06b6 and green locally: tsc, lint, format --check, build, and 140 tests. 
Happy to rebase again if it has gone stale by then.

PR #1117, opened 2026-08-11
cnpg: add a read-only CloudNativePG plugin

gambtho:
## What this adds

A new `cnpg` plugin giving day-2 visibility into [CloudNativePG](https://cloudnative-pg.io/)
Postgres clusters, built entirely from the `postgresql.cnpg.io/v1` custom resources
a user already has.

**The plugin is read-only.** It never triggers a failover, switchover, backup or
any other mutating operation, and it exposes no control that could. It also makes
no network calls of its own: no metrics backend, no external service. It reads
three CRDs — `Cluster`, `Backup`, `ScheduledBackup` — and nothing else.

The **CloudNativePG** sidebar section only appears on clusters where the CRDs are
installed; elsewhere the plugin stays dormant and adds nothing to the UI.

### Clusters list

Name, namespace, phase (colour-coded, with the operator's phase reason on hover),
ready/desired instances, current primary, last successful backup, and age.

CloudNativePG clusters list

### Cluster detail

- **Insights** — findings from a deterministic rules engine, each shown with the
  evidence it was derived from. Seven checks: backups behind schedule, no
  restorable point, unhealthy cluster phase, fewer instances ready than the spec
  asks for, failing WAL archiving, single instance, and a recent primary change.
  Nothing is inferred from metrics, logs or events, and no finding proposes a
  remediation, because the plugin performs none.
- **Topology** — each instance with its role, operator-reported state, timeline
  and pod IP.
- **Backup and restore confidence** — the configured destination, the operator's
  WAL-archiving verdict, last successful and last failed backup, any backup in
  progress, the oldest restorable point, the schedules targeting the cluster, and
  the ten most recent `Backup` objects.

CloudNativePG cluster detail

The provider interface behind the insights panel is open, so further insight
sources can be registered without touching the views. This change ships exactly
one provider: the rules engine above, pure functions over the three CRDs.

## Deprecated backup fields

`Cluster.status.lastSuccessfulBackup` (and its `*ByMethod` variant),
`lastFailedBackup` and `firstRecoverabilityPoint` are marked upstream as
*"Deprecated: the field is not set for backup plugins"*. A cluster archiving
through the barman-cloud plugin leaves them empty no matter how healthy its
backups are, so reading them would report a false "no backups" on exactly the
clusters that are backed up.

Backup facts are therefore derived from `Backup` objects, which carry no such
caveat. The one fact with no `Backup`-object equivalent is the oldest restorable
point, so that — and only that — falls back to `firstRecoverabilityPoint`, and
shows "Unknown" when absent.

## Behaviour when data is missing

A field the operator has not reported renders as "Unknown" with the reason on
hover, never as a zero or a dash; unreported instance counts show `?`, not `0`.
An RBAC denial renders as an empty state naming the exact permission
(`list backups.postgresql.cnpg.io in namespace "prod"`), scoped to the affected
view, so losing access to `Backup` objects leaves topology and phase intact.
Checks that cannot be decided report "unknown" rather than assuming the good case.

One host-version caveat, documented in the README: Headlamp releases before 0.44 do
not surface a denied *namespaced* list to plugins (fixed upstream in 0.44,
kubernetes-sigs/headlamp 347c13f16), so on older hosts the detail view shows empty
backup sections instead of naming the permission. The list view names it on every
version, as does the detail view from 0.44 onward.

## Verification

- `npm run build`, `npm run tsc`, `npm run lint`, `npm run format -- --check` and
  `npm run test` all pass (152 unit tests).
- Verified live against the CloudNativePG **1.29** and **1.30** API shapes, on
  Headlamp **0.39** and **0.44**, on kind and on AKS, including healthy,
  single-instance, unbacked and stuck-provisioning clusters; barman object store,
  WAL-archiver plugin and volume-snapshot destinations; a partially-denied
  ServiceAccount; and a backup-heavy cluster (1000 `Backup` objects), where the
  views issue no per-row API calls.


cnpg/src/components/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,126 @@
...
+ * Split from the fetch above so the backup hook is called unconditionally: the
+ * cluster is known to exist here, so there is no early return above the hook.
+ */
+function ClusterDetailContent({ cluster, title }: { cluster: ClusterClass; title: string }) {
+  const { t } = useTranslation();
+  const backupData = useClusterBackupData(cluster);
+
+  const image = cluster.status.image ?? cluster.spec.imageName ?? null;
+
+  return (
```
yolossn, line 131: This should use `DetailsGrid` component instead of custom component to provide a consistent UX.

cnpg/src/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+// Phase 1 ships exactly one insights provider: the deterministic rules engine.
+// It is registered here, at plugin load, rather than imported by the panel, so
+// that adding a provider later is a registration and not a change to the UI.
+registerInsightsProvider(rulesProvider);
+
+// Neutral database glyph: the PostgreSQL elephant is a trademarked mark and is
+// deliberately not reproduced here.
+addIcon('custom:cnpg', {
+  body: `<path fill="currentColor" d="M12 2C7.58 2 4 3.79 4 6v12c0 2.21 3.58 4 8 4s8-1.79 8-4V6c0-2.21-3.58-4-8-4m0 2c3.87 0 6 1.5 6 2s-2.13 2-6 2s-6-1.5-6-2s2.13-2 6-2M6 8.61c1.46.86 3.61 1.39 6 1.39s4.54-.53 6-1.39V12c0 .5-2.13 2-6 2s-6-1.5-6-2zm0 6c1.46.86 3.61 1.39 6 1.39s4.54-.53 6-1.39V18c0 .5-2.13 2-6 2s-6-1.5-6-2z"/>`,
```
yolossn: nit: `database-outline` icon can be used instead of this.

https://pictogrammers.com/library/mdi/icon/database-outline/

cnpg/src/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+    // shows the entry — because a thrown error is not evidence of absence.
+    installedByCluster[cluster] = installedByCluster[cluster] ?? null;
+  } finally {
+    // Stamped in `finally`, so a failed check backs off for the TTL like a
+    // successful one. Stamping only on success lets a persistently failing
+    // cluster re-fire the request on every sidebar render.
+    lastCheckedAt[cluster] = Date.now();
+    inFlight[cluster] = false;
+  }
+}
```
yolossn, line 68: Nice pattern for doing cluster-level installation check

cnpg/src/components/clusters/ClusterDetail.tsx
yolossn: Is this not using the `DetailsGrid` so that the `Delete` quick action isnt shown by default?

yolossn (changes requested):

cnpg/src/components/clusters/ClusterDetail.tsx
gambtho: Yes, exactly.  I didn't use DetailsGrid because the default header ships Restart/Scale/Edit/Delete and this plugin is strictly read-only. Switched to DetailsGrid with noDefaultActions (and no withEvents), and kept the RBAC empty states that name the missing permission — re-verified that path live before pushing. Icon switched to database-outline.

PR #852, opened 2026-06-24
metal3: Add BareMetalHost plugin with list and detail views

mastermaxx03:
## Summary

Initial Metal3 plugin. Adds a Metal3 section to Headlamp for the
BareMetalHost resource. First phase of the plugin development; more resources and a
map view to follow.

## Included

- BareMetalHost list view with a composite Status column (operational
  status, provisioning state, and error type in one cell).
- BareMetalHost detail view (config, status, consumer, power, BMC).
- A reusable CRD guard that shows a clear message when the Metal3 CRDs
  aren't installed in the cluster.
- Unit tests for the status logic and a Storybook story.

## How to test

1. A cluster with the Metal3 CRDs installed.
2. `cd metal3 && npm install && npm start`, then run Headlamp.
3. Apply the sample host: `kubectl apply -f metal3/test-files/baremetalhost.yaml`.
4. Open the Metal3 section: the host lists, and clicking it opens the detail view.

## Snapshots

Metal3 BareMetalHost list and detail views - the composite Status column and detail page surface all four status axes (operational status, provisioning state, error type, power) together.

image

image

## Scope

Read-only, BareMetalHost only for now. The cluster-api-provider-metal3
resources, the relationship map, and mutating actions are planned follow-ups.

Relates to kubernetes-sigs/headlamp#5268.

PR #1119, opened 2026-08-11
metal3: BareMetalHost: Show hardware inventory on the detail

mastermaxx03:
## Summary

Adds a hardware inventory to the BareMetalHost detail view. Once a host has been inspected, Ironic
records its CPU, memory, disks, and NICs under `status.hardware`, and this surfaces that as three
sections on the detail page: Hardware (CPU model and cores, memory, system vendor and model),
Storage (each disk with its size and whether it is an SSD or HDD), and Network (each NIC with its
MAC, IP, and speed). It is the same reference information MAAS and other bare-metal tools put on a
machine's page, and it is what an operator uses to pick or verify a host.

Hosts that have not been inspected yet, still registering or inspecting, carry no `status.hardware`,
so the sections simply do not appear for them.

It is pure display over data the host already reports, so nothing new is fetched.

## Dependencies

Not stacked on anything. Branches off main and extends the existing BareMetalHost detail from #852.

## How to test

1. A cluster with the Metal3 CRDs and an inspected BareMetalHost.
2. `cd metal3 && npm install && npm start`, then run Headlamp.
3. Open an inspected host's detail; the Hardware, Storage, and Network sections appear below the
   existing fields. A host still registering or inspecting shows none of them.

## Media

image

### DIFF
--- a/metal3/src/BareMetalHost/Details.tsx
+++ b/metal3/src/BareMetalHost/Details.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 The Kubernetes Authors
+ * Copyright 2026 The Kubernetes Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


mastermaxx03:
@Rozzii Added an inspected-hardware section (CPU, memory, disks, NICs) to the BareMetalHost
detail, since similar plugin/tools had it. do we keep this?

illume (approved):
🎉 thanks!

Rozzii:
> @Rozzii Added an inspected-hardware section (CPU, memory, disks, NICs) to the BareMetalHost detail, since similar plugin/tools had it. do we keep this?

Yes it is fine to keep

PR #1122, opened 2026-08-11
fix(karpenter): safely handle undefined conditions in NodeClaim views

### LINKED ISSUE #1120: karpenter: NodeClaim list and detail views crash when status.conditions is undefined
### Description
While looking through the karpenter plugin code, I noticed the NodeClaim (Scaling) list and detail views will crash with a `TypeError` if a NodeClaim resource doesn't have `status.conditions` populated yet.

This happens because the code uses optional chaining up to `status?.conditions`, but then immediately indexes into the result with `[length - 1]` without checking if `conditions` is actually defined.

When Karpenter is actively provisioning nodes, there's a short window where a NodeClaim exists but `status.conditions` hasn't been populated by the controller yet. During that window, opening the NodeClaims list or clicking into a NodeClaim detail page crashes the entire view.

### Steps to Reproduce
1. Have Karpenter installed and trigger a scale-up so new NodeClaims are being created.
2. Open Headlamp and navigate to **Karpenter -> Scaling (NodeClaims)**.
3. If any NodeClaim is in an early lifecycle state where `status.conditions` is missing/undefined, the page crashes.

Alternatively, you can reproduce by creating a NodeClaim manifest where `status` or `status.conditions` is absent and viewing it in Headlamp.

### Expected Behavior
The Status column should gracefully show `"-"` or `"Unknown"` when `status.conditions` is missing.

### Actual Behavior
The page throws a `TypeError: Cannot read properties of undefined (reading '0')` and the entire list/detail view crashes.

### Where it's happening in code

**`karpenter/src/Scaling/List.tsx` (line 67-70):**
```tsx
const finalCondition = item.jsonData?.status?.conditions?.length;
// If conditions is undefined, finalCondition is undefined.
// undefined - 1 = NaN, and conditions[NaN] throws TypeError.
return (
  <StatusLabel>
    {item.jsonData?.status?.conditions[finalCondition - 1]?.reason || '-'}
  </StatusLabel>
);
```
`karpenter/src/Scaling/Details.tsx` (line 27, 53):
```tsx
const conditionsLength = item.jsonData.status?.conditions?.length || 0;
// conditionsLength is 0 when conditions is undefined.
// conditions[0 - 1] = conditions[-1] = undefined... but conditions itself is undefined, so:
// undefined[-1] throws TypeError.
{item.jsonData.status?.conditions[conditionsLength - 1]?.reason || '-'}
```

rakshaak29:
### Description
The NodeClaim list and detail views crash with `TypeError: Cannot read properties of undefined` when certain fields are missing from the resource object.

Two categories of crashes were identified:

1. **`status.conditions` is undefined** — When a NodeClaim has just been created and the controller hasn't populated conditions yet, accessing the last element of an undefined array throws.
2. **`metadata.labels` is undefined** — When a NodeClaim has no labels, the bracket-notation access (`labels['key']`) on `undefined` throws. This affects both rendering and sorting — clicking a column header like "Instance Type" or "Zone" will crash the page.

### Changes
- **`Scaling/Details.tsx`**: Safe-guard `status.conditions` with `Array.isArray()` check, and add optional chaining before all bracket-notation label accesses (`labels?.['...']`).
- **`Scaling/List.tsx`**: Same safe-guards for conditions and all label accesses in both `render` and `getValue`.
- **`Scaling/Scaling.test.ts`** *(new)*: Unit tests covering undefined status, empty conditions, undefined metadata, undefined labels, and fully populated objects.

Closes #1120

### Checklist
- [x] Signed commits with Developer Certificate of Origin (DCO `-s`)
- [x] Code follows repository style guidelines
- [x] TypeScript compilation passes (`tsc --noEmit`)
- [x] Unit tests added and passing

### DIFF
--- a/karpenter/src/Scaling/Details.tsx
+++ b/karpenter/src/Scaling/Details.tsx
@@ -24,8 +24,14 @@ export function ScalingDetailView(props: { name?: string }) {
       name={name}
       withEvents
       extraInfo={item => {
-        const conditionsLength = item.jsonData.status?.conditions?.length || 0;
-        const nodepool = item.jsonData.metadata?.ownerReferences?.find(x => x.kind === 'NodePool');
+        const conditions = item?.jsonData?.status?.conditions;
+        const lastCondition =
+          Array.isArray(conditions) && conditions.length > 0
+            ? conditions[conditions.length - 1]
+            : null;
+        const nodepool = item?.jsonData?.metadata?.ownerReferences?.find(
+          (x: any) => x.kind === 'NodePool'
+        );
 
         return (
           item && [
@@ -48,11 +54,7 @@ export function ScalingDetailView(props: { name?: string }) {
             },
             {
               name: t('Status'),
-              value: (
-                <StatusLabel>
-                  {item.jsonData.status?.conditions[conditionsLength - 1]?.reason || '-'}
-                </StatusLabel>
-              ),
+              value: <StatusLabel>{lastCondition?.reason || '-'}</StatusLabel>,
             },
             {
               name: t('Instance Type'),
--- a/karpenter/src/Scaling/Details.tsx
+++ b/karpenter/src/Scaling/Details.tsx
@@ -58,11 +58,11 @@ export function ScalingDetailView(props: { name?: string }) {
             },
             {
               name: t('Instance Type'),
-              value: item.jsonData.metadata?.labels['node.kubernetes.io/instance-type'] || '-',
+              value: item.jsonData.metadata?.labels?.['node.kubernetes.io/instance-type'] || '-',
             },
             {
               name: t('Capacity Type'),
-              value: item.jsonData.metadata?.labels['karpenter.sh/capacity-type'] || '-',
+              value: item.jsonData.metadata?.labels?.['karpenter.sh/capacity-type'] || '-',
             },
           ]
         );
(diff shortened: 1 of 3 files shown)


Ralthos (commented):
All eight sites are covered. The `Array.isArray` guard is a better shape than the plain optional chaining I had in mind, since it also handles `conditions` being present and empty.

Two things.

**The tests do not exercise the code they test.** `Scaling.test.ts` defines its own `getLastConditionReason`, `getInstanceType`, `getCapacityType` and `getZone`, and imports nothing. Revert both source files and every one of these still passes.

There is a good reason for that. Importing those modules under vitest does not work today: they pull in `@kinvolk/headlamp-plugin/lib/...`, which does not resolve on disk. I filed that as kubernetes-sigs/headlamp#7036, and it appears to be why most plugins in this repo have no tests at all.

The way around it is to move the helper into a module with no SDK imports, then import it from both the component and the test. I did that in #1079 for keda's map edges for this same reason. `getLastConditionReason(jsonData)` is already written in your test file; it needs to live in `Scaling/` and be imported in both places, and then a regression in `List.tsx` fails something.

**The Status column's `getValue` still returns the raw array.**

```ts
getValue: item => item.jsonData?.status?.conditions,
```

`render` now shows `lastCondition?.reason`, but sorting keys off the whole conditions array, so the column does not sort by what it displays. The SDK types this as `(item) => string | number | null | undefined`, and `jsonData` being `any` is why TypeScript stays quiet about an array going through. `lastCondition?.reason ?? ''` would line the two up.

That one is the same split I mentioned on the issue, landing in the one column where the fix went to `render` alone.

Smaller: the `(x: any)` on the `ownerReferences.find` callback was inferred before this change, so it can come off.

Issue #938, opened 2026-07-26
WIP: kyverno improvements


illume:
Please do not start on this issue.

hasancoded:
Hi @illume, noted that #938 is still WIP and reserved, so I have not started on any of the program work. Leaving one scoping observation from reading the plugin, for whenever it opens up.

### On the i18n item

The project description lists *"incomplete i18n coverage"*, but the state looks different from incomplete. The plugin is already instrumented throughout and has no catalogues at all.

Every major component calls `useTranslation()` and wraps its strings:

| Component | `useTranslation()` | `t('...')` calls |
|---|---:|---:|
| `PolicyViewer.tsx` | 9 | 32 |
| `CELPolicyViewer.tsx` | 3 | 35 |
| `ViolationsView.tsx` | 2 | 18 |
| `Dashboard.tsx` | 2 | 16 |
| `ComplianceBadge.tsx` | 2 | 10 |
| `ResultsTable.tsx` | 2 | 8 |

That is roughly 119 wrapped strings across the six. But there is no `kyverno/locales/` directory. `cert-manager`, `keda`, `prometheus`, `app-catalog`, `plugin-catalog` and `ai-assistant` each ship 19 locale files under `locales/<lang>/translation.json`; kyverno ships none, so every `t()` currently falls through to its English default.

So the remaining work does not look like "wrap the strings that were missed", which is largely done. It looks like adding the extraction step and the catalogues the mature plugins already have.

> **Question:** should i18n here mean matching the cert-manager language set, or standing up `locales/en/translation.json` plus the extraction script and leaving the other languages to the community? Those are quite different in size, and it changes whether i18n is a small task or a real slice of the term.

I applied for this project for Term 3 and have Headlamp running locally against a kind cluster with Kyverno installed. Happy to hold off on anything until the issue opens.

PR #1123, opened 2026-08-11
kyverno: add aria-pressed to violation status filter chips

0xMoni:
## What

The status filter chips in `ViolationsView` are toggle controls, but they expose no pressed state. Sighted users can tell selection apart from the `filled` vs `outlined` variant — screen reader users get no indication of which filters are currently active.

- `aria-pressed` on each chip, reflecting selection state
- `role="group"` with a translated label, so the chips are announced as one filter set rather than three unrelated buttons
- `aria-disabled` on the last remaining selected chip. `toggle()` already refuses to deselect it so the table is never empty, but that refusal was silent to everyone.

Follows the `aria-pressed` pattern already used by `AIAssistantToggle`.

## Why

The Kyverno plugin currently has no `aria-label` anywhere across its 21 components. This is a small first step; related to the accessibility work described in #938.

## Testing

- `npx tsc --noEmit -p tsconfig.json` — passes
- `npx eslint src/components/ViolationsView.tsx` — passes
- Prettier formatting verified

I could not run `headlamp-plugin tsc`/`lint` locally — the toolchain hits a `yargs` ESM error on Node 26, unrelated to this change.

## Question for maintainers

For the locked last chip, `aria-disabled` seemed the most honest signal, but an alternative would be keeping it fully enabled and showing a tooltip explaining why at least one filter must stay on. Happy to switch if you'd prefer that.

### DIFF
--- a/kyverno/src/components/ViolationsView.tsx
+++ b/kyverno/src/components/ViolationsView.tsx
@@ -120,6 +120,8 @@ function StatusFilterChips({
   onChange: (statuses: PolicyResultStatus[]) => void;
   counts: Record<PolicyResultStatus, number>;
 }) {
+  const { t } = useTranslation();
+
   function toggle(status: PolicyResultStatus) {
     if (selected.includes(status)) {
       if (selected.length > 1) {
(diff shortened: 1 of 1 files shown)


skoeva (changes requested):
hi, could you sign off your commit with `git commit -S`?

0xMoni:
Thanks! Signed off and force-pushed -- DCO is green now.

PR #1129, opened 2026-08-11
ai-assistant: AKS mcp configured by default for aks-desktop

ashu8912:
## What this does

Makes the AI assistant useful out of the box on AKS Desktop by preconfiguring the
`aks-mcp` MCP server and the AKS skills, and by actually binding discovered MCP
tools to the chat model.

The AKS Desktop distribution ships the `aks-mcp` binary on `PATH`, so the plugin
can offer it without asking the user to configure anything. The host is detected
by feature-testing `desktopApi.registerAKSCluster`, which vanilla Headlamp does
not expose, so nothing changes for other hosts.

## Changes

**Preconfigure the `aks-mcp` server** (`packages/ai-common/src/mcp/config/builtinServers.ts`,
`src/mcp/seedBuiltinServers.ts`)

- Registers the server on startup when running in AKS Desktop.
- Only components that need no extra CLI binary are enabled. `aks-mcp` validates
  every enabled component at startup and exits when its CLI is missing, so the
  default set of all components fails on a machine without `cilium`/`hubble`.
  `kubectl`, `helm`, `cilium` and `hubble` stay opt-in via MCP settings.
- Runs with `--access-level readonly`.

**Preconfigure the AKS skills** (`packages/ai-common/src/skills/builtinSources.ts`,
`src/skills/seedBuiltinSources.ts`)

- Seeds the skills from `microsoft/azure-skills`, filtered to the
  `skills/azure-kubernetes` subtree so the rest of the Azure collection stays
  out. The source is pinned to a commit with the content hash the loader
  verifies.

**Reconciliation, shared by both**

- The plugin records the definition it last wrote, not just the name, so later
  changes to a built-in reach installs that were already seeded.
- Anything the user edited, deleted, repinned, or created themselves under the
  same name is left untouched, and the `enabled`/`autoApprove` toggles always
  belong to the user.
- Seeding is skipped entirely when the stored configuration cannot be read, so a
  transient read failure cannot reconcile against an empty config and drop the
  user's servers.

**Bind discovered MCP tools to the chat model**
(`packages/ai-common/src/assistant/LangChainAssistantSession.ts`)

- The assistant session always constructed its tool manager without an MCP
  bridge, so the manager fell back to `NullToolClient`, discovery was skipped,
  and the model only ever saw the built-in `kubernetes_api_request` tool. The
  settings dialog worked because it builds its own client.
- Adds an `mcpClient` option forwarded to `LangChainToolManager`, and passes a
  shared `ElectronMCPClient` from the plugin. The client reports itself
  unavailable outside Electron, so non-desktop hosts keep the previous
  behaviour.
- Tool discovery now returns early when the global MCP toggle is off, since the
  host keeps per-tool metadata around and those tools would always fail to run.

**Project events in the assistant context**
(`src/components/appbar/HeadlampEventHandler.tsx`, `src/prompts/promptGenerator.ts`)

- Mirrors Headlamp's project events (list view, details view, details tab
  change, project create/delete) into plugin state so the assistant knows which
  project, namespaces, clusters and tab the user is looking at, and suggests
  project-aware prompts.

## Testing

Unit tests cover reconciliation (first seed, refresh, user customization,
deletion, legacy state migration, env comparison), MCP tool discovery including
the globally-disabled path, host detection, the project event handler, and the
prompt suggestions.

### DIFF
--- a/ai-assistant/packages/ai-common/package.json
+++ b/ai-assistant/packages/ai-common/package.json
@@ -53,6 +53,7 @@
     "./skills/SkillLoader": "./src/skills/SkillLoader.ts",
     "./skills/SkillManager": "./src/skills/SkillManager.ts",
     "./skills/adapters/browser": "./src/skills/adapters/browser.ts",
+    "./skills/builtinSources": "./src/skills/builtinSources.ts",
     "./skills/config": "./src/skills/config.ts",
     "./skills/routing/EmbeddingSkillRouter": "./src/skills/routing/EmbeddingSkillRouter.ts",
     "./skills/routing/KeywordSkillRouter": "./src/skills/routing/KeywordSkillRouter.ts",
--- a/ai-assistant/packages/ai-common/package.json
+++ b/ai-assistant/packages/ai-common/package.json
@@ -28,6 +28,7 @@
     "./mcp/routing/EmbeddingToolRouter": "./src/mcp/routing/EmbeddingToolRouter.ts",
     "./mcp/routing/ToolRouter": "./src/mcp/routing/ToolRouter.ts",
     "./mcp/types": "./src/mcp/types.ts",
+    "./mcp/config/builtinServers": "./src/mcp/config/builtinServers.ts",
     "./mcp/config/serverConfig": "./src/mcp/config/serverConfig.ts",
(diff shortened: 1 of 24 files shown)


illume (approved):
🎉 thanks!

illume (approved):
Nice work!

PR #1159, opened 2026-08-13
app-catalog: Bump deps with npm audit fix

skoeva:


illume (changes requested):
Looks like dco is missing

PR #864, opened 2026-06-29
kueue: Add Kueue ClusterQueue views

Utkarshpandey0001:
## Summary

Adds support for Kueue `ClusterQueue` resources in the Headlamp Kueue plugin.

## Changes

- Adds a typed `ClusterQueue` resource model for `kueue.x-k8s.io/v1beta2`
- Adds ClusterQueue list and detail pages
- Registers ClusterQueue sidebar entry and routes
- Displays resource groups, referenced ResourceFlavors, workload counts, conditions, preemption, admission checks, and fair sharing details
- Adds formatter tests and a sample ClusterQueue manifest
- Updates the Kueue README with ClusterQueue usage notes

##Screenshots
Screenshot 2026-07-05 at 10 09 40 AM
Screenshot 2026-07-05 at 10 10 00 AM
Screenshot 2026-07-05 at 10 10 17 AM

https://github.com/user-attachments/assets/919dd1ef-d672-46a9-b6fb-ea6c05ecbb75

## Testing

```bash
cd kueue
npm run tsc
npm run lint
node_modules/.bin/vitest run -c node_modules/@kinvolk/headlamp-plugin/config/vite.config.mjs
npm run build

PR #1177, opened 2026-08-14
kueue: fix ClusterQueue flavor names to match ResourceFlavor examples

Adi-204:
## Summary
- `kueue/test-files/deploy/clusterqueue-team-a.yaml` referenced flavors named `default` and `spot`, but the corresponding `ResourceFlavor` test manifests (`resourceflavor-default.yaml` / `resourceflavor-spot.yaml`) are named `default-flavor` and `spot-flavor`.
- Following the documented steps (apply ResourceFlavors, then ClusterQueue, then a Job), the ClusterQueue never becomes ready to admit workloads because it points at flavors that don't exist.
- Renamed the flavor references in the ClusterQueue to `default-flavor` / `spot-flavor` so they match the ResourceFlavor examples.

## Root cause
The `ClusterQueue` test file was added in #864 with flavor names `default`/`spot`. The `ResourceFlavor` example files were added later already named `default-flavor`/`spot-flavor`, and the `ClusterQueue` was never updated to match.

## Verification
Reproduced against a live cluster (minikube + Kueue v0.19.1, `v1beta2` API):
- Before fix: `kubectl get clusterqueue team-a` → `status.conditions: Active=False, reason=FlavorNotFound`. No workload can ever be admitted.
- After fix: `Active=True, reason=Ready, message="Can admit new workloads"`.
- Ran the full flow (ResourceFlavors → ClusterQueue → LocalQueue → sample Job): the sample Job's workload reached `QuotaReserved=True`, `Admitted=True`, and pods ran successfully.

## Test plan
- [x] Applied `resourceflavor-default.yaml`, `resourceflavor-spot.yaml`, `resourceflavor-topology.yaml`, then the original `clusterqueue-team-a.yaml` — confirmed `Active=False (FlavorNotFound)`.
- [x] Re-applied with the fix — confirmed `Active=True (Ready)`.
- [x] Applied `localqueue-team-a.yaml` and `job-sample-workload.yaml` — confirmed the Job's Workload was admitted and pods ran to completion.

### DIFF
--- a/kueue/test-files/deploy/clusterqueue-team-a.yaml
+++ b/kueue/test-files/deploy/clusterqueue-team-a.yaml
@@ -13,13 +13,13 @@ spec:
         - cpu
         - memory
       flavors:
-        - name: default
+        - name: default-flavor
           resources:
             - name: cpu
               nominalQuota: "8"
             - name: memory
               nominalQuota: 32Gi
-        - name: spot
+        - name: spot-flavor
           resources:
             - name: cpu
               nominalQuota: "16"


Ralthos (commented):
Traced the sequence in the README and this holds up.

`resourceflavor-default.yaml` and `resourceflavor-spot.yaml` create `default-flavor` and
`spot-flavor`, while the ClusterQueue named `default` and `spot`, so following the documented
apply order leaves `team-a` pointing at flavors that do not exist.

The diff also stops in the right place. Line 10 is `kubernetes.io/metadata.name: default` inside
the `namespaceSelector`, which names the `default` namespace and not a flavor. Leaving it alone
is what keeps the selector matching the namespace the Job runs in.

The rest of the chain still lines up: `localqueue-team-a` targets `clusterQueue: team-a`, the Job
carries `kueue.x-k8s.io/queue-name: team-a-queue` in namespace `default`, and that namespace is
what the selector matches.

One thing outside this PR. The README applies `resourceflavor-topology.yaml`, but no ClusterQueue
in `test-files` references `topology-aware-flavor`, so nothing consumes it.

PR #1179, opened 2026-08-14
kueue: Open resource details in a side panel instead of full navigation

Adi-204:
## Summary
- Clicking a ClusterQueue, LocalQueue, ResourceFlavor, or Workload name (and cross-links between them, e.g. ClusterQueue → ResourceFlavor, LocalQueue → ClusterQueue, Workload → LocalQueue/ClusterQueue) previously navigated to a full-page detail view instead of opening in Headlamp's side panel.
- Root cause: Headlamp core's drawer/split-panel feature only activates for kinds present in its hardcoded `kindComponentMap` (see `frontend/src/components/resourceMap/details/KubeNodeDetails.tsx`). Kueue's CRDs aren't in that map, so `Link`'s default navigation always won.
- Fix follows the pattern already used by the `kyverno` and `kmesh` plugins in this repo: each list's name column (and each cross-link) now calls `Activity.launch({ location: 'split-right', ... })` with the same `Detail` component as content, so it opens as a side panel with a "Fullscreen" option to maximize.
- Each `Detail` component now accepts optional `name`/`namespace` props (falling back to `useParams()`), so the same component works both as a routed page and as `Activity` content.

## Test plan
- [x] `npm run tsc` passes with no errors
- [x] `npm run lint` passes with no errors/warnings
- [x] Manually verify in a running Headlamp instance: clicking names in ClusterQueues, LocalQueues, ResourceFlavors, and Workloads lists opens a side panel (not a full navigation), and the panel's menu has a working "Fullscreen" option
- [x] Manually verify cross-links (ClusterQueue's ResourceFlavor references, LocalQueue's ClusterQueue reference, Workload's LocalQueue/ClusterQueue references) also open in the side panel

### DIFF
--- a/kueue/src/components/clusterqueues/Detail.tsx
+++ b/kueue/src/components/clusterqueues/Detail.tsx
@@ -1,10 +1,12 @@
+import { Icon } from '@iconify/react';
+import { Activity } from '@kinvolk/headlamp-plugin/lib';
 import {
   ConditionsSection,
   DetailsGrid,
-  Link,
   SectionBox,
   SimpleTable,
 } from '@kinvolk/headlamp-plugin/lib/components/common';
+import { Link as MuiLink } from '@mui/material';
 import { useParams } from 'react-router-dom';
 import {
   ClusterQueue,
@@ -12,8 +14,8 @@ import {
   ResourceGroup,
   ResourceQuota,
 } from '../../resources/clusterQueue';
-import { kueueRouteNames } from '../../utils/kueueRoutes';
 import KueueAdminResourceAccess from '../common/KueueAdminResourceAccess';
+import { openResourceFlavorActivity } from '../resourceflavors/Detail';
 
 /** Flattened row rendered in the ClusterQueue resource groups table. */
 interface ResourceGroupRow {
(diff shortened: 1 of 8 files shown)


Ralthos (commented):
Checked the precedent this follows and it holds up. `kyverno/src/components/ClusterPolicyList.tsx`
uses the same `MuiLink component="button"` with `sx={{ textAlign: 'left' }}` and an
`Activity.launch` handler, so this matches the shape already in the repo.

Two things I checked and found clean. `index.tsx` is untouched, so the routes still resolve and a
full-page detail URL keeps working for anyone who has one bookmarked. The new cross-imports stay
acyclic, `workloads` to `localqueues` to `clusterqueues` to `resourceflavors`, so nothing is
half-initialised at module load.

Inside an Activity there is no route match, so `useParams` returns empty and anything read from it
is undefined. `props.namespace ?? params.namespace` is what covers that.

It also overlaps with #1258, which I have open on the same two detail files. That one scopes the
access check to the namespace and reads it the same way, so without the fallback here the check
inside a side panel would ask a cluster-wide question and deny the page, which is the bug in
#1199. They should land together, and whichever goes second will need to keep the other's line.

PR #1181, opened 2026-08-14
kueue: add ProvisioningRequestConfig list and detail views

Thesmoothengineer:
## What this PR does

Adds a new `ProvisioningRequestConfig` resource to the Kueue plugin, following the existing
patterns used by `ClusterQueue`, `LocalQueue`, `ResourceFlavor`, and `Workload`.
`ProvisioningRequestConfig` had no dedicated views in this plugin — it's the configuration
object referenced by `AdmissionCheck.spec.parameters` when using Kueue's autoscaler
integration (e.g. Cluster Autoscaler's ProvisioningRequest API).

- Adds `src/resources/provisioningRequestConfig.ts` — a `KubeObject`-based resource class
  typed against the real `ProvisioningRequestConfigSpec` fields (`provisioningClassName`,
  `managedResources`, `retryStrategy`, `podSetMergePolicy`), per the
  [Kueue API reference](https://kueue.sigs.k8s.io/docs/reference/kueue.v1beta2/#provisioningrequestconfig).
- Adds `src/resources/provisioningRequestConfigFormatters.ts` with exported, documented, pure
  functions: `renderProvisioningClassName`, `renderManagedResources`, `renderRetryStrategy`,
  `renderPodSetMergePolicy`.
- Adds `src/resources/provisioningRequestConfig.test.ts` with unit tests covering each
  formatter, including empty/undefined edge cases and a zero-value edge case for
  `backoffLimitCount`.
- Adds `src/components/provisioningrequestconfigs/List.tsx` and `Detail.tsx`, mirroring the
  `resourceflavors` components structure (cluster-scoped, using `KueueAdminResourceAccess`
  and `ResourceListView`/`DetailsGrid`).
- Registers the sidebar entry and routes in `index.tsx` and `utils/kueueRoutes.ts`.

## Testing

- `npm run test` — all 33 tests pass (11 new).
- `npm run build` — builds cleanly.
- Verified locally in a running Headlamp instance against a kind cluster: applied a real
  `ProvisioningRequestConfig` resource and confirmed both the List and Detail views correctly
  render the provisioning class, managed resources, and retry strategy.

## Why

`ProvisioningRequestConfig` is a core part of Kueue's autoscaler integration story
(referenced directly by `AdmissionCheck`), but had no dedicated view yet. This is a small,
additive, low-risk addition that gives users visibility into their provisioning configs
without touching any existing resource types.

### DIFF
--- a/kueue/src/components/provisioningrequestconfigs/Detail.tsx
+++ b/kueue/src/components/provisioningrequestconfigs/Detail.tsx
@@ -1,0 +1,44 @@
+import { DetailsGrid } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import { useParams } from 'react-router-dom';
+import { ProvisioningRequestConfig } from '../../resources/provisioningRequestConfig';
+import KueueAdminResourceAccess from '../common/KueueAdminResourceAccess';
+
+export default function ProvisioningRequestConfigDetail() {
+  const { name } = useParams<{ name: string }>();
+
+  return (
+    <KueueAdminResourceAccess
+      resourceClass={ProvisioningRequestConfig}
+      resourceLabel="ProvisioningRequestConfigs"
+      verb="get"
+    >
+      <DetailsGrid
+        resourceType={ProvisioningRequestConfig}
+        name={name}
+        withEvents
+        extraInfo={config =>
+          config
+            ? [
+                {
+                  name: 'Provisioning Class',
+                  value: config.provisioningClassName,
+                },
+                {
+                  name: 'Managed Resources',
+                  value: config.managedResourcesDisplay,
+                },
+                {
+                  name: 'Retry Strategy',
+                  value: config.retryStrategyDisplay,
+                },
+                {
+                  name: 'Pod Set Merge Policy',
+                  value: config.podSetMergePolicyDisplay,
+                },
+              ]
+            : []
+        }
+      />
+    </KueueAdminResourceAccess>
+  );
+}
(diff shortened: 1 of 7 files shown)


Ralthos (commented):
Splitting the formatters into their own module and testing them directly makes this easy to
check, and the `backoffLimitCount: 0` case in the tests is the one that usually gets missed.

The thing I would change before it merges is `renderManagedResources`. Kueue treats an empty
list as the inverse of what a dash suggests:

> managedResources contains the list of resources managed by the autoscaling.
>
> If empty, all resources are considered managed.

(`apis/kueue/v1beta2/provisioningrequestconfig_types.go`)

The interface in this PR carries that same note, but the formatter renders both `undefined` and
`[]` as `-`, which reads as "none". So the list column and the detail tile report that nothing
is managed on exactly the configs where everything is. `provisioningRequestConfig.test.ts` pins
the current behaviour, so the change is the formatter plus that one assertion. Wording is yours;
something like `All resources` would match the API.

Two smaller ones.

`podSetMergePolicy` is typed `string`. Upstream it is an enum with two values,
`IdenticalPodTemplates` and `IdenticalWorkloadSchedulingRequirements`, carrying
`+kubebuilder:validation:Enum`. A union type would catch a typo at compile time, and the test
already uses one of the literals.

Five of the new files end without a trailing newline: both components,
`provisioningRequestConfig.ts`, its formatters and its test. Everything else under `kueue/src`
ends with one.

I did check `backoffLimitCount` against the doc comment mentioning a default of 3, but it has
`+kubebuilder:default=3` upstream, so the API server fills it in and the rendering is fine.

PR #1185, opened 2026-08-15
argocd: Add Open in Argo CD Application action

Joshna907:
## Summary

This PR adds a safe **Open in Argo CD** action to the existing Argo CD Application detail header. The Argo icon opens the exact Application in the configured native Argo CD UI in a new browser tab.

## Why

Headlamp provides Kubernetes-native visibility and operations. Argo CD's own UI remains the right place for advanced diffs, diagnostics, and resource actions. This gives operators a safe direct path from an Application in Headlamp to that same Application in Argo CD.

## Changes

- Added optional `status.controllerNamespace` metadata and an Application getter.
- Reads only the `argocd-cm` ConfigMap using Headlamp's public ConfigMap API.
- Looks up the ConfigMap in `status.controllerNamespace`, falling back to the Application namespace.
- Shows the action only when `argocd-cm.data.url` is a valid absolute HTTP or HTTPS URL.
- Safely builds `/applications/<namespace>/<name>` URLs, preserving configured installation paths and encoding names.
- Rejects unsafe, malformed, relative, credential-bearing, or missing-host URLs.
- Opens the external UI with `noopener,noreferrer`.
- Documents the optional least-privilege ConfigMap permission and adds a fake local test manifest.

## Validation

- All 6 test files and 45 tests passed.
- Production build passed.

## Manual test

- Configured a local Argo CD UI URL.
- Confirmed the Argo icon appears beside Sync and Refresh on the Application detail page.
- Confirmed it opens the exact `guestbook` Application in the native Argo CD UI.
- Confirmed native Argo CD authentication, application tree, and advanced actions remain handled by Argo CD itself.

## Permissions

The optional action requires only:

```yaml
- apiGroups: [""]
  resources: ["configmaps"]
  resourceNames: ["argocd-cm"]
  verbs: ["get"]
```

When the ConfigMap is unavailable, forbidden, missing a URL, or contains an invalid URL, the action is simply hidden.

### DIFF
--- a/argocd/README.md
+++ b/argocd/README.md
@@ -10,6 +10,7 @@ To learn more about Argo CD, see the [Argo CD Getting Started Guide](https://arg
 - **Application Detail View** — Displays detailed metadata for a single Application: project, source, destination, sync status, health status, and Kubernetes events.
 - **Sync Action** — Triggers a sync by patching the Application's `.operation` field via the Kubernetes API. The Argo CD application-controller picks this up exactly as it would from the Argo CD UI.
 - **Refresh Action** — Requests a refresh by setting the `argocd.argoproj.io/refresh` annotation (supports `normal` and `hard` refresh types).
+- **Open in Argo CD** — Opens the current Application in the configured Argo CD web UI when `argocd-cm.data.url` is available.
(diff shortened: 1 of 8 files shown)


illume (approved):
🎉 thanks!

PR #1187, opened 2026-08-15
kueue: add MultiKueueConfig list and detail views

Thesmoothengineer:
## What this PR does

Adds initial Headlamp UI support for Kueue's `MultiKueueConfig` resource, following the same
pattern established by `ClusterQueue`, `LocalQueue`, `ResourceFlavor`, and `Workload`:

- `src/resources/multiKueueConfig.ts` — `MultiKueueConfig` resource class with a `spec.clusters`
  getter.
- `src/resources/multiKueueConfigFormatters.ts` — exported, documented, pure formatter functions
  for rendering the cluster list and cluster count.
- `src/resources/multiKueueConfig.test.ts` — unit tests for the formatters.
- `src/components/multikueueconfigs/List.tsx` and `Detail.tsx` — list and detail views.
- Registers a new `MultiKueueConfigs` sidebar entry and routes in `index.tsx` /
  `kueueRoutes.ts`.

## Testing

- `npm run test` — all 27 tests pass (5 new).
- `npm run build` — builds cleanly.
- Verified locally in a running Headlamp instance against a kind cluster with a real
  `MultiKueueConfig` resource — both the list and detail views render the cluster names and
  cluster count correctly.

## Why

MultiKueue is called out as a potential focus area for this term. This PR adds basic
visibility into `MultiKueueConfig` resources as a foundation that later work (e.g. manager/
worker cluster relationships, connectivity, workload routing) can build on.

### DIFF
--- a/kueue/src/components/multikueueconfigs/Detail.tsx
+++ b/kueue/src/components/multikueueconfigs/Detail.tsx
@@ -1,0 +1,36 @@
+import { DetailsGrid } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import { useParams } from 'react-router-dom';
+import { MultiKueueConfig } from '../../resources/multiKueueConfig';
+import KueueAdminResourceAccess from '../common/KueueAdminResourceAccess';
+
+export default function MultiKueueConfigDetail() {
+  const { name } = useParams<{ name: string }>();
+
+  return (
+    <KueueAdminResourceAccess
+      resourceClass={MultiKueueConfig}
+      resourceLabel="MultiKueueConfigs"
+      verb="get"
+    >
+      <DetailsGrid
+        resourceType={MultiKueueConfig}
+        name={name}
+        withEvents
+        extraInfo={config =>
+          config
+            ? [
+                {
+                  name: 'Clusters',
+                  value: config.clustersDisplay,
+                },
+                {
+                  name: 'Cluster Count',
+                  value: config.clusterCount,
+                },
+              ]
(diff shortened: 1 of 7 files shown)


Ralthos (commented):
The formatters and their tests read cleanly, and having `renderClusterCount` return a number
keeps that column sortable.

One thing about the clusters list. Upstream, the order is meaningful:

> clusters is a list of MultiKueueClusters names where the workloads from the ClusterQueue
> should be distributed.
>
> The order of the list is significant: the Incremental dispatcher nominates clusters following
> this order, so the most preferred clusters should be listed first.

(`apis/kueue/v1beta2/multikueue_types.go`)

The formatter's doc comment and the "preserving order" test both track that, but the rendering
does not surface it. A column labelled `Clusters` holding
`worker-onprem, worker-aws, worker-gcp` reads as a set, so an operator cannot tell the first
entry is the preferred one. A label like `Clusters (priority order)`, or numbering the entries
on the detail tile, would carry it through. If the label does that, the separate `Cluster Count`
column has less to do, with the list sitting right beside it.

The empty-clusters branches are unreachable. `clusters` is `+required` with
`+kubebuilder:validation:MinItems=1`, so the API server rejects a config with none and
`Cluster Count` can never render 0. Harmless as a guard, though the `-` placeholder describes a
state that cannot exist.

`List.tsx` also has a trailing space after its closing brace, and four of the new files end
without a newline. Everything else under `kueue/src` ends with one.

PR #1193, opened 2026-08-16
fix(kueue): resolve ClusterQueue cohort and admission checks for v1beta1

### LINKED ISSUE #1150: kueue: ClusterQueue Cohort and Admission Checks show "-" on clusters serving only v1beta1
### Description

`kueueApiVersions` registers two API versions, and `kueueApi.test.ts` asserts that
this is deliberate:

    export const kueueApiVersions = [`${kueueApiGroup}/v1beta2`, `${kueueApiGroup}/v1beta1`];

Headlamp honours that fallback. `KubeObject.apiEndpoint` maps the array into
per-version factory arguments, and `multipleApiFactory` documents its behaviour as
"It first tries the first endpoint, then the second, and so on until it gets a
successful response." So on a cluster where only `kueue.x-k8s.io/v1beta1` is
served, the plugin lists ClusterQueues successfully via v1beta1.

The `ClusterQueueSpec` type, however, was written from the v1beta2 reference alone
— every `@see` link points at `kueue.v1beta2` — so two fields that Kueue renamed
or restructured between the versions are never read.

Kueue's own conversion code names both:

    func Convert_v1beta1_ClusterQueueSpec_To_v1beta2_ClusterQueueSpec(...) {
        out.CohortName = v1beta2.CohortReference(in.Cohort)
        ...
        out.AdmissionChecksStrategy = &v1beta2.AdmissionChecksStrategy{...}

| Field | v1beta1 | v1beta2 |
|---|---|---|
| Cohort | `spec.cohort` | `spec.cohortName` |
| Admission checks | `spec.admissionChecks: string[]` | `spec.admissionChecksStrategy.admissionChecks[].name` |

`get cohortName()` reads only `spec.cohortName`, and `admissionChecksDisplay`
passes only `spec.admissionChecksStrategy` to `renderAdmissionChecks`.

### Steps to reproduce

1. Connect Headlamp to a cluster running a Kueue release that serves only
   `kueue.x-k8s.io/v1beta1`.
2. Apply a ClusterQueue with a cohort and an admission check:

       apiVersion: kueue.x-k8s.io/v1beta1
       kind: ClusterQueue
       metadata:
         name: team-a
       spec:
         cohort: research
         admissionChecks:
         - prov-request
         namespaceSelector: {}

3. Open Kueue > ClusterQueues, and the detail page for `team-a`.

### Actual

The Cohort column in the list and the Cohort row on the detail page both read `-`.
The Admission Checks row reads `-`. Measured against the current getters:

    { "cohort": "-", "admissionChecks": "-" }

### Expected

    { "cohort": "research", "admissionChecks": "prov-request (all flavors)" }

Cohorts are the mechanism by which ClusterQueues borrow unused quota from each
other, so this hides a relationship rather than an optional detail. Admission
checks gate whether a workload can be admitted at all.

### Proposed fix

Resolve both shapes before rendering. The v1beta1 admission check form carries no
per-flavor scoping, so those entries render as applying to all flavors.

I have this working with tests. The plugin's resource classes can't be unit tested
under Vitest because they import `KubeObject`, a build-time virtual module, which
is why the existing tests only exercise the pure `*Formatters` modules. So the
resolvers belong in a pure module the class delegates to, matching that structure.
Happy to open a PR.

### Note

There is a third v1beta1/v1beta2 difference I have not treated as a bug:
`flavorFungibility.whenCanBorrow` and `whenCanPreempt` used `Borrow` and `Preempt`
in v1beta1, which v1beta2 renamed to `MayStopSearch`. The plugin's TypeScript union
lists only the v1beta2 values, but the renderer prints whatever string the API
returned, so the display stays correct. That is a type-completeness gap.

Kueue also ships conversion files for LocalQueue, Workload and MultiKueue, so the
same class of gap may exist there. Those files are currently touched by open PRs
(#1068, #1046), so I have left them alone.

### Environment

- kueue plugin, `main` @ (current)
- @kinvolk/headlamp-plugin 0.14.0

Prathamesh09-coder:
## Summary

The Kueue `ClusterQueue` resource class only reads the v1beta2
field shapes for Cohort and Admission Checks.

When Headlamp falls back to the v1beta1 API, valid values in
`spec.cohort` and `spec.admissionChecks` were not displayed.

Fixes #1150

## Changes

- Add v1beta1 `cohort` and `admissionChecks` fields to `ClusterQueueSpec`
- Add `resolveCohortName()` with v1beta2 → v1beta1 fallback
- Add `resolveAndRenderAdmissionChecks()` with v1beta2 → v1beta1 fallback
- Preserve existing v1beta2 admission-check rendering
- Update `ClusterQueue` getters to use the new resolvers
- Add unit tests for both API versions and fallback behavior

## Testing

- `cd kueue && ./node_modules/.bin/vitest run --reporter=verbose`
- 25/25 tests passing across 4 test files
- Existing v1beta2 behavior remains covered

### DIFF
--- a/kueue/src/components/clusterqueues/Detail.tsx
+++ b/kueue/src/components/clusterqueues/Detail.tsx
@@ -12,6 +12,7 @@ import {
   ResourceGroup,
   ResourceQuota,
 } from '../../resources/clusterQueue';
+import { getAdmissionChecksStrategy } from '../../resources/clusterQueueCompat';
 import { kueueRouteNames } from '../../utils/kueueRoutes';
 import KueueAdminResourceAccess from '../common/KueueAdminResourceAccess';
 
@@ -130,7 +131,7 @@ function getFlavorUsageRows(flavorUsage: FlavorUsage[] = []): FlavorUsageRow[] {
 /** Extract admission checks and their flavor scope from the ClusterQueue spec. */
 function getAdmissionCheckRows(clusterQueue: ClusterQueue): AdmissionCheckRow[] {
   return (
-    clusterQueue.spec.admissionChecksStrategy?.admissionChecks?.map(check => ({
+    getAdmissionChecksStrategy(clusterQueue.spec)?.admissionChecks?.map(check => ({
       name: check.name,
       flavors: check.onFlavors || [],
     })) || []
--- a/kueue/src/resources/clusterQueue.test.ts
+++ b/kueue/src/resources/clusterQueue.test.ts
@@ -5,8 +5,6 @@ import {
   renderClusterQueueStatus,
   renderLabelSelector,
   renderResourceGroupsSummary,
-  resolveAndRenderAdmissionChecks,
-  resolveCohortName,
 } from './clusterQueueFormatters';
 
 describe('ClusterQueue formatters', () => {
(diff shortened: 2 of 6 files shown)


Ralthos (commented):
Putting the version resolution in its own module and leaving the formatters to render one shape is
the right split. The v1beta1 to v1beta2 conversion in `getAdmissionChecksStrategy` means the
existing rendering path did not have to learn about the older API at all.

The cohort side is complete. `cohortName` now goes through `getCohortName`, and both readers
(`clusterqueues/Detail.tsx:299` and `List.tsx:20`) use that getter, so they pick the fix up for
free.

The admission checks side is not, and it misses in the place the feature is most visible.
`clusterQueue.ts` routes the summary through the resolver:

```ts
return renderAdmissionChecks(getAdmissionChecksStrategy(this.spec));
```

but `clusterqueues/Detail.tsx:133` still reads the raw v1beta2 field:

```ts
function getAdmissionCheckRows(clusterQueue: ClusterQueue): AdmissionCheckRow[] {
  return (
    clusterQueue.spec.admissionChecksStrategy?.admissionChecks?.map(check => ({
      name: check.name,
      flavors: check.onFlavors || [],
    })) || []
  );
}
```

On a v1beta1 cluster, `admissionChecksStrategy` is absent and `spec.admissionChecks` holds the
names. So after this PR the ClusterQueue detail page shows the checks in the summary field and an
empty table underneath, for the same ClusterQueue, from the same spec. That is a more confusing
state than the current one, where both are empty and at least agree.

Passing the spec through the resolver fixes it in one line:

```ts
return (
  getAdmissionChecksStrategy(clusterQueue.spec)?.admissionChecks?.map(check => ({
```

Worth a test alongside the ones you have: the two v1beta1 cases in `clusterQueueCompat.test.ts`
pass because they call the resolver directly, and nothing currently fails if a caller forgets to.

Prathamesh09-coder:
Thanks for catching this call site. I've updated getAdmissionCheckRows() in Detail.tsx to use getAdmissionChecksStrategy(clusterQueue.spec), so both the summary and detail table now consume the same normalized admission-check representation.
I also added a regression test covering the v1beta1 flat admissionChecks representation at the Detail call site, so this won't be protected only by the resolver unit tests.
Full validation passes: TypeScript, lint, formatting, and the Kueue Vitest suite.

Ralthos:
The resolver changes look right, and the four new cases in `clusterQueueCompat.test.ts` cover the
precedence question well, including v1beta2 present but empty falling through to v1beta1.

I cannot see the `Detail.tsx` change though. At `dd32112` the PR still touches four files:

```
kueue/src/resources/clusterQueue.test.ts
kueue/src/resources/clusterQueue.ts
kueue/src/resources/clusterQueueCompat.test.ts
kueue/src/resources/clusterQueueCompat.ts
```

and `clusterqueues/Detail.tsx:133` on that same commit still reads the raw field:

```ts
function getAdmissionCheckRows(clusterQueue: ClusterQueue): AdmissionCheckRow[] {
  return (
    clusterQueue.spec.admissionChecksStrategy?.admissionChecks?.map(check => ({
```

I would guess the commit went to a different branch or has not been pushed, since the description
matches what is needed. Worth a look before this lands, because the summary field and the table
below it would still disagree on a v1beta1 cluster.

Prathamesh09-coder:
Thanks for catching this call site. I've updated getAdmissionCheckRows() in Detail.tsx to use getAdmissionChecksStrategy(clusterQueue.spec), so both the summary and detail table now consume the same normalized admission-check representation.

I also added a regression test covering the v1beta1 flat admissionChecks representation at the Detail call site, so this won't be protected only by the resolver unit tests.

Full validation passes: TypeScript, lint, formatting, and the Kueue Vitest suite.

PR #1195, opened 2026-08-16
fix(flux): clear the reconciliation timeout on unmount

### LINKED ISSUE #1006: flux: RemainingTimeDisplay never clears its timeout on unmount
## What happens

`RemainingTimeDisplay` in `flux/src/common/RemainingTimeDisplay.tsx` schedules a `setTimeout` to refresh the countdown, and keeps the handle in `timeoutRef`. The effect meant to clean that up on unmount looks like this:

```js
React.useEffect(() => {
  // Just to clear out any remaining timeouts when the component is unmounted
  if (!!timeoutRef.current) {
    window.clearInterval(timeoutRef.current);
    timeoutRef.current = undefined;
  }
}, []);
```

The clearing code is in the effect body instead of in a returned cleanup function. With an empty dependency array the body runs once on mount, at which point `timeoutRef.current` is still `undefined`, so the `if` is false and nothing happens. Because there is no `return`, React has no cleanup to call, and the effect never runs again. The comment describes what was intended, but the timeout is never cleared on unmount.

The pending timeout therefore survives unmount and fires `setTimeRemaining` on a component that is gone. Every Flux list row that shows a next reconciliation time has one of these, so navigating away from a list of any size leaks one timer per row.

There is a second, smaller problem in the same lines. The handle comes from `window.setTimeout` but is released with `window.clearInterval`. Browsers share one id space for timeouts and intervals so this happens to work, but it is misleading, and the same mismatch appears at line 141.

## Why it matters

This is the same class of problem as the auto refresh timers being cleaned up elsewhere in the repo, except here the guard was written and simply never wired up, so it has always been inactive.

## Steps to reproduce

1. Open a Flux list view with several resources that have a `spec.interval`, for example Kustomizations.
2. Navigate to a different page before the countdown ticks.
3. The scheduled timeout still fires and calls `setTimeRemaining` on the unmounted component.

## Suggested fix

Return the cleanup from the effect rather than running it in the body, and use `window.clearTimeout` to match `window.setTimeout`.

I have a fix ready and will open a PR alongside this issue.

Ahmad-Faraj:
`RemainingTimeDisplay` kept its cleanup in the effect body instead of returning it, so with `[]` deps it ran once on mount while the ref was still undefined and never again. The pending timeout outlived the component and fired `setTimeRemaining` after unmount, one per Flux list row showing a next reconciliation time.

Returns the cleanup now, and uses `clearTimeout` rather than `clearInterval` since the handle comes from `setTimeout`.

Added a test that renders the component, unmounts it and asserts the timeout is cleared. It fails without the change.

Fixes #1006

### DIFF
--- a/flux/src/common/RemainingTimeDisplay.test.tsx
+++ b/flux/src/common/RemainingTimeDisplay.test.tsx
@@ -65,11 +65,15 @@ describe('RemainingTimeDisplay', () => {
     const clearSpy = vi.spyOn(window, 'clearTimeout');
 
     const { unmount } = render(<RemainingTimeDisplay item={itemDueIn('5m')} />);
+
+    // The handle this component scheduled, which is the one unmount has to clear.
     expect(setSpy).toHaveBeenCalled();
+    const scheduled = setSpy.mock.results.at(-1)?.value;
+    expect(scheduled).toBeDefined();
 
     clearSpy.mockClear();
     unmount();
 
-    expect(clearSpy).toHaveBeenCalled();
+    expect(clearSpy).toHaveBeenCalledWith(scheduled);
   });
 });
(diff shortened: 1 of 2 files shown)


Ahmad-Faraj:
@illume any chance of a look at this one? Small, green, closes #1006.

Ahmad-Faraj:
This duplicates #1008, which was opened first on Aug 4.

PR #1197, opened 2026-08-17
argocd: Add safe Application rollback from Sync History

Joshna907:
## Summary

This PR adds a deliberate, Kubernetes-native rollback action to the existing Argo CD Application detail page. Operators can select an earlier usable deployment from Sync History, review its full revision and deployment time, and confirm before Headlamp submits the rollback operation.

Rollback is hidden when automated sync is enabled or no earlier complete deployment is available. It uses the existing Kubernetes patch permission and never modifies the Application spec or its configured Git target revision.

## Why Kubernetes-native rollback?

The plugin already performs Sync and Refresh through the Kubernetes API. Rollback follows the same model by patching only the Application operation field. Argo CD executes the operation asynchronously, while Git remains the source of truth.

## Changes

- Model complete single-source and multi-source rollback history snapshots.
- Sort history newest-first and exclude the current deployment.
- Reject malformed, incomplete, or duplicate history records.
- Add a rollback API operation that preserves historical sources, revisions, sync options, and the apply strategy.
- Add an RBAC-protected Rollback header action and confirmation dialog.
- Require an explicit history selection; no revision is preselected.
- Prevent duplicate submissions and conflicting Sync, Refresh, or Rollback actions while loading.
- Keep the dialog open after failure and close it after a successful request.
- Improve warning readability in Headlamp's dark theme.
- Document behavior, limitations, and required permissions.

## Safety

- Patches only the top-level operation field.
- Never patches spec.source.targetRevision or spec.sources.
- Uses prune: false and dryRun: false.
- Hides rollback while automated sync is enabled.
- Hides rollback when no complete earlier history snapshot exists.
- Uses the existing Kubernetes patch RBAC check and error handling.
- Reports that rollback was triggered, not completed, because Argo CD performs it asynchronously.

## Validation

- All 13 test files and 118 tests passed.
- Production build passed.
- All four commits include Signed-off-by lines.

## Steps to test

1. Open a manually synced Argo CD Application containing at least two successful history entries.
2. Confirm the Rollback action appears in the Application detail header.
3. Open Rollback and confirm no deployment is selected automatically.
4. Select an earlier deployment and verify its complete revision and deployment time are shown.
5. Cancel and confirm no Kubernetes request is sent.
6. Reopen the dialog, select a deployment, and confirm rollback.
7. Verify Argo CD creates a sync operation for the historical source snapshot.
8. Verify spec.source.targetRevision or spec.sources remains unchanged.
9. Enable automated sync and confirm the Rollback action disappears.
10. Verify a user without Application patch permission cannot see the action.

## Scope

This PR does not call the Argo CD REST API, change Git, enable pruning, or claim that a triggered rollback has finished.

### DIFF
--- a/argocd/README.md
+++ b/argocd/README.md
@@ -12,6 +12,7 @@ To learn more about Argo CD, see the [Argo CD Getting Started Guide](https://arg
 - **Application Detail View** — Displays detailed metadata for a single Application: project, source, destination, sync status, health status, and Kubernetes events.
 - **Sync Action** — Triggers a sync by patching the Application's `.operation` field via the Kubernetes API. The Argo CD application-controller picks this up exactly as it would from the Argo CD UI.
 - **Refresh Action** — Requests a refresh by setting the `argocd.argoproj.io/refresh` annotation (supports `normal` and `hard` refresh types).
+- **Rollback Action** — Lets an authorized user deliberately select a complete earlier deployment from Sync History and trigger it as a one-time Argo CD operation without changing the Application's configured Git revision.
 - **Open in Argo CD** — Opens the current Application in the configured Argo CD web UI when `argocd-cm.data.url` is available.
 - **Argo CD Sidebar Icon** — Registers the official Argo CD octopus logo as an offline Iconify icon (CSP-safe, no external fetch).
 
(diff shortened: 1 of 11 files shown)


Ralthos (commented):
The shape of this is careful in the ways that matter for a destructive action. Patching only
`.operation` and never `spec.source`, requiring a complete history snapshot before building the
request, refusing a multi-source entry whose `revisions` and `sources` lengths disagree, and
starting the dialog with nothing selected are all the right defaults. `isUsableHistoryEntry`
rejecting an unparseable `deployedAt` is a nice touch, since a rollback target you cannot date is
one a person cannot verify.

One defect, in the predicate the whole safeguard rests on.

```ts
export function isAutomatedSyncEnabled(spec: ArgoApplicationSpec): boolean {
  const automated = spec.syncPolicy?.automated;
  return automated !== undefined && automated.enabled !== false;
}
```

The optional chain stops at `syncPolicy`, so `automated` is only guarded against the parent being
absent. If `syncPolicy` exists and `automated` is `null`, the second half dereferences it:

```
no syncPolicy          -> false
automated undefined    -> false
automated {}           -> true
enabled false          -> false
automated null         -> THROWS: TypeError: Cannot read properties of null (reading 'enabled')
```

`automated: null` is what YAML produces from `automated:` with nothing under it, and from a Helm
template whose value renders empty. The declared type says `automated?: { ... }`, but that is a
compile-time claim about data that arrives from the API server at runtime, so it does not rule the
case out.

The blast radius is larger than the button. `isAutomatedSyncEnabled` is reached through the
`app.isAutomatedSyncEnabled` getter while the actions are being built, so the throw happens during
render of the Application detail page, not at click time. An Application with that spec takes the
whole page down.

`!=` covers null and undefined together:

```ts
return automated != null && automated.enabled !== false;
```

Worth a row in the table at `application.test.ts:1087`, which currently maps `undefined`, `{}` and
`{ enabled: false }`. The helper's own signature is `(automated?: { enabled?: boolean | null })`,
so `null` cannot be passed to it as written and the case cannot be expressed until that widens to
`| null`.

For what it is worth, `null` failing closed would also be the wrong answer if it did not throw.
Argo CD treats a null `automated` as no automation, so the safe reading is manual sync, which is
the branch that shows Rollback.

illume (approved):
🎉 thanks!

PR #1198, opened 2026-08-17
argocd: Add read-only ApplicationSets and Project GitOps views

Joshna907:
## Summary

This PR adds read-only Argo CD ApplicationSet visibility and Argo CD GitOps information to Headlamp Projects.

ApplicationSets receive CRD-aware list and detail pages with safe generator summaries, template information, controller conditions, generated Application counts, and live generated Application status. Headlamp Project pages receive an Argo CD overview and an Applications tab for Applications that explicitly deploy into the Project's local namespaces.

No ApplicationSet mutation or new Argo CD action is introduced.

## Why read-only first?

ApplicationSets can generate and manage many Applications. Inventory, ownership, and status visibility provide immediate operational value without exposing risky create, edit, delete, preview, sync, or generator mutation controls.

Project integration answers a separate operational question: which Argo CD Applications deploy into the namespaces owned by this Headlamp Project? Membership is based on destination cluster and namespace, not where the Application custom resource is stored.

### ApplicationSets

- Add an ArgoApplicationSet KubeObject model with typed template, generator, condition, health, resource-count, and progressive-sync accessors.
- Add concise summaries for List, Clusters, Git, SCM Provider, Pull Request, Cluster Decision Resource, Plugin, Matrix, and Merge generators.
- Avoid rendering secret references, credentials, tokens, or arbitrary plugin inputs.
- Add a read-only ApplicationSet list with generator, template project, destination, generated count, and age columns.
- Add a read-only detail page with template sources and destination, generator summaries, generated Applications, controller conditions, and Events.
- Resolve generated Applications only through exact same-cluster, same-namespace owner references matching the ApplicationSet UID.
- Preserve controller counts when live Application listing is forbidden or unavailable.

### Headlamp Projects

- Register an Argo CD summary on Project overview pages.
- Register an Argo CD Applications Project detail tab.
- Include only Applications loaded from a Project cluster that explicitly target in-cluster or https://kubernetes.default.svc and deploy into one of the Project namespaces.
- Exclude remote or unverified destinations and Applications without an explicit destination namespace.
- Preserve cluster context in Application links and deduplicate by cluster, Application namespace, and name.
- Show Total, Synced, Healthy, and Needs Attention counts.
- Preserve successful results and show a partial-results warning when only some Project clusters fail.

### Registration and CRD handling

- Refine the cached CRD guard to track exact Argo CD CRD names per cluster.
- Hide only the sidebar entry whose CRD is unavailable.
- Keep the Argo CD parent visible when any supported Argo CD CRD exists.
- Add ApplicationSet routes, sidebar navigation, Project integrations, and offline Argo icon registrations.
- Keep existing Applications, AppProjects, Resource Tree Map, namespace insights, API availability, and Open in Argo CD behavior intact.

## Safety

- No ApplicationSet create, edit, delete, sync, preview, or generator controls.
- No Kubernetes mutation requests.
- Generated Applications are never inferred from names, labels, or templates.
- Remote Applications never become Project members based only on a matching namespace name.
- Missing destination namespaces are excluded instead of being assumed to be default.
- No network-loaded icons or Argo CD API tokens are introduced.

## Validation

- All 15 test files and 112 tests passed.
- Production build passed.
- All six commits include Signed-off-by lines.

## Steps to test

1. Install an ApplicationSet CRD and create a List-generator ApplicationSet producing at least two Applications.
2. Open Argo CD > ApplicationSets and verify the read-only inventory columns.
3. Open an ApplicationSet and verify template information, generator summaries, generated Applications, conditions, and Events.
4. Select a generated Application and confirm its existing cluster-aware Argo CD Application detail page opens.
5. Confirm no create, edit, delete, sync, preview, or generator actions are present.
6. Remove or deny ApplicationSet list access and verify the page handles the unavailable state without mutation controls.
7. Open a Headlamp Project containing the generated destination namespace.
8. Verify the overview counts and Argo CD Applications tab include the local generated Applications.
9. Create or inspect a remote Application targeting the same namespace name and verify it is excluded.
10. If using a multi-cluster Project, verify cluster identity is shown and partial cluster failures retain successful results.
11. On a cluster without the ApplicationSet CRD, verify only the ApplicationSets sidebar entry is hidden.

## Required access

This feature uses Kubernetes read access only: get, list, and watch for ApplicationSets and Applications.

### DIFF
--- a/argocd/README.md
+++ b/argocd/README.md
@@ -14,6 +14,8 @@ To learn more about Argo CD, see the [Argo CD Getting Started Guide](https://arg
 - **Refresh Action** — Requests a refresh by setting the `argocd.argoproj.io/refresh` annotation (supports `normal` and `hard` refresh types).
 - **Rollback Action** — Lets an authorized user deliberately select a complete earlier deployment from Sync History and trigger it as a one-time Argo CD operation without changing the Application's configured Git revision.
 - **Open in Argo CD** — Opens the current Application in the configured Argo CD web UI when `argocd-cm.data.url` is available.
+- **ApplicationSet Inventory** — Adds read-only ApplicationSet list and detail pages with safe generator summaries, template information, controller conditions, and live generated Application status.
+- **Headlamp Project GitOps Views** — Adds an Argo CD summary and Applications tab to Headlamp Projects for Applications that explicitly deploy into the Project's local namespaces.
 - **Argo CD Sidebar Icon** — Registers the official Argo CD octopus logo as an offline Iconify icon (CSP-safe, no external fetch).
 
 ### API availability scope
@@ -21,6 +23,18 @@ To learn more about Argo CD, see the [Argo CD Getting Started Guide](https://arg
 The Managed Resources table checks whether the selected Kubernetes cluster serves the exact API group, version, and kind reported by Argo CD. For example, it can show that a resource uses an API version that the cluster no longer serves.
 
 This is an **API availability check**, not a full controller compatibility test. A matching API does not prove that an installed controller implements every field or feature, and it does not prove that application traffic works. Remote or unverified Application destinations are not queried and are shown as `Not checked — remote`.
+
+### Read-only ApplicationSet behavior
+
+ApplicationSet pages intentionally have no create, edit, delete, sync, preview, or generator controls. Generated Applications are linked only when a live Application has an exact Kubernetes owner reference to the ApplicationSet UID in the same namespace and cluster. Names, labels, and naming conventions are not used as ownership evidence.
+
+Generator summaries show only safe identifiers such as generator type, repository URL, path count, provider organization, or plugin ConfigMap name. Secret references, tokens, credentials, and arbitrary plugin inputs are not rendered.
+
+The ApplicationSets sidebar entry is hidden when `applicationsets.argoproj.io` is not installed. Missing ApplicationSet support does not hide the existing Applications or Projects pages.
+
+### Headlamp Project matching
+
+An Argo CD Application appears in a Headlamp Project only when it was loaded from one of the Project's clusters, explicitly targets `in-cluster` or `https://kubernetes.default.svc`, and has an explicit destination namespace listed in the Project. Remote or unverified destinations and Applications without a destination namespace are excluded.
 
 ### Why the Kubernetes API instead of the Argo CD REST API?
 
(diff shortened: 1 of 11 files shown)


Ralthos (commented):
The resolution work here is the strongest part. `isGeneratedApplication` requires a UID, the same
cluster, the same namespace, and an owner reference matching group, kind and UID, so it never
falls back to a name. `getProjectApplications` is equally strict: explicit local destination,
explicit destination namespace, membership checked against the Project's own sets, deduplicated
by cluster and namespace and name, then sorted deterministically. Both refuse to guess, and the
README says out loud which Applications are excluded and why. That is the right posture for a
read-only view.

One gap, and it is the same question this PR answers well everywhere else.

The section computes two numbers that can disagree:

```ts
const verifiedLiveCount = applications
  ? getGeneratedApplications(applications, applicationSet).length
  : undefined;
const controllerCount = getGeneratedApplicationCount(applicationSet.status, verifiedLiveCount);
```

When nothing is verifiable, the mismatch is stated plainly, and this is the sentence I would keep:

```
`The controller reports ${controllerCount} generated Applications, but none are currently
available to list.`
```

When the mismatch is partial, nothing is said. If the controller reports five and three carry a
matching owner reference in a listable namespace, the table renders three rows and stops. A
reader gets a definitive-looking list and no hint that the controller believes there are more.

That is reachable without anything being broken. A user whose Application list is scoped by RBAC
to a subset of namespaces sees exactly this, and so does anyone looking while the controller is
mid-reconcile. It also runs the other way: a stale `status.resourcesCount` lower than the number
of live owned Applications leaves the table showing more rows than the count above it.

The fix is the branch you already wrote, applied to inequality rather than to zero:

```ts
controllerCount !== undefined && controllerCount !== generatedApplications.length
```

and the same wording, naming both numbers. The distinction worth preserving is that the table is
what Headlamp verified and the count is what the controller claims, which is exactly the framing
the empty case uses.

Smaller point on `getGeneratedApplicationCount`: `status.resources` and `status.resourcesCount`
are read as the controller's own tally, ahead of the verified count. That ordering is right,
since the controller knows about Applications this user may not be able to list. Worth a comment
saying so, because as written it looks like the less trustworthy source is winning.

Joshna907:
ptal @illume

illume (approved):
🎉 thanks!

PR #1200, opened 2026-08-17
 fix(kueue): resolve namespace-scoped RBAC blocks for Workloads and L…

### LINKED ISSUE #1199: Kueue: plugin blocks non-admins from viewing namespaced Workloads/LocalQueues
## Describe the bug
The Kueue plugin prevents non-admin users from being able to view namespaced resources (Workloads, LocalQueues) in the UI, even though they do have the appropriate RBAC permissions to get and list those resources in their namespace. 

**Expected outcome:** 
Users who have RBAC permissions on a namespace should be able to view the lists of `Workloads` and `LocalQueues`, and Headlamp should automatically filter and show the resources for the namespaces to which they have access.

## To Reproduce
Steps to reproduce the bug:
1. Make a namespace and a service account, for example, `data-scientist` in the `data-sci` namespace.
2. Make a `Role` that gives `get`, `list`, and `watch` permissions on `workloads.kueue.x-k8s.io` and `localqueues.kueue.x-k8s.io`.
3. Use a namespace-scoped RoleBinding in the data-sci namespace to bind the Role to the service account.
4. Log in to Headlamp using a token that has been generated for the `data-scientist` service account.
5. In the left sidebar, click on **Kueue**.
6. Click on the tab labelled **Workloads**.
7. Look at the error message: "The Kubernetes credentials you currently have are not authorized to list this page."

## Environment (please provide info about your environment):
 - Installation type: Linux Desktop App(using flatpak)
 - Headlamp Version: latest `main`
 Other: Kueue Plugin (version 0.1.0-alpha)

## Can you fix this issue?
I'll submit a pull request to fix this bug.

**Screenshots:**

Image

Image

harsh-aghara:
## Summary

This PR fixes namespace-scoped RBAC blocks for Kueue Workloads and LocalQueues by ensuring localized access checks are used instead of strictly enforcing cluster-scoped admin access.

## Related Issue

Fixes #1199 

## Changes

- **Updated** `KueueAdminResourceAccess` to accept and pass an optional `namespace` prop.
- **Fixed** RBAC checks on LocalQueue and Workload `Detail` views by passing the `namespace` to `KueueAdminResourceAccess`.
- **Refactored** LocalQueue and Workload `List` views by removing the cluster-scoped `KueueAdminResourceAccess` wrapper, allowing Headlamp's native `ResourceListView` to natively handle localized RBAC.

## Steps to Test

1. Configure a user or ServiceAccount with namespace-scoped permissions (no cluster-wide access) for Kueue Workloads and LocalQueues.
2. Log into Headlamp with that user/ServiceAccount.
3. Navigate to the Kueue Workloads and LocalQueues list views and verify they load successfully without cluster-wide RBAC block errors.
4. Navigate to the detail view of a Workload and LocalQueue within the permitted namespace and verify access is correctly granted.

## Screenshots

image
image

### DIFF
--- a/kueue/src/components/common/KueueAdminResourceAccess.tsx
+++ b/kueue/src/components/common/KueueAdminResourceAccess.tsx
@@ -31,11 +31,21 @@ export default function KueueAdminResourceAccess({
   children,
 }: KueueAdminResourceAccessProps) {
   const [allowed, setAllowed] = useState<boolean | null>(null);
+  const [error, setError] = useState<Error | null>(null);
 
   return (
     <>
-      {allowed === null && <Loader title={`Checking access to Kueue ${resourceLabel}`} />}
-      {allowed === false && (
+      {allowed === null && error === null && (
+        <Loader title={`Checking access to Kueue ${resourceLabel}`} />
+      )}
+      {error !== null && (
+        <SectionBox title={`Kueue ${resourceLabel}`}>
+          <EmptyContent color="error.main">
+            {`Failed to check access: ${error.message}`}
+          </EmptyContent>
+        </SectionBox>
+      )}
+      {allowed === false && error === null && (
         <SectionBox title={`Kueue ${resourceLabel}`}>
           <EmptyContent color="text.secondary">
             {`${accessDescription} Your current Kubernetes credentials are not authorized to ${
@@ -49,7 +59,7 @@ export default function KueueAdminResourceAccess({
         authVerb={verb}
         namespace={namespace}
         onAuthResult={result => setAllowed(result.allowed)}
-        onError={() => setAllowed(false)}
+        onError={err => setError(err)}
       >
         {children}
       </AuthVisible>
--- a/kueue/src/components/common/KueueAdminResourceAccess.tsx
+++ b/kueue/src/components/common/KueueAdminResourceAccess.tsx
@@ -14,6 +14,8 @@ interface KueueAdminResourceAccessProps {
   resourceLabel: string;
   /** Kubernetes verb to verify before rendering the page. */
   verb: 'get' | 'list';
+  /** Optional namespace for namespace-scoped resources. */
+  namespace?: string;
   /** Optional sentence describing the resource scope in denied messages. */
   accessDescription?: string;
   /** Page content to render after the user is authorized. */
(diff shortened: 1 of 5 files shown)


Ralthos (commented):
The diagnosis is right. `AuthVisible` with `item={resourceClass}` and no namespace asks whether
the user can list Workloads cluster-wide, and a user with RBAC in one namespace answers no to
that while being perfectly able to list their own. Gating the page on it locks out exactly the
people the page is for.

Two things about the fix.

**The check is only removed from the List views, and the Detail views still have it.**
`workloads/Detail.tsx:460` and `localqueues/Detail.tsx:42` still wrap in
`KueueAdminResourceAccess`, with the same cluster-scoped question. So after this PR a
namespace-scoped user gets through the list, clicks a Workload, and is told they are not
authorized to view it. The inconsistency is worse than the original bug, because the list now
promises something the detail page refuses.

**`AuthVisible` already takes a namespace**, and its own doc comment describes this exact case:

```
/** The namespace for which we're checking the permission, if applied. This is mostly useful
    when checking "creation" using a resource class, instead of an instance. */
namespace?: string;
```

Threading `namespace` through `KueueAdminResourceAccess` into `AuthVisible` fixes all four call
sites with one change and keeps the thing the check was buying: a user with no access at all
sees "you are not authorized" instead of an empty list.

That distinction is worth keeping. Without the check, "no Workloads in this namespace" and "not
allowed to see Workloads in this namespace" render identically, and the second one is the case
where the operator needs to be told something.

One smaller thing if you keep the component: `onError={() => setAllowed(false)}` treats a failed
`SelfSubjectAccessReview` as a denial. A review that could not be performed is not a no, and on a
transient failure the page will claim a permissions problem that does not exist. Same shape as
what came up on #1206 and #972.

harsh-aghara:
@Ralthos Thanks for the review,
I have pushed a commit that fixes all the addressed issues.

Ralthos:
Checked, and this went further than what I raised. Threading `namespace` into `AuthVisible`
rather than dropping the guard keeps the thing that made it worth having: a user with no access
still gets told so instead of reading an empty list as "no Workloads here".

Both Detail files now pass it too, so the list and the detail page ask the same question. That was
the inconsistency I was most worried about.

The error handling is the part I would call out as better than my suggestion. Splitting `error`
from `allowed` gives three distinct states instead of two:

```
allowed === null && error === null   -> checking
error !== null                       -> "Failed to check access: ..."
allowed === false && error === null  -> not authorized
```

A `SelfSubjectAccessReview` that could not be performed now says so, and does not borrow the
wording for a denial. That is the distinction I was asking for and it reads cleanly.

PR #1205, opened 2026-08-17
fix(prometheus): formatBytes breaks above 1TB

magic-peach:
formatBytes indexes into a fixed units array (B/KB/MB/GB/TB) using floor(log1024(bytes)), but never clamps it. Once bytes crosses 1024^5 the index goes past TB and units[i] is undefined, so the chart tooltip prints stuff like "1.00undefined" instead of a number.

This shows up in the Network/Disk/Filesystem charts, whose values are cumulative counters and can plausibly cross that line on a node that's been up a while.

Clamped the index to the last unit instead of adding PB/EB entries, same as it already does for the 0-byte case. Added a few cases to util.test.ts.

### DIFF
--- a/prometheus/src/util.test.ts
+++ b/prometheus/src/util.test.ts
@@ -205,6 +205,10 @@ describe('formatBytes', () => {
     // values above the largest unit used to come out as "1.00undefined"
     [1024 ** 5, '1024.00TB'],
     [1024 ** 6, '1048576.00TB'],
+    // same defect on the other side: log(bytes) < 0 for 0 < bytes < 1, which
+    // used to index units[-1] and print "undefined" too
+    [0.5, '0.50B'],
+    [-5, '-5.00B'],
   ])('formatBytes(%d) -> %s', (bytes, expected) => {
     expect(formatBytes(bytes)).toBe(expected);
   });
--- a/prometheus/src/util.test.ts
+++ b/prometheus/src/util.test.ts
@@ -1,4 +1,9 @@
-import { getTimeRangeAndStepSize, isArgoCDApplication, supportsPrometheusMetrics } from './util';
+import {
+  formatBytes,
+  getTimeRangeAndStepSize,
+  isArgoCDApplication,
+  supportsPrometheusMetrics,
+} from './util';
 
 beforeAll(async () => {
   global.TextEncoder = require('util').TextEncoder;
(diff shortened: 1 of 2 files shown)


Ralthos (commented):
The clamp is right and the table makes the old failure obvious, including keeping `0` pinned as
its own case so the fix cannot swallow it.

One thing: the clamp is one-sided. `Math.min` caps the index at `TB`, but nothing stops it going
below `B`, and the same `units[i]` lookup falls off the other end of the array.

```
formatBytes(0.5)    -> "512.00undefined"
formatBytes(0.001)  -> "1.02undefined"
formatBytes(-5)     -> "NaNundefined"
```

For `0 < bytes < 1` the log is negative, `Math.floor` gives -1, and `units[-1]` is `undefined`.
A negative input gives `NaN` from the log, `Math.min(4, NaN)` is `NaN`, and `units[NaN]` is
`undefined` too.

This is reachable from the charts rather than only in theory. `NetworkChart`, `DiskChart` and the
`StrimziChart` axis `tickFormatter` all pass Prometheus values straight in, and a rate query
returns bytes per second, which sits below 1 on a quiet interface. Recharts also generates its
own tick values from the domain, so the formatter sees numbers no query produced.

`Math.max(0, ...)` around the existing expression covers both, since `Math.max(0, NaN)` is `NaN`
and would still need the guard, so something like:

```ts
const exponent = Math.floor(Math.log(bytes) / Math.log(1024));
const i = Number.isFinite(exponent) ? Math.min(units.length - 1, Math.max(0, exponent)) : 0;
```

Then `0.5` reads `0.50B` and a negative reads `-5.00B`, which is at least a unit that exists.

Worth two more rows in the table you already have, since it is the same defect on the other side
and the current cases would not catch a regression there.

magic-peach:
@Ralthos could you please check now?

Ralthos:
Checked, and it holds at both ends now:

```
0        -> 0.00B
0.5      -> 0.50B
-5       -> -5.00B
1024**4  -> 1.00TB
1024**5  -> 1024.00TB
1024**6  -> 1048576.00TB
```

The `Number.isFinite(exponent)` guard also covers `NaN` and `Infinity`, which now come out as
`NaNB` and `InfinityB`. Not pretty, and a real unit beats `undefined`, so I would leave that
alone unless you want a separate case for it.

The two new rows are the ones that would have caught this, and they sit next to the row for the
upper bound, so the table now describes the whole domain rather than one side of it.

Issue #1083, opened 2026-08-08
Install detection: one pattern for all CRD plugins


Ralthos:
## Summary

Nine plugins detect whether their operator is installed, in four different ways, and most collapse every probe failure into "not installed". [#972](https://github.com/headlamp-k8s/plugins/issues/972) and [#1033](https://github.com/headlamp-k8s/plugins/issues/1033) report the consequence against Flux and Argo CD, [#1034](https://github.com/headlamp-k8s/plugins/issues/1034) reports the inverse symptom against Kyverno, and there are four open PRs fixing it plugin by plugin.

On [#972](https://github.com/headlamp-k8s/plugins/issues/972) the policy question was answered: sidebar entries should only be hidden when a successful probe shows the CRDs are absent. This proposes how to apply that consistently, and raises one measurement that I think changes which probe each plugin should use.

## What the plugins do today

| Plugin | Probe | On failure |
|---|---|---|
| cert-manager | `ApiProxy.request('/apis/cert-manager.io/v1')` | `catch` returns false |
| keda | `ApiProxy.request('/apis/keda.sh/v1alpha1')` | `catch` returns false |
| kubeflow | `ApiProxy.request(apiPath)`, five groups | `catch` returns false |
| volcano | `ApiProxy.request(apiPath)`, `Promise.all` | `catch` returns false |
| kyverno | `ApiProxy.request(path)` via `checkAPIGroup`, five groups | `catch` returns false |
| flux | `CustomResourceDefinition.apiList` | error callback sets false |
| argocd | `CustomResourceDefinition.apiList` | error callback sets false |
| knative | `CustomResourceDefinition.apiGet` | error callback settles false |
| kro | `CustomResourceDefinition.useGet` | `notInstalled` only when `error?.status === 404` |

kro is the one that already does what [#972](https://github.com/headlamp-k8s/plugins/issues/972) asks, and says so in a comment:

> Only a 404 counts as "not installed": on RBAC errors the check stays neutral so pages can attempt to render and degrade on their own.

## The measurement

The two probe styles do not behave the same way under RBAC, which I do not think has come up in the existing threads.

Using a ServiceAccount whose ClusterRole grants get/list/watch on pods, services, namespaces and deployments, and nothing else:

**API group discovery is not RBAC gated.**

```
$ kubectl get --raw '/apis/velero.io/v1' --as=system:serviceaccount:default:limited
[...]

PR #1206, opened 2026-08-18
cert-manager, keda, volcano: Only report not installed on a 404 (#972)

prabindersinghh:
These three plugins decide whether their operator is installed by asking for the
API group and treating any failure as absence:

```ts
try {
  await ApiProxy.request('/apis/cert-manager.io/v1', { method: 'GET' });
  return true;
} catch (error) {
  return false;
}
```

A transient failure therefore renders the "not installed" banner on a cluster
that does have the operator. The SDK synthesises `status: 502` when `fetch`
throws and `408` on timeout (`lib/k8s/api/v1/clusterRequests.js`), and both land
in that catch. The user gets told to install something that is already there,
and the underlying error is never shown.

The probes now return one of three results instead of a boolean:

```ts
export type InstallProbeResult = 'installed' | 'absent' | 'unreachable';
```

Only a 404 counts as `absent`. Per the measurements @Ralthos posted on #972, API
group discovery is not RBAC gated: Kubernetes binds the `system:discovery`
ClusterRole to `system:authenticated`, so any authenticated user gets a real
answer from `/apis/<group>/<version>`, and a missing group still returns
NotFound. That makes 404 a dependable signal for these three, and it means the
RBAC wrinkle that complicates the CRD probe plugins does not arise here.

When the probe cannot complete, the views render as normal and the actual list
or detail request reports the real error. That seemed better than guessing,
since the request that follows will fail with a more useful message anyway.

### Volcano

Volcano probes more than one group, and the old code wrapped a single
try/catch around `Promise.all`, so the first rejection threw away the results of
every other probe. Each group now catches its own failure and the results get
combined afterwards. If any group is unreachable the whole check is unreachable,
since not being able to reach a group is not evidence that the feature is
missing.

### Cluster switching

The hooks ran their probe once on mount with an empty dependency array, so the
answer from the first cluster stuck around after switching. They now re-probe
when the cluster changes and ignore a result that arrives after unmount.

### Scope

This is the discovery probe half of what came up in #972. #984 covers flux,
which probes CRDs and has the RBAC case to deal with. Kubeflow uses the same
discovery pattern in `kubeflow/src/checks/isKubeflowInstalled.ts` and still has
the old behaviour. I left it out to keep this reviewable, but happy to send a
follow-up or fold it in here if you would rather have it in one go.

Worth flagging that #1075 adds a knative test asserting the probe returns false
on failure, which codifies the behaviour this PR moves away from.

### Testing

New unit tests for each probe cover 404 to absent, and 403, 502, 408 and a
rejection with no status to unreachable. The volcano tests also cover the
multi group combinations, including one group absent while another is
unreachable.

Ran per plugin against the local toolchain:

| plugin | tsc | test | lint | build |
| --- | --- | --- | --- | --- |
| cert-manager | pass | 6 passed, 1 skipped | pass | pass |
| keda | pass | 6 passed, 1 skipped | pass | pass |
| volcano | pass | 26 passed, 1 skipped | pass | pass |

The skipped one in each is the existing storyshots test.

I could not exercise the unreachable path against a live cluster, so that side is
covered by unit tests only.

### DIFF
--- a/cert-manager/src/components/certificateRequests/Detail.tsx
+++ b/cert-manager/src/components/certificateRequests/Detail.tsx
@@ -13,11 +13,11 @@ import {
 export function CertificateRequestDetail() {
   const { t } = useTranslation();
   const { name, namespace } = useParams<{ name: string; namespace: string }>();
-  const { isManagerInstalled, isCertManagerCheckLoading } = useCertManagerInstalled();
+  const { notInstalled, isLoading } = useCertManagerInstalled();
 
   return (
     <>
-      {isManagerInstalled ? (
+      {!notInstalled && !isLoading ? (
         <DetailsGrid
           resourceType={CertificateRequest}
           name={name}
@@ -94,7 +94,7 @@ export function CertificateRequestDetail() {
           }
         />
       ) : (
-        <NotInstalledBanner isLoading={isCertManagerCheckLoading} />
+        <NotInstalledBanner isLoading={isLoading} />
       )}
     </>
   );
--- a/cert-manager/src/components/certificateRequests/List.tsx
+++ b/cert-manager/src/components/certificateRequests/List.tsx
@@ -6,9 +6,9 @@ import { NotInstalledBanner } from '../common/CommonComponents';
 
 export function CertificateRequestsList() {
   const { t } = useTranslation();
-  const { isManagerInstalled, isCertManagerCheckLoading } = useCertManagerInstalled();
+  const { notInstalled, isLoading } = useCertManagerInstalled();
 
-  return isManagerInstalled ? (
+  return !notInstalled && !isLoading ? (
     <ResourceListView
       title={t('Certificate Requests')}
       resourceClass={CertificateRequest}
@@ -44,6 +44,6 @@ export function CertificateRequestsList() {
       ]}
     />
   ) : (
-    <NotInstalledBanner isLoading={isCertManagerCheckLoading} />
+    <NotInstalledBanner isLoading={isLoading} />
   );
 }
--- a/cert-manager/src/components/certificates/Detail.tsx
+++ b/cert-manager/src/components/certificates/Detail.tsx
@@ -12,11 +12,11 @@ import { IssuerRef, NotInstalledBanner, StringArray } from '../common/CommonComp
 export function CertificateDetail() {
   const { t } = useTranslation();
   const { name, namespace } = useParams<{ name: string; namespace: string }>();
-  const { isManagerInstalled, isCertManagerCheckLoading } = useCertManagerInstalled();
+  const { notInstalled, isLoading } = useCertManagerInstalled();
 
   return (
     <>
-      {isManagerInstalled ? (
+      {!notInstalled && !isLoading ? (
         <DetailsGrid
           resourceType={Certificate}
           name={name}
@@ -260,7 +260,7 @@ export function CertificateDetail() {
           }
         />
       ) : (
-        <NotInstalledBanner isLoading={isCertManagerCheckLoading} />
+        <NotInstalledBanner isLoading={isLoading} />
       )}
     </>
   );
--- a/cert-manager/src/components/certificates/List.tsx
+++ b/cert-manager/src/components/certificates/List.tsx
@@ -7,9 +7,9 @@ import { NotInstalledBanner } from '../common/CommonComponents';
 
 export function CertificatesList() {
   const { t } = useTranslation();
-  const { isManagerInstalled, isCertManagerCheckLoading } = useCertManagerInstalled();
+  const { notInstalled, isLoading } = useCertManagerInstalled();
 
-  return isManagerInstalled ? (
+  return !notInstalled && !isLoading ? (
     <ResourceListView
       title={t('Certificates')}
       resourceClass={Certificate}
@@ -45,6 +45,6 @@ export function CertificatesList() {
       ]}
     />
   ) : (
-    <NotInstalledBanner isLoading={isCertManagerCheckLoading} />
+    <NotInstalledBanner isLoading={isLoading} />
   );
 }
--- a/cert-manager/src/components/challenges/Detail.tsx
+++ b/cert-manager/src/components/challenges/Detail.tsx
@@ -17,11 +17,11 @@ import {
 export function ChallengeDetail() {
   const { t } = useTranslation();
   const { name, namespace } = useParams<{ name: string; namespace: string }>();
-  const { isManagerInstalled, isCertManagerCheckLoading } = useCertManagerInstalled();
+  const { notInstalled, isLoading } = useCertManagerInstalled();
 
   return (
     <>
-      {isManagerInstalled ? (
+      {!notInstalled && !isLoading ? (
         <DetailsGrid
           resourceType={Challenge}
           name={name}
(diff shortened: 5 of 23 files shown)


Ralthos (commented):
This is the right fix, and the three-value result is the shape it needed. `unreachable` renders
the content and lets the real request report the failure, so nobody gets told to install what
they already have.

The tests are doing real work too. Covering 403, 502, 408 and a rejection carrying no `status`
at all pins the exact behaviour the old boolean threw away, and mocking `ApiProxy` at the module
boundary means they exercise the shipped function, not a copy of it.

Two things.

**The precedence in `probeApiGroups` discards a result the code has already proven.**

```ts
if (results.includes('unreachable')) {
  return 'unreachable';
}
return results.includes('absent') ? 'absent' : 'installed';
```

I can see from `prefers unreachable over absent when a probe was inconclusive` that this is
deliberate, so I am asking about the semantics here.

The groups passed to `probeVolcanoCoreInstalled` are a conjunction: the views need scheduling
*and* job. If scheduling returns 404 and job returns 502, the feature is definitively not
installed. One required group is provably missing, and no outcome for the other group can change
that. The 404 settles the question on its own, and the 502 currently overrides it.

That inverts the bug this PR fixes. On a cluster genuinely without Volcano, one flaky probe
turns a correct "not installed" banner into the resource views, which then fail on their own.
Less harmful than the false banner, and still a case where the code holds the answer and does
not use it.

I think `absent` should win when any required group returns 404, leaving `unreachable` for when
nothing definitive came back at all:

```ts
if (results.includes('absent')) {
  return 'absent';
}
return results.includes('unreachable') ? 'unreachable' : 'installed';
```

That keeps the property you are protecting, since a probe which never completed still cannot
produce `absent` by itself. If the current order is deliberate for a reason I have missed, it is
worth a line in the doc comment: the comment argues the general principle without noting that a
conjunctive requirement inverts it.

**`isInstalled` has no caller, and it is a trap.**

All three hooks return `isInstalled`, `notInstalled` and `isLoading`, and every call site gates
on `!notInstalled && !isLoading` or `notInstalled || isLoading`. Nothing reads `isInstalled`.

That is fine today, and it is how `unreachable` ends up rendering the content. The problem is
the next person to touch this. `if (isInstalled)` looks like the obvious way to use the hook,
and it silently restores the old behaviour: `unreachable` stops rendering the content and the
banner comes back. Either drop it, or say in a comment that gating on it reintroduces the bug
this PR removes.

One follow-up, and not something I am asking you to change here. The probe, the type and the doc
comment are now duplicated across cert-manager, keda and volcano. That is three copies of a
decision it took a measurement to reach, and the failure mode is one of them drifting later.
#1083 was about exactly this. Happy to do that consolidation once this lands, since the
behaviour is what matters first.

On the RBAC line in the doc comments: that matches what I measured on #972. A ServiceAccount
denied every resource in the group still got 200 from discovery, and the CRD read returned 403.
One detail worth having is that the bound subject is `system:authenticated`, so an anonymous
request is not covered and can see a 403 on discovery. That lands in `unreachable`, which is the
correct answer for it.

Ralthos:
One more thing, about the way these tests are written.

All three test files reach for the same escape hatch:

```ts
const { request } = vi.hoisted(() => ({ request: vi.fn() }));
vi.mock('@kinvolk/headlamp-plugin/lib', () => ({ ApiProxy: { request } }));
```

That works, and for a probe whose only dependency is `ApiProxy.request` it is arguably the right
call. But it is the second distinct workaround this repo has grown for the same underlying
problem, and the two are worth comparing while they are both in one thread.

On #1079 I hit the other side of it. `findAuthenticationEdges` lived in `mapView.tsx`, which
imports the detail components, which import the resource classes, which import
`@kinvolk/headlamp-plugin/lib/k8s/cluster`. The module could not be loaded under test at all, so
I moved the logic into its own file instead of mocking the SDK. Mocking would have meant
stubbing a surface I did not use.

Both exist because of kubernetes-sigs/headlamp#7036: plugin modules that import SDK
values cannot be unit tested because the path does not resolve on disk. 86 files are affected,
and it looks like the reason most plugins ship with no tests at all.

The two workarounds have different costs. Mocking the module is cheap and local, and it goes
stale silently when the SDK changes shape, since the test keeps passing against a fake that no
longer resembles the real thing. Extracting is more invasive and survives that, but it only
works when the logic is separable, which a probe calling one SDK function largely is not.

Nothing here is a change request for this PR. I raise it because a reader of this diff will
copy the mock into the next plugin, and it would be better if that were a decision the repo made
once. If there is appetite, the shared install-probe helper I mentioned above is a natural place
to settle it: one module, one mock, tested in one location, and the nine plugins in #1083 stop
each inventing their own answer.

prabindersinghh:
Thanks, the precedence point is right and I had it backwards.

I was applying the general principle, that an inconclusive probe is not evidence of
absence, without noticing that a conjunction inverts it. If scheduling returns 404
then Volcano's core cannot be there, and nothing the job probe returns changes that.
Flipped in c2ea975:

```ts
if (results.includes('absent')) {
  return 'absent';
}
return results.includes('unreachable') ? 'unreachable' : 'installed';
```

The test that pinned the old order now asserts `absent`, and I added one for the
property you were checking I had not lost: two inconclusive probes with no 404
anywhere still give `unreachable`, so a probe that never completed cannot produce
`absent` by itself. The doc comment says why a single 404 is conclusive here, since
the old one argued the general rule and skipped the exception.

On `isInstalled`, I dropped it. Nothing read it, and I would rather it not exist than
ship a comment asking people not to use it. The hooks return `notInstalled` and
`isLoading`, and the doc now says why there is no positive flag.

I also took the `system:authenticated` detail into the probe docs. An anonymous
request is outside that binding and can get a 403 from discovery, and `unreachable`
is the right place for that to land.

On the shared helper and the mocking question, agreed on both, and both look like
they want to be one decision rather than three. Happy to do the consolidation as a
follow-up once this lands, and if the repo wants to settle the test approach in the
same place that seems like the right home for it. I have no strong preference on
which way it goes, only that the nine plugins in #1083 should not each answer it
differently.

Ralthos:
Checked the new commit. The conjunction case returns `absent` now, `never reports absent from
inconclusive probes alone` pins the property in the other direction so the flip cannot quietly
go too far, and the doc comment says why one 404 settles it. That all reads right.

Dropping `isInstalled` is better than what I suggested. I proposed a comment warning people off
it; removing it means there is nothing to warn about.

On whether the testing approach should be settled in the same place: I would keep the two apart,
and I think the consolidation makes that easier rather than harder.

Consolidating the probe answers the mocking question for install detection largely by dissolving
it. One module, one dependency on `ApiProxy.request`, mocked in one file, tested in one file, and
the plugins import a function instead of each standing up their own fake. That is worth doing for
its own sake.

What it does not settle is the general case, which is kubernetes-sigs/headlamp#7036: any plugin
module importing SDK values cannot be loaded under test, and install detection is one instance.
A consolidation PR is the wrong room for repo-wide test policy, because the people who would
have to live with the decision are not reading that diff. Better to let the consolidation show
the pattern working and argue the general question on #7036, where it stands on its own.

I will take the consolidation once this lands. I will keep it to the behaviour you have arrived
at here, so it is a move and not a redesign, and open it against the plugins in #1083 that this
PR does not touch. Converting cert-manager, keda and volcano can come last, so nothing here has
to move.

PR #1212, opened 2026-08-18
fix(keda): sort Jobs list by completions, not parallelism

magic-peach:
sortByCompletions in the keda Jobs list sorted by `parallelism` first and only fell back to `completions` on a tie, which is backwards from what the name says it does. Two jobs with the same parallelism but very different completion counts would land in the right order by luck, but anything where the two disagreed sorted wrong.

Also switched to nullish coalescing on both fields — a job missing `completions` or `parallelism` (both are optional on the Job spec) turned the comparison into `NaN`, which `Array.sort` treats unpredictably.

Exported the function out of `JobsListRenderer` so it can actually be unit tested instead of only exercised through the rendered table.

### DIFF
--- a/keda/src/components/common/CommonComponents.test.ts
+++ b/keda/src/components/common/CommonComponents.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { sortByCompletions } from './CommonComponents';
+import { getCompletions, sortByCompletions } from './CommonComponents';
 
 function job(completions?: number, parallelism?: number) {
   return { spec: { completions, parallelism } } as any;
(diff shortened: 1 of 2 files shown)


Ralthos (commented):
Sorting by completions matches what the column shows, since `getCompletions` renders
`completions/parallelism` and the label is Completions. Ordering on the second number while
displaying the first was the bug, and the test that pins job(1, 5) before job(5, 1) is the one
that proves the direction actually changed.

Moving it out of the component to make it importable is the right call too.

One thing you are half way to fixing. The `?? 0` guards are new here, and the function directly
above still has the unguarded version of the same problem:

```ts
function getCompletions(job: Job) {
  return `${job.spec.completions}/${job.spec.parallelism}`;
}
```

`completions` and `parallelism` are both optional in `batch/v1`, so a Job with neither set now
sorts as 0, which is sensible, and displays as `undefined/undefined`, which is not. After this PR
the two behaviours disagree about the same missing field, and the sort order will look wrong to
anyone reading the column, because the rows say `undefined` and sit where `0` belongs.

Cheap to close while you are in here:

```ts
export function getCompletions(job: Job) {
  return `${job.spec.completions ?? 0}/${job.spec.parallelism ?? 0}`;
}
```

Or render a dash if unset should read as unknown instead of zero. Either is defensible. What is
hard to defend is the sort treating it as 0 and the cell treating it as a string that says
`undefined`.

magic-peach:
@Ralthos could you please check now?

Ralthos:
Checked. `getCompletions` is exported and guarded, and the test that pins `0/0` sits directly
beside the sort tests, so the display and the ordering are now held to the same answer for a
missing field by the same file.

That was the part worth closing. A cell reading `undefined` above rows ordered as though it read
`0` would have sent someone looking for a sorting bug that was not there.

PR #1215, opened 2026-08-18
kmesh: map: Add traffic map for Waypoints, Namespaces, Services

itvi-1234:
### Description

Registers a KMesh source into Headlamp's shared Map view showing Namespace -> Waypoint -> Service relationships. Edges are derived from the istio.io/use-waypoint label (Service label takes priority over its Namespace's label), not ownerReferences, since KMesh/Istio ambient resources are routed to a waypoint rather than owned by it.

Only namespaces/services actually enrolled with a waypoint are shown, so the map doesn't balloon to the whole cluster. The Waypoint node reuses the existing WaypointDetail component for its side-panel, via the same makeDetailsComponent adapter pattern used by the Strimzi and Volcano plugins' map views.

The istio.io/use-waypoint label logic and the Gateway condition-to-map-status mapping live in pure, unit-tested helpers (src/map/mapUtils.ts), following the existing waypointUtils.ts convention so they're verifiable in isolation from the map source hooks that consume them.

Pod/backend-level nodes from live eBPF data are left for a follow-up, since sourcing them needs the daemon proxy rather than a plain useList() hook.

image

### DIFF
--- a/kmesh/src/index.tsx
+++ b/kmesh/src/index.tsx
@@ -14,6 +14,7 @@ import KmeshNodeInfoDetail from './components/nodeinfo/Detail';
 import KmeshNodeInfoList from './components/nodeinfo/List';
 import WaypointDetail from './components/waypoints/Detail';
 import WaypointList from './components/waypoints/List';
+import { registerKmeshMapExtensions } from './mapRegistration';
 import { kmeshRouteNames, kmeshRoutePaths } from './utils/kmeshRoutes';
 
 /**
@@ -190,3 +191,6 @@ registerRoute({
 
 // Register Namespace Enrollment section in the details view
 registerDetailsViewSection(NamespaceEnrollment);
+
+// Traffic Map: Namespace -> Waypoint -> Service, into Headlamp's shared Map view
+registerKmeshMapExtensions();
--- a/kmesh/src/map/mapUtils.test.ts
+++ b/kmesh/src/map/mapUtils.test.ts
@@ -1,0 +1,61 @@
+import { describe, expect, it } from 'vitest';
+import {
+  getWaypointNodeStatus,
+  isWaypointEnrolled,
+  resolveEffectiveWaypointName,
+  USE_WAYPOINT_LABEL,
+} from './mapUtils';
+
+describe('getWaypointNodeStatus', () => {
+  it('returns success for Programmed', () => {
+    expect(getWaypointNodeStatus('Programmed')).toBe('success');
+  });
+
+  it('returns success for Accepted', () => {
+    expect(getWaypointNodeStatus('Accepted')).toBe('success');
+  });
+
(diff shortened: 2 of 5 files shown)


Ralthos (commented):
The edge building is careful in the place it usually goes wrong. `serviceSource` matches the
waypoint on name *and* namespace and returns `null` when nothing matches, so an unresolvable
reference produces no edge instead of a broken one. That is the right call and it is not the
default outcome when relationships come from labels.

Two things.

**A namespace that hosts the waypoint it uses gets two edges.**

`waypointSource` draws namespace to waypoint for every KMesh waypoint:

```ts
const ns = namespaces.find(n => n.metadata.name === waypoint.metadata.namespace);
return ns ? makeKubeToKubeEdge(ns, waypoint, 'hosts') : null;
```

`namespaceSource` draws the same pair again when the namespace carries the label:

```ts
const target = kmeshWaypoints.find(
  w => w.metadata.name === waypointName && w.metadata.namespace === ns.metadata.name
);
return target ? makeKubeToKubeEdge(ns, target, 'uses') : null;
```

The second lookup requires the waypoint to be in that same namespace, which is the case the
first one already covers. So a namespace labelled `istio.io/use-waypoint: <waypoint in itself>`,
which I understand to be the ordinary ambient layout, ends up with a `hosts` edge and a `uses`
edge between the same two nodes.

If both facts are worth showing, one edge carrying both labels reads better than two lines. If
`hosts` is really about placement and `uses` about routing, it may be clearer to let
`namespaceSource` handle only waypoints outside its own namespace, so the two sources stop
overlapping.

**An enrolled service with no matching waypoint becomes a node with nothing attached.**

`enrolledServices` decides membership from the label alone:

```ts
const enrolledServices = services.filter(svc => {
  const ns = namespacesByName.get(svc.metadata.namespace);
  return isWaypointEnrolled(svc.metadata.labels, ns?.metadata.labels);
});
```

Every one of those becomes a node, but the edge is only drawn when a waypoint of that name
exists in the same namespace. A service labelled for a waypoint that is missing, misspelled, or
deployed elsewhere still lands on the map, floating, with no indication of why.

The related question: can `istio.io/use-waypoint` name a waypoint in another namespace? If
Istio's cross-namespace form is in play, then services legitimately enrolled against a waypoint
elsewhere are exactly the ones that appear unconnected here. Either way the node and the edge
should agree: include the service when it resolves, or keep it and make the unresolved state
visible.

One note on testing, and not a change request. `mapUtils.test.ts` covers
`getWaypointNodeStatus` across all five phases and `resolveEffectiveWaypointName` including the
precedence case. The part with the bugs in it
is the edge building inside `useData`, which cannot be reached from a test while it sits in the
hook. On #1079 I moved KEDA's edge logic into its own module for that reason and the tests went
from impossible to ten cases. Worth considering if you extend this map later.

PR #1218, opened 2026-08-18
kmesh: waypoints/nodeinfo: Add related-resource links to detail pages

itvi-1234:
## Summary
- Adds a "Related Resources" section to the Waypoint detail page, linking to its proxy Pods, its proxy Service (matched via the `istio.io/gateway-name` label set by the Gateway API deployer), and any Namespaces enrolled to use that waypoint (via `istio.io/use-waypoint`).
- Adds a "Related Resources" section to the KmeshNodeInfo detail page, linking to the underlying K8s Node and the kmesh-daemon Pod running on it.

### DIFF
--- a/kmesh/src/components/nodeinfo/Detail.tsx
+++ b/kmesh/src/components/nodeinfo/Detail.tsx
@@ -2,11 +2,14 @@
  * Detail view for a single KmeshNodeInfo resource.
  * Shows full IPsec security state for one node: SPI, addresses, Pod CIDRs, and boot ID.
  */
+import { K8s } from '@kinvolk/headlamp-plugin/lib';
 import {
   MainInfoSection,
+  SectionBox,
   SimpleTable,
   StatusLabel,
 } from '@kinvolk/headlamp-plugin/lib/CommonComponents';
+import { Link as HeadlampLink } from '@kinvolk/headlamp-plugin/lib/components/common';
 import Alert from '@mui/material/Alert';
 import Box from '@mui/material/Box';
 import Chip from '@mui/material/Chip';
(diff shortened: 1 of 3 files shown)


Ralthos (commented):
The Pods and Services rows are scoped correctly. Both `useList` calls pass `namespace` alongside
the `istio.io/gateway-name` selector, so they cannot pick up a same-named waypoint's workloads
from somewhere else.

The namespaces row does not have that scoping:

```ts
const enrolledNamespaces = (namespaces ?? []).filter(
  ns => ns.metadata?.labels?.[USE_WAYPOINT_LABEL] === name
);
```

`namespaces` is an unfiltered cluster-wide list and the predicate compares the waypoint name
only. A Waypoint is namespaced, so two teams can each have one called `waypoint`. Open the one in
`team-a` and the row lists every namespace enrolled with `team-b`'s waypoint as well, presented
as namespaces using this one.

The plugin already draws this distinction elsewhere. In the map source on #1206's neighbour
#1215, the service lookup requires both:

```ts
const target = kmeshWaypoints.find(
  w => w.metadata.name === waypointName && w.metadata.namespace === svc.metadata.namespace
);
```

Two ways to close it, and the choice depends on something I do not know about your deployment
model.

If enrollment is same-namespace only, add `&& ns.metadata.name === namespace`. The row then
resolves to at most the waypoint's own namespace, which may be thin enough that the row is not
worth showing on its own.

If a namespace is meant to be able to point at a waypoint in another namespace, then the name
alone was never enough to resolve it and `istio.io/use-waypoint-namespace` is the field that
disambiguates. It does not appear anywhere in this diff.

Either way the current predicate answers a question nobody asked, which is "which namespaces
name a waypoint with this string".

The two tests here pass because the fixtures have one namespace each. A third fixture, with a
second namespace enrolled against a same-named waypoint elsewhere, would fail today and is the
case worth pinning.

PR #1227, opened 2026-08-20
prometheus: Safely detect Argo CD Application metrics

Joshna907:
## Summary

Adds a safe Prometheus metrics eligibility check for Argo CD Applications.

This is the first half of Argo CD metrics support. It only recognizes the exact argoproj.io/v1alpha1 Application resource, so another Kubernetes resource with the same kind name cannot receive Argo CD metric charts by mistake.

## Changes

- Add Application to the supported Prometheus resource kinds.
- Require the exact argoproj.io/v1alpha1 API version.
- Add positive and negative unit coverage.

## Validation

- TypeScript passed.
- Tests passed.
- Production build passed.

## Follow-up

The chart UI itself is deliberately in a separate follow-up PR, matching the earlier Volcano metrics work.

### DIFF
--- a/prometheus/src/util.test.ts
+++ b/prometheus/src/util.test.ts
@@ -1,4 +1,4 @@
-import { getTimeRangeAndStepSize, supportsPrometheusMetrics } from './util';
+import { getTimeRangeAndStepSize, isArgoCDApplication, supportsPrometheusMetrics } from './util';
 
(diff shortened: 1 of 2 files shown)


Joshna907:
Let's merge this @illume ?

illume (approved):
🎉 thanks!

PR #1228, opened 2026-08-20
prometheus: Add Argo CD Application metric charts

Joshna907:
## Summary

Adds Prometheus charts to Argo CD Application details.

This PR follows the Volcano metrics UI pattern: chart selection, time range, resolution, pause/resume, and the normal no-data and error states.

## Charts

- Sync activity
- Average sync duration
- Orphaned resources

Metrics are queried using the Application CR namespace and name. Prometheus remains optional: if it is disabled or Argo CD does not expose these metrics, the normal no-data state is shown.

## Safety and scope

- Uses only the selected cluster's existing Prometheus connection.
- Adds no Argo CD API token, network configuration, or mutation action.
- Does not claim that an available metric proves Application health or traffic correctness.

## Validation

- TypeScript passed.
- Tests passed.
- Production build passed.

## Dependency

Depends on #1227. Once it merges, this branch will be rebased so this PR contains only the chart changes.

### DIFF
--- a/prometheus/README.md
+++ b/prometheus/README.md
@@ -2,6 +2,11 @@
 
 This plugin adds advanced charts to the details view of workload resources.
 
+It also shows optional Argo CD Application charts for sync activity, average sync duration, and
+orphaned resources. These charts appear only when Prometheus is enabled for the selected cluster
+and the resource is `argoproj.io/v1alpha1` `Application`. Missing Argo CD metrics simply show the
+normal no-data state.
+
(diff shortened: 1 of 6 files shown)


illume (changes requested):
Has some git conflicts now, can you please take a look?

Joshna907:
> Has some git conflicts now, can you please take a look?

@illume ptal now

illume (approved):
🎉 thanks!

PR #1266, opened 2026-09-03
Fix GPT-5.6 Copilot API routing

ashu8912:
## Summary

- route GitHub Copilot GPT-5.6 models through the Responses API instead of `/chat/completions`
- add `gpt-5.6-sol` to the Copilot model picker
- add regression coverage while preserving chat-completions routing for older models

## Validation

- `npx vitest run src/providers/createChatModel.test.ts --passWithNoTests --reporter=dot` (29 tests passed)
- `npm run tsc -- --pretty false`
- `npm run lint -- --format unix`


illume (changes requested):
Thanks for this.

Please take care the git commit messages follow the guidelines?

Can you please confirm you manually tested this? 

ps. This fix does seems very specific to this particular model version and deployment of it, and I wonder if we should be handling this better for other models over all? Are missing APIs going to be more of a thing going forward? Out of scope for this PR, I know.

illume:
Please respond to my review?

> Thanks for this.
> 
> Please take care the git commit messages follow the guidelines?
> 
> Can you please confirm you manually tested this?
> 
> ps. This fix does seems very specific to this particular model version and deployment of it, and I wonder if we should be handling this better for other models over all? Are missing APIs going to be more of a thing going forward? Out of scope for this PR, I know.

ashu8912:
> Thanks for this.
> 
> Please take care the git commit messages follow the guidelines?
> 
> Can you please confirm you manually tested this?
> 
> ps. This fix does seems very specific to this particular model version and deployment of it, and I wonder if we should be handling this better for other models over all? Are missing APIs going to be more of a thing going forward? Out of scope for this PR, I know.

yes i have tested this manually, this only doesn't work with open ai models

illume (approved):

PR #1281, opened 2026-09-08
kyverno: i18n: Fix broken interpolation and stale column caching

Athang69:
## The problem

Two separate, real bugs, both about the plugin's translations never actually working, found while verifying the plugin's UI against a live cluster.

The first: the plugin's `package.json` never declared a `headlamp.i18n` block. Headlamp's plugin loader only creates a real translation instance for a plugin once it declares supported locales that way. Without it, the plugin's `t()` function has nothing to call into, so it just returns whatever string was passed to it, completely unprocessed. For a plain label like `t('Resource')` that happens to look fine, the untranslated key and the intended English text are identical. But any call with a `{{variable}}` placeholder breaks visibly, since nothing ever substitutes the value in. There are 21 of these across the plugin: every "X ({{count}})" section title, the compliance badge's pass and fail counts and its "{{pct}}% compliant" line, both policy list views' "{{fail}} / {{total}} failed" chips, and every "Failed to load X: {{error}}" message. All of them have been rendering literal, unprocessed template text to every user of this plugin, in any language, since it shipped.

The second, related bug: the Violations table builds its columns without an explicit `id` on each one, so Headlamp's shared Table component falls back to using each column's position in the array as its id. Table's internal header cell renderer is memoized against that id only, not against the actual header text, so once a header cell renders once, it will not update again unless its id changes, no matter what text is passed to it afterward. This matters because fixing the first bug (declaring `headlamp.i18n`) means the plugin's translations now load asynchronously for the first time, so there's a real, if brief, window where headers render with no text before the real text is ready. Without a way to signal that change, that gap could leave headers permanently blank instead of showing the fix working.

## The fix

1. Added `"headlamp": { "i18n": ["en"] }` to `kyverno/package.json`, and generated `locales/en/translation.json` with `npm run i18n`, the plugin's own extraction tool. This is the same convention already used by other plugins in this repository, an identity-mapped English locale file, no other languages added yet.
2. Added a `columnId(key, header)` helper in `common.tsx` that builds a column's id out of its own live header text, so the id changes whenever the text does. Applied it to the 8 existing columns in `ViolationsView.tsx`, and stopped memoizing the columns array on a value that was never actually stable across renders.

Together, the second fix is what makes the first one safe to ship. It closes the exact gap the async translation loading opens up.

## Before and after

The Violations page title used to read literally `Violations ({{count}})`. After this fix it reads `Violations (36)`, using the real count. Same mechanism was breaking the compliance percentage, the pass and fail chip counts across list views, and every error message in the plugin, all fixed the same way.

## Verification

- `npx tsc --noEmit` on the plugin: clean
- `npx eslint` on the changed files: clean
- Ran the plugin against a real kind cluster with Kyverno installed and real policy violations across two namespaces, not just the default sample data
- Confirmed the dev build actually picks up the new locale file end to end: rebuilt, copied into the installed plugin directory, loaded by the running app

## Screenshots

1. Before: the Violations page title reading the literal `Violations ({{count}})`, with real violation rows underneath.

Screenshot from 2026-09-07 23-23-07

2. After: the same page, same cluster, title now reading `Violations (36)`.

Screenshot from 2026-09-08 10-32-07

3. The Kyverno Overview dashboard after the fix, real compliance percentage, pass and fail counts, all rendering correctly with no leftover template syntax anywhere on the page.

Screenshot from 2026-09-08 10-33-20

4. Terminal output of `npx tsc --noEmit`

5. Terminal output of `npx eslint src/components/common.tsx src/components/ViolationsView.tsx`.

Screenshot from 2026-09-08 10-40-58


Athang69:
@illume I've addressed the copilot's reviews. PTAL.
Thanks!

PR #1290, opened 2026-09-10
plugin-catalog: Fetch all Artifact Hub pages

### LINKED ISSUE #1289: plugin-catalog: Catalog stops after first Artifact Hub page
## Description

The Plugin Catalog appears to stop after the first Artifact Hub results page when the Artifact Hub `/api/v1/packages/search` response does not include `metadata.total`.

This means valid verified Headlamp plugins beyond the first 60 results are not loaded into the catalog. The search box then cannot find them because it filters the already-loaded local list instead of querying Artifact Hub.

## Example

Artifact Hub currently shows more verified Headlamp plugins than Headlamp displays in the catalog. In my local Headlamp catalog, with "Only Official" disabled, I see 63 plugins but missing verified plugins such as:

- `kubebuddy-headlamp-plugin`
- `kubefleet-manager`

Direct package lookup and install work:

```bash
npx @kinvolk/headlamp-plugin install https://artifacthub.io/packages/headlamp/kubebuddy/kubebuddy-headlamp-plugin

PixelRobots:
This PR fixes plugin catalog pagination when Artifact Hub does not return metadata.total. Previously, missing metadata.total was treated as 0, which caused the catalog to stop after the first page of 60 results. Since search is performed client-side against the loaded plugin list, plugins on later Artifact Hub pages could not be found. 

Fixes #1289

Steps to Test:
1. Run npm run test.
2. Run npm run tsc
3. Run npm run lint
4. Open the Plugin Catalog with Only Official disabled.
5. Verify plugins beyond the first Artifact Hub page are loaded and searchable.

### DIFF
--- a/plugin-catalog/src/components/plugins/List.tsx
+++ b/plugin-catalog/src/components/plugins/List.tsx
@@ -9,6 +9,7 @@ import { isEqual } from 'lodash';
 import React, { useEffect, useMemo, useState } from 'react';
 import semver from 'semver';
 import { getExternalProxyEndpoint } from './externalProxy';
+import { fetchAllPages } from './pagination';
 import { PluginCard } from './PluginCard';
 
 const PAGE_SIZE = 60; // Maximum allowed by the API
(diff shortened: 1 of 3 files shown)


skoeva (approved):
lgtm, thanks for this!

